Issue #56, opened 2019-04-18
Add Composer type attribute


christianc1:
To ease installation via Composer, suggesting we add the following to the `composer.json`.
```js
{
  "type": "wordpress-plugin"
}
```

jeffpaul:
@christianc1 seems reasonable, want to fire off a quick PR to cover that?

christianc1:
Done @jeffpaul .  Also added `license` attribute so `composer validate` won't throw any errors/warnings.

jeffpaul:
@christianc1 awesome, many thanks!  I'll let this sit until @helen can give the `composer.json` file another review to see if anything else should be added alongside your additions.  Thanks again!

helen:
Fixed in #57

PR #57, opened 2019-04-18
Improve composer.json

christianc1:
### Improve composer.json

This pull request makes two changes to the composer.json file.

The first adds the `type` attribute, which designates this as WordPress plugin to ease the process for adding this as a dependency to a WordPress installation as a composer package. [More info here.](https://getcomposer.org/doc/04-schema.md#type)

The second adds the `license` attribute to the composer.json file.  Without this present, the composer.json file is not technically valid, as demonstrated by running `composer validate`. [More info here.](https://getcomposer.org/doc/04-schema.md#license)

### Benefits

Valid composer.json that a WordPress site can add/manage via Composer.

### Verification Process

`composer validate`

### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://github.com/10up/classifai-for-wordpress/blob/develop/CONTRIBUTING.md) document.

### Applicable Issues
#56


helen:
@christianc1 Does it work to include this via Composer when the built files are not included in the repo?

christianc1:
@helen I installed this on a site that's still on Classic Editor.  Looks like the dist js is for Gutenberg, so I didn't even realize.  We can probably hold while we figure out if we want to support Composer based installs.  Thinking about how we distribute this, does it might make sense regardless to have build files in the repository?

helen:
Sorry, I completely lost track of this! So the production-ready version of this plugin, like Distributor, lives in the `stable` branch. If we keep the `composer.json` file in there, would that be good enough?

christianc1:
Works for me @helen

PR #62, opened 2019-04-25
Add a GitHub Action for building a production-ready version

helen:
### Description of the Change

Use GitHub Actions to build the plugin on pushes to `master`, copy only necessary files for running in production to the `stable` branch, and commit those changes. Releasing a new version via tag on `stable` is a manual process to retain a safety net.

### Alternate Designs

This is heavily inspired by the way Distributor does things, but leverages GitHub Actions instead of requiring somebody to run this locally.

### Benefits

Automation and fewer human-susceptible steps for releasing, yay!

### Possible Drawbacks

Might be a bit of an intermediate step if ever this becomes a broader-use Action but that's okay because this can inform that.

### Verification Process

Ran various commands locally. Still won't know for sure what is or isn't working until we actually run the Action in the GH environment (suspect that we need to add the actual GH remote and use the token to push to it).

### Checklist:

- [x] I have read the [**CONTRIBUTING**](https://github.com/10up/classifai-for-wordpress/blob/develop/CONTRIBUTING.md) document.


helen:
If this works we will probably want to ignore `vendor/yahnis-elsts` and `languages/classifai.pot` and have the build process also run `composer install` and `npm run makepot`. I also don't know if I love the rsync filter file but let's just go with it for now.

Issue #78, opened 2019-06-11
PHP version support


jeffpaul:
**Describe the bug**
On a test site running PHP 5.6, I get the following syntax error:
> [Mon Jun 10 18:19:00 2019] [warn] [client 216.112.17.10] mod_fcgid:
stderr: PHP Parse error:  syntax error, unexpected ‘?’ in
/wp-content/plugins/classifai-1.3.0/includes/Classifai/Providers/Watson/NLU.php
on line 267

**Steps to Reproduce**

1. Install ClassifAI v1.3.0 on an install running PHP 5.6
2. Site explodes

**Expected behavior**
ClassifAI is either unable to be installed and notes the minimum PHP version or catches scenarios where <PHP7 is installed and properly handles it in the code.  Could add code to index file to say if PHP < 7, kill the plugin and show an alert.

**Screenshots**
n/a

**Environment information**
 - Device: Mac mini (2018)
 - OS: MacOS 10.14.5 (18F132)
 - Browser and version: Firefox 67.0.1 (64-bit)
 - ClassifAI version: 1.3.0
 - Theme and version: Twenty Fifteen 2.5
 - Other installed plugin(s) and version(s): Akismet Anti-Spam Version 4.1.2, Authy Two Factor Authentication Version 3.0.1, Customize Posts Version 0.9.1, Gutenberg Version 5.8.0, Instructables Version 2.0.4, VaultPress Version 1.9.10

**Additional context**
Issue appears to reference: https://github.com/10up/classifai/blob/cf5c80b9d56daf347754834d84e0bce11cc4da5e/includes/Classifai/Providers/Watson/NLU.php#L267

PR #108, opened 2019-07-26
Add plugin debug information to the Site Health screen

### LINKED ISSUE #83: Add debug info to Site Health Info screen
WordPress 5.2 introduced the Site Health Info screen.  Adding ClassifAI-specific debug info there can help solve issues that come to us from support forums, GitHub, email, etc. as users can copy this info and send to us for faster assistance with issues.

- [x] display ClassifAI version
- [x] display settings info (service provider details & whether those API connections are valid, plus general Language and Image settings)

**Designs**
n/a

n/a

**Additional context**
n/a

johnwatkins0:
This resolves #83, which describes a need to display plugin debug information on the new Site Health screen introduced to core in 5.2.

### Description of the Change

It adds a new accordion drawer to the Site Health info screen that looks like this: 

classifai-site-health

We do this by hooking into the new `debug_information` filter provided by core, adding a key for `classifai`, and providing a set of plugin details that would be useful for debugging purposes. This is done via a `classifai_debug_information` filter that can be hooked into throughout the plugin.

### Alternate Designs

- The `DebugInfo` class provides the `classifai_debug_information` filter and delegates the adding of debug information to the individual features in the plugin via the filter. Alternatively, we could do all the work needed to retrieve debug information in the `DebugInfo` class. This might be simpler, but I liked the extensibility the filter provides. 
- Currently the IBM Watson features display as a JSON encoded string. These could be broken out into their own lines. I just didn't see a clean way to do it within the limited structure of the Site Health drawer.

### Benefits

Plugin users have a concrete, nontechnical step to share information needed for debugging.

### Possible Drawbacks

Expecting debug information to be provided with all plugin features adds a small amount of overhead.

### Verification Process

Using an up-to-date Version of WP: 

1. In the admin, navigate to Tools/Site Health and switch to the "Info" tab.
2. Find and open the "ClassifAI" accordion drawer.
3. Verify that the information shown matches your environment.
4. Verify the same in a different environment with different settings, or modify your settings and re-check.

### Checklist:

### Applicable Issues

#83


includes/Classifai/Plugin.php
```diff
@@ -78,6 +78,7 @@ public function init_services() {
 		$this->services = [
 			'service_manager'     => new Services\ServicesManager( $classifai_services ),
 			'admin_notifications' => new Admin\Notifications(),
```
johnwatkins0: Putting non-service-related classes in `$this->services` doesn't seem quite right, but I followed what was already there.

includes/Classifai/Providers/Azure/ComputerVision.php
```diff
@@ -199,7 +199,7 @@ public function sanitize_settings( $settings ) {
 			);
 		}
 
-		if ( is_numeric( $settings['caption_threshold'] ) && (int) $settings['caption_threshold'] >= 0 && (int) $settings['caption_threshold'] <= 100 ) {
+		if ( isset( $settings['caption_threshold'] ) && is_numeric( $settings['caption_threshold'] ) && (int) $settings['caption_threshold'] >= 0 && (int) $settings['caption_threshold'] <= 100 ) {
```
johnwatkins0: Not having this check was causing some tests to error.

includes/Classifai/Providers/Watson/NLU.php
```diff
@@ -552,6 +552,8 @@ public function sanitize_settings( $settings ) {
 	 *
 	 * @return bool
 	 * @since  1.2
+	 *
+	 * @todo Is this function supposed to be here?
```
johnwatkins0: This function doesn't seem like it should be here -- it exists in another class, and this copy isn't used anywhere -- but I left it alone.

includes/Classifai/Plugin.php
ryanwelcher: @johnwatkins0 this was a side-effect of the refactor work done to allow the plugin to support multiple services. At the time, there wasn't a need to handle non-services differently. This can probably be added now as we have both notices and debug information .

includes/Classifai/Providers/Provider.php
```diff
@@ -171,4 +180,14 @@ public function sanitize_settings( $settings ) {
...
 		return $settings;
 	}
+
+	/**
+	 * Provides debug information related to the provider.
+	 *
+	 * @return string|array Debug info to display on the Site Health screen. Accepts a string or key-value pairs.
+	 * @since 1.4.0
+	 */
+	public function get_provider_debug_information() {
```
ryanwelcher: @johnwatkins0 as this class is abstract, it can never be instantiated directly so this stub is never called. It may make sense to mark this method as abstract as well to enforce that any providers implement this.

includes/Classifai/Services/ServicesManager.php
```diff
@@ -343,4 +345,31 @@ public function check_license_key( $email, $license_key ) {
...
+
+		$valid_licence       = 1 === intval( $settings['valid_license'] ?? 0 );
+		$debug_information[] = [
+			'label' => __( 'Valid license', 'classifai' ),
+			'value' => $valid_licence,
+		];
+
+		$debug_information[] = [
+			'label' => __( 'Email', 'classifai' ),
+			'value' => $settings['email'] ?? '',
```
ryanwelcher: I think we ran into an issue using the null coalescing operator with a client install - @helen @jeffpaul ?

includes/Classifai/Providers/Azure/ComputerVision.php
ryanwelcher: nice catch!

includes/Classifai/Providers/Watson/NLU.php
ryanwelcher: This is most likely a leftover from the refactoring effort. I'd imagine we'd be safe to remove it but should test that in a separate PR.

includes/Classifai/Services/ServicesManager.php
johnwatkins0: I'm interested in the answer to this question and will hold on my updates until we have it.

FWIW, I took the 2-3 existing instances of the null coalescing operator as implicit permission -- e.g.: 

https://github.com/10up/classifai/blob/develop/includes/Classifai/Providers/Watson/NLU.php#L298

Of course, switching them all out with the PHP5-friendly equivalents is no problem.

ryanwelcher: @johnwatkins0 I followed up on this and this was the issue I was thinking of #78. I think we're fine to move forward with the null coalescing operator. Thanks for your patience!

johnwatkins0:
Thanks, @ryanwelcher ! I've pushed the following updates:

### From CR

- [Made the `get_provider_debug_information` method abstract, forcing child classes to implement it.](https://github.com/10up/classifai/pull/108/commits/a26852d41f315425e50734bc0fbcfe9c3b622e4a)
- [In the `Plugin` class, we now register admin helper class instances separately from the services](https://github.com/10up/classifai/pull/108/commits/43e50cabae08d5ba15171b8678e65dc950ab72fc)

### Not from CR

[Made booleans that show in the debug information more human-friendly](https://github.com/10up/classifai/pull/108/commits/ce9a7e498cd7089934936ed425a4c45be4f5e8fd) -- "yes"/"no" instead of "1"/empty.


Let me know if you have questions or more suggestions.

helen:
I think what I'd like to see added here is the results of the most recent request, maybe to each provider. Could help a lot with troubleshooting. So store those (request response I guess?) in options somewhere and translate them into human-ish terms maybe.

jeffpaul:
I'll add a separate issue to include the details on services within Site Health, that way this PR can land in 1.4 and we can enhance it in 1.5

PR #110, opened 2019-07-29
Computer Vision - send next largest image when full is too big

### LINKED ISSUE #101: Handle `Full` image sizes above 4MB
[Azure Computer Vision currently has a 4MB filesize limit](https://github.com/10up/classifai#set-up-image-processing-via-microsoft-azure), so attempts to send images for processing larger than 4MB won't work with ClassifAI.

We'll likely want to add a check for the image filesize and if it's greater than 4MB, then we'll want to loop through alternate image sizes (e.g., `Large`, `Medium_Large`, `Medium`, `Thumbnail`) to check if any of those are smaller than 4MB and in which case send that image size variant to Azure Computer Vision for processing.

**Designs**
n/a

none.

**Additional context**
A side benefit for adding in this check is that it may add enough of a delay for sites that offload image storage to an external service (e.g., [Microsoft Azure Storage](https://github.com/10up/windows-azure-storage)) for those images to make it to their external storage in time for ClassifAI to send them along to Azure Computer Vision for processing.

johnwatkins0:
This would resolve #101 by providing backup behavior when the full-sized version of an uploaded image is greater than the maximum size accepted by Computer Vision.

Note: This does not intentionally address the "Additional Context" comment on #101 about this update potentially giving time for external image-hosting services to receive the image before the request is sent to Computer Vision. I suggest a separate issue for that. Maybe this work could be done with `wp_schedule_single_event` instead of in the same pageload as the image upload.

### Description of the Change

In the `ComputerVision` class, this adds a method to provide the URL of a newly uploaded attachment's largest acceptable version.

In most cases -- assuming most WP users don't routinely upload images greater than 4MB -- we would expect the full-size image to work, in which case this update doesn't substantively change any existing plugin behavior. When an uploaded image is greater than 4MB, however, we make a copy of the image's `size` metadata, sort the sizes from largest to smallest, and use the first size whose file is less than 4MB. 

### Alternate Designs

A simpler approach would be to rely on the built-in core file sizes ('large', 'medium', etc.), but we don't actually know whether 'large' will be a given site's largest file size below 'full' -- and it is possible to disable core intermediate image sizes entirely -- so this approach uses a simple algorithm to sort images from largest to smallest regardless of how the sizes are named.

### Benefits

Within any site that does not have some radically altered way of handling images, Computer Vision will be able to provide data for any uploaded image.

### Possible Drawbacks

Calls to the PHP `filesize` function become expensive when images are hosted externally, so the behavior implemented here should be a fallback and should not run routinely.

### Verification Process

Using a site that is reachable over the web -- which is required because Computer Vision accesses the images via a public URL -- upload an image that is larger than 4MB. After the upload is complete, the expected image metadata should have been generated.

### Checklist:

### Applicable Issues

#101


includes/Classifai/Providers/Azure/ComputerVision.php
```diff
@@ -81,6 +106,50 @@ public function generate_alt_tags( $metadata, $attachment_id ) {
 		return $metadata;
 	}
 
+	/**
+	 * Retrieves the URL of the largest version of an attachment image accepted by the ComputerVision service.
+	 *
+	 * @param string $full_image The path to the full-sized image source file.
```
adamsilverstein: please keep these `@param` descriptions aligned

adamsilverstein:
@johnwatkins0 Overall the code looks good here.

Testing on a temporary public site however, I never saw images above 4m working. I set the threshold at 50% and small images always worked, large images never worked.

Am I missing some step to testing, are you always getting tags for larger images after your PR?

Spin up a site and test: http://poopy.life/create?plugins=https://github.com/10up/classifai/zipball/feature/handle-overlarge-images

johnwatkins0:
Thanks, @adamsilverstein . I'll see if I can reproduce the issue you're having.

johnwatkins0:
@adamsilverstein This is fixed now. Seems like I introduced a last-minute bug involving a function parameter. Thanks for testing!

Feel free to use this environment if you get to it in the next 22 hours -- it has the latest on this branch: http://adaptable-wren.w6.poopy.life/ I removed my image processing credentials. 

Tested with 23MB image:

Screen Shot 2019-09-05 at 8 52 09 PM

adamsilverstein:
@johnwatkins0 Thanks for fixing that up, I'll give it another test (this link should install from this branch: http://poopy.life/create?plugins=https://github.com/10up/classifai/zipball/feature/handle-overlarge-images).

Looks like some merge conflicts have cropped up, can you take a look at those?

adamsilverstein:
Generally this is working much better. I received some tagging results on both small and large images. Yea!

I still see an issue that I think is coming from this PR: when I drag and drop an image to upload it (vs clicking 'Select Image' I regularly see an HTTP error message. Oddly the same image uploads fine with the button select method, or sometimes when dragging and dropping, maybe if other uploads are already in progress?

Have you seen this?

image

johnwatkins0:
Thanks, @adamsilverstein . I will look into that.

adamsilverstein:
@johnwatkins0 Thanks! once you resolve the merge conflicts I can re-test; the issue may have been fixed elsewhere.

johnwatkins0:
@adamsilverstein I've fixed the merged conflicts. I'm not able to reproduce that issue so far, but I might not be following the right steps. Let me know if you still see it.

adamsilverstein:
@johnwatkins0 

Steps to reproduce the HTTP error issue:

* Go to Media and Click the `Add New` button and use this interface in Media:

image

* The "Add New" page also has the same issue: `/wp-admin/media-new.php`

* Drag to upload an image below 5MB in size

Note you get an HTTP error:

image

image


I see this notice in the logs:
`PHP Notice:  Undefined index: sizes in /plugins/classifai-for-wordpress/includes/Classifai/Providers/Azure/ComputerVision.php on line 96`

includes/Classifai/Providers/Azure/ComputerVision.php
```diff
@@ -60,7 +60,23 @@ public function register() {
...
+	 *
+	 * @since 1.4.0
+	 */
+	public function get_max_filesize() {
+		/**
+		 * Filters the ComputerVision maximum allowed filesize.
+		 *
+		 * @param int Default 4MB.
+		 */
+		return apply_filters( 'classifai_computervision_max_filesize', 4000000 ); // 4MB default.
```
helen: WordPress has some fun built-in constants for file sizes and time units, like `MB_IN_BYTES`. Would advise using that here for legibility and technical accuracy.

includes/Classifai/Providers/Azure/ComputerVision.php
```diff
@@ -74,25 +90,78 @@ public function process_image( $metadata, $attachment_id ) {
...
+	 *
+	 * @since 1.4.0
+	 */
+	public function get_largest_acceptable_image_url( $full_image, $full_url, $sizes ) {
+		$file_size = @filesize( $full_image ); // phpcs:ignore WordPress.PHP.NoSilencedErrors.Discouraged
+		if ( $file_size && $this->get_max_filesize() >= $file_size ) {
+			return $full_url;
+		}
+
+		// Sort the image sizes in order of total width + height, descending.
```
helen: Clever!

Issue #75, opened 2019-06-07
Expand upon Azure Computer Vision for image classification


jeffpaul:
**Is your enhancement related to a problem? Please describe.**
Expand upon the initial work done in #51 and add additional features from Computer Vision.

**Describe the solution you'd like**
- [x] Automated tagging
- [x] All automated tasks available to turn on/off via Settings
- [x] UI to add tags alongside automated ones
- [x] ability to remove automated tags

**Out of scope features**
~- [ ] hook to catch the alt text being sent back from the Azure API (in case these need to be presented/filtered by specific consumer integrations)~ {moved to #138}
~- [ ] UI for filtering in the Media Library (currently in backbone.js) and then batch sending images for alt text classifying~ {moved to #129}
~- [ ] error messaging or hook available to catch when ClassifAI doesn't match something for an image's alt text~ {moved to #139}
~- [ ] Automated captions~ {moved to #140}
~- [ ] Automated descriptions~ {moved to #141}

**Designs**
n/a

**Describe alternatives you've considered**
none

**Additional context**
Related to #51.

rickalee:
Would users have UI to to add their own tags along with automated ones?

jeffpaul:
@rickalee the actual UX / UI hasn't been finalized, but I'll make note of your functionalities in the list of todos in the issue description as those seem like reasonable use cases... thanks!

jeffpaul:
@rickalee if you take a look at @ryanwelcher's PR #125 you'll see that the tags does have a UI when you pull up the "Edit more details" part of an image in the Media Library.  That UI allows for users to add their own tags and remove automated ones.

helen:
@jeffpaul Do we want to close this on this milestone and split what's left out into a new issue, or something else?

jeffpaul:
@helen good call, done.

PR #134, opened 2019-09-19
Add image tags and alt text to search

helen:
### Description of the Change

Allow image searches to search tags and alt text. This is modeled after the core function `_filter_query_attachment_filenames()`

### Alternate Designs

n/a

### Benefits

More intuitive finding of your media.

### Possible Drawbacks

Slow queries. It would be a good place to suggest usage of Elasticsearch/ElasticPress if you find that this is becoming slow.

### Verification Process

Tested searches for unique keywords in alt text and generated tag meta.

### Checklist:

### Applicable Issues

See #75

### DIFF
--- a/includes/Classifai/Helpers.php
+++ b/includes/Classifai/Helpers.php
@@ -302,3 +302,33 @@ function get_feature_taxonomy( $feature ) {
 	 */
 	return apply_filters( 'classifai_taxonomy_for_feature', $taxonomy, $feature );
 }
+
+/**
+ * Filter the SQL clauses of an attachment query to include tags and alt text.
+ *
+ * @param array $clauses An array including WHERE, GROUP BY, JOIN, ORDER BY,
+ *                       DISTINCT, fields (SELECT), and LIMITS clauses.
+ * @return array The modified clauses.
+ */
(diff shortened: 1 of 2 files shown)


includes/Classifai/Helpers.php
```diff
@@ -302,3 +302,33 @@ function get_feature_taxonomy( $feature ) {
...
+
+	$clauses['where'] = preg_replace(
+		"/\({$wpdb->posts}.post_content (NOT LIKE|LIKE) (\'[^']+\')\)/",
+		'$0 OR ( classifai_postmeta.meta_value $1 $2 )',
+		$clauses['where']
+	);
+
+	return $clauses;
+}
+\add_filter( 'posts_clauses', __NAMESPACE__ . '\\filter_attachment_query_keywords' );
```
ryanwelcher: Do we need the extra `\` before the function name? I feel like we don't but I might just be under-caffeinated.

helen: No, I don’t think so, I was just trying to see if it would fix the linting error. Any ideas there @ryanwelcher? I don’t know why the file is autoloaded like that, I guess I could move this into a different file if that makes more sense.

PR #135, opened 2019-09-20
Process Existing Images with Computer Vision

ryanwelcher:
### Description of the Change
This PR introduces two features:

1. The ability to process an image from the associated attachment page.
2. The ability to process any image via the Media Modal.

### Benefits
These features will allow users to process existing images via the admin.

### Checklist:

### Applicable Issues

#75


ryanwelcher:
Meta box added that is tied into the post save. There are opt-in and are not tied to the automatic checkbox options in the settings.

Image without scanned data:

no-scanned-data

Image with scanned data:
with-scanned-data

ryanwelcher:
@jeffpaul I've implemented the buttons and associated endpoints to allow scanning in the media modal.

Screenshot:
scanning_buttons

includes/Classifai/Providers/Azure/ComputerVision.php
```diff
@@ -56,9 +56,69 @@ public function can_register() {
...
+	}
+
+	/**
+	 * Display meta data
+	 *
+	 * @param \WP_Post $post The post object.
+	 */
+	public function attachment_data_meta_box( $post ) {
+		$captions = get_post_meta( $post->ID, 'classifai_computer_vision_captions', true ) ? __( 'Rescan Captions', 'classifai' ) : __( 'Generate Captions', 'classifai' );
+		$tags     = get_post_meta( $post->ID, 'classifai_computer_vision_image_tags', true ) ? __( 'Rescan Tags', 'classifai' ) : __( 'Generate Tags', 'classifai' );
```
dinhtungdu: Delete alt text of an image > save > still show `Rescan Captions`, same for tags.
The captions and tags still remain in meta after delete, but the button labels are a little bit confused.

ryanwelcher: @dinhtungdu I'm not sure what you mean here :)
Are you suggesting that the text of the buttons should be dependent on if there is existing data or not?

ryanwelcher:
@dinhtungdu thanks for the review! I've addressed the issues you pointed out and just had one question for you.

includes/Classifai/Providers/Azure/ComputerVision.php
dinhtungdu: @ryanwelcher I mean when the alt text is deleted, the label should say Generate instead of Rescan. Note that when we delete the alt text of an image, the `classifai_computer_vision_captions` remains unchanged. This is a minor issue btw.

ryanwelcher:
@dinhtungdu this is ready for review. Thanks!

src/js/media.js
```diff
@@ -20,6 +20,7 @@
 			button.removeAttribute( 'disabled' );
 			spinner.style.display = 'none';
 			spinner.classList.remove( 'is-active' );
+			button.textContent = 'Rescan';
```
dinhtungdu: look like we need localized string here

ryanwelcher: Updated

includes/Classifai/Providers/Azure/ComputerVision.php
ryanwelcher: I've updated the logic accordingly. Thanks for the great suggestion!

ryanwelcher:
@dinhtungdu I've added a couple of changes. This is ready for review.

jeffpaul:
@ryanwelcher noting there are a couple merge conflicts to resolve before we get this merged in, you able to handle those conflicts?

ryanwelcher:
@dinhtungdu I resolved the merge conflicts and pushed. Can you review so I can merge, please?

dinhtungdu:
I found a case when the API credentials are wrong or expired, the spinner will run forever without any feedback to users. In the media detail page, check `Generate caption` and hit `Update` (with wrong/expired API credentials), no feedback about the error after saving too. But due to our milestone, I think we can ship this PR and address the mentioned issue separately.

PR #144, opened 2019-09-25
Add a dev version notice

### LINKED ISSUE #66: Add dev version banner
ClassifAI is not actually runnable out of `develop`/`master`, so we should show a banner warning that you need to run `npm install && npm run build && composer install` for full functionality (if it doesn't straight up fatal without the vendor library, if it does we need to handle that more gracefully as well). We can take cues from Distributor for this, if using the `-dev` version suffix will need to update the build workflow to handle that as well.

helen:
### Description of the Change

Adds a dev version notice saying "You appear to be running a development version of ClassifAI. Certain features may not work correctly without regularly running `composer install && npm install && npm run build`. If you're not sure what this means, you may want to download and install the stable version of ClassifAI instead." This happens if the plugin update checker library is not present as expected in the vendor directory and only shows on ClassifAI pages (settings).

### Alternate Designs

Considered adding the `-dev` suffix to the version number and the whole nine yards with the build process like in Distributor but decided against it for 1.4. Let's do it for 1.5 though.

### Benefits

More immediate understanding of why the plugin might not be working the way you expect if you just clone from GitHub instead of downloading the built package as intended for end users.

### Possible Drawbacks

It will stop showing once you've run `composer install`, even though you are still on a dev version and need to be building with changes.

### Verification Process

Tested in the admin UI by renaming the `vendor` directory and navigating around.

### Checklist:

### Applicable Issues

Fixes #66

### DIFF
--- a/classifai.php
+++ b/classifai.php
@@ -158,7 +158,12 @@ register_activation_hook( __FILE__, 'classifai_activation' );
 classifai_autorun();
 
 // Include in case we have composer issues.
-require_once __DIR__ . '/vendor/yahnis-elsts/plugin-update-checker/plugin-update-checker.php';
+if ( file_exists( __DIR__ . '/vendor/yahnis-elsts/plugin-update-checker/plugin-update-checker.php' ) ) {
+	require_once __DIR__ . '/vendor/yahnis-elsts/plugin-update-checker/plugin-update-checker.php';
+} else {
+	add_action( 'admin_notices', 'classifai_dev_notice' );
+	add_action( 'network_admin_notices', 'classifai_dev_notice' );
+}
 
 if ( class_exists( 'Puc_v4_Factory' ) ) {
 	/*
@@ -184,5 +189,19 @@ if ( class_exists( 'Puc_v4_Factory' ) ) {
 			}
 		);
 		// @codingStandardsIgnoreEnd
+	}
+}
+
+/**
+ * Show dev version notice on ClassifAI pages if necessary.
+ */
+function classifai_dev_notice() {
+	if ( 0 === strpos( get_current_screen()->parent_base, 'classifai' ) ) {
+		?>
+		<div class="notice notice-warning">
+		<?php /* translators: %1$s: CLI install commands, %2$s: classifai url */ ?>
+		<p><?php echo wp_kses_post( sprintf( __( 'You appear to be running a development version of ClassifAI. Certain features may not work correctly without regularly running %1$s. If you&rsquo;re not sure what this means, you may want to <a href="%2$s">download and install</a> the stable version of ClassifAI instead.', 'distributor' ), '<code>composer install && npm install && npm run build</code>', 'https://classifaiplugin.com/' ) ); ?></p>
+		</div>
+		<?php
 	}
 }
--- a/classifai.php
+++ b/classifai.php
@@ -200,7 +200,7 @@ function classifai_dev_notice() {
 		?>
 		<div class="notice notice-warning">
 		<?php /* translators: %1$s: CLI install commands, %2$s: classifai url */ ?>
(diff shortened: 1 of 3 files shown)


helen:
Note to squash and merge this because of some junk commits :)

classifai.php
```diff
@@ -186,3 +191,17 @@ function( $plugin_info, $http_response = null ) {
...
+
+/**
+ * Show dev version notice on ClassifAI pages if necessary.
+ */
+function classifai_dev_notice() {
+	if ( 0 === strpos( get_current_screen()->parent_base, 'classifai' ) ) {
+		?>
+		<div class="notice notice-warning">
+		<?php /* translators: %1$s: CLI install commands, %2$s: classifai url */ ?>
+		<p><?php echo wp_kses_post( sprintf( __( 'You appear to be running a development version of ClassifAI. Certain features may not work correctly without regularly running %1$s. If you&rsquo;re not sure what this means, you may want to <a href="%2$s">download and install</a> the stable version of ClassifAI instead.', 'distributor' ), '<code>composer install && npm install && npm run build</code>', 'https://classifaiplugin.com/' ) ); ?></p>
```
adamsilverstein: The encoded apostrophe caught my eye here. At first I was going to say use `If you´re not sure` here instead of `If you&rsquo;re not sure`... instead thinking maybe `If you are not sure`?

adamsilverstein: This winds up outputting `Certain features may not work correctly without regularly running composer install && npm install && npm run build` which seems to imply you need to keep running composer install and npm install regularly. maybe just drop the word "regularly"?
```suggestion
		<p><?php echo wp_kses_post( sprintf( __( 'You appear to be running a development version of ClassifAI. Certain features may not work correctly without running %1$s. If you&rsquo;re not sure what this means, you may want to <a href="%2$s">download and install</a> the stable version of ClassifAI instead.', 'distributor' ), '<code>composer install && npm install && npm run build</code>', 'https://classifaiplugin.com/' ) ); ?></p>
```

helen: I copied this out of Distributor :) I guess I'm a dinosaur who still believes in encoding special characters?

helen: I suppose you don't need to run them regularly as in at a given interval or all that frequently but since this is really about people who are cloning from the GH repo and expecting things to work I would rather have the full initial build instructions. Dropping "regularly" seems fine, although hopefully it doesn't then imply you only have to do it once :)

helen: And the reason I prefer it that way and I think I will keep it that way is because it comes across as friendlier IMO, which important after an intimidating-sounding sentence.

adamsilverstein: 👍 nice!

classifai.php
```diff
@@ -186,3 +191,17 @@ function( $plugin_info, $http_response = null ) {
...
+
+/**
+ * Show dev version notice on ClassifAI pages if necessary.
+ */
+function classifai_dev_notice() {
+	if ( 0 === strpos( get_current_screen()->parent_base, 'classifai' ) ) {
+		?>
+		<div class="notice notice-warning">
+		<?php /* translators: %1$s: CLI install commands, %2$s: classifai url */ ?>
+		<p><?php echo wp_kses_post( sprintf( __( 'You appear to be running a development version of ClassifAI. Certain features may not work correctly without running %1$s. If you&rsquo;re not sure what this means, you may want to <a href="%2$s">download and install</a> the stable version of ClassifAI instead.', 'distributor' ), '<code>composer install && npm install && npm run build</code>', 'https://classifaiplugin.com/' ) ); ?></p>
```
jeffpaul: @helen should this reference `classifai` instead of `distributor` here?

helen: Ah dangit, I thought I caught all of them. Thanks.

helen:
```suggestion
		<p><?php echo wp_kses_post( sprintf( __( 'You appear to be running a development version of ClassifAI. Certain features may not work correctly without running %1$s. If you&rsquo;re not sure what this means, you may want to <a href="%2$s">download and install</a> the stable version of ClassifAI instead.', 'classifai' ), '<code>composer install && npm install && npm run build</code>', 'https://classifaiplugin.com/' ) ); ?></p>
```

Issue #143, opened 2019-09-25
Add service provider request response info to Site Health Info screen


jeffpaul:
**Is your enhancement related to a problem? Please describe.**
Building on the work in #108, let's add the results of the most recent request to each provider as this will likely help when troubleshooting service provider-related issues.

**Describe the solution you'd like**
- [ ] store latest request response per service provider in options
- [ ] translate request response into human-ish terms in Site Health Info screen

**Designs**
n/a

**Describe alternatives you've considered**
none

**Additional context**
n/a

PR #149, opened 2019-10-14
Add smart image cropping with Computer Vision

### LINKED ISSUE #112: Integrate Azure Computer Vision for smart image cropping
Azure's [Computer Vision](https://docs.microsoft.com/en-us/rest/api/cognitiveservices/computervision/generatethumbnail/generatethumbnail) is able to use smart cropping when generating thumbnails for images.  This could be useful on sites utilizing images that don't always crop "nicely" via existing image size generation scripts.  Smart cropping ensures that the most relevant parts of your uploaded image are retained on the various image sizes generated.

**Designs**
n/a

none

**Additional context**
none

johnwatkins0:
This PR introduces smart image cropping provided by Computer Vision. 

It would resolve #112. I think #135 could integrate smart cropping without much additional effort.

### Description of the Change

Sites are opted into the service through a checkbox on the image processing settings page. As an image is uploaded, each intermediate image size with cropping generates a request to the external service. If a cropped image is successfully returned, that image is saved and added to the image metadata in place of the version generated by WordPress (which remains in the file system).

Screen Shot 2019-10-13 at 8 10 05 PM

Screen Shot 2019-10-13 at 8 11 07 PM

Screen Shot 2019-10-13 at 8 11 51 PM

### Alternate Designs

- I considered deleting the original non-smart-cropped file when Computer Vision successfully returns a cropped image. This would save server space. I'm still on the fence about it.
- I also considered adding a checkbox group on the settings page allowing the feature to be opted into on the basis of specific image sizes, but this felt too complex. I did introduce a couple of filters that could be used to turn off smart-cropping specific image sizes.

### Benefits

Thumbnails are nicely cropped around faces and other focal points.

### Possible Drawbacks

A separate request needs to be made for each cropped image size. Only one core image size is cropped by default, but themes/plugins can register any number of cropped sizes. I think at some point the external requests around image processing are going to need to be done asynchronously rather than during pageload execution. 

### Verification Process

**This only works on a publicly accessible site.**

1. On the "Image Processing" Classifai settings page, with credentials in place, check the "Enable smart cropping" checkbox.
2. Upload a file to the media library. Useful files for testing are large portraits of people or similar photos in which the main focal point is significantly off center. The 150x150 thumbnail size is good for testing.
3. Upload the same file with the "Enable smart cropping" checkbox unchecked, and note the difference.
4. You might need to try a few files before you see a big difference.

### Checklist:


includes/Classifai/Helpers.php
```diff
@@ -302,3 +302,70 @@ function get_feature_taxonomy( $feature ) {
...
 }
+
+/**
+ * Provides the max filesize for the Computer Vision service.
+ *
+ * @since 1.4.0
+ *
+ * @return int
+ */
+function computer_vision_max_filesize() {
```
johnwatkins0: Moved these functions out of ComputerVision class because they're now called from multiple places and don't use ComputerVision class state.

includes/Classifai/Providers/Azure/SmartCropping.php
```diff
@@ -0,0 +1,289 @@
...
+		/**
+		 * Fires when the generateThumbnail smart-cropping API response did not have a 200 status code.
+		 *
+		 * @since 1.5.0
+		 *
+		 * @param array|WP_Error Response data or a WP_Error if the request failed.
+		 * @param string The request URL with query args added.
+		 * @param array  Array containing the image height and width.
+		 */
+		do_action( 'classifai_smart_cropping_unsuccessful_response', $response, $url, $data );
```
johnwatkins0: Can use the failed response for https://github.com/10up/classifai/issues/143

jeffpaul:
In regards to **Alternate Designs** 1: "I considered deleting the original non-smart-cropped file when Computer Vision successfully returns a cropped image. This would save server space. I'm still on the fence about it.", I _think_ I'm on the side of the fence of deleting the original non-smart-cropped file as I'm not sure there would be an easy way for a site owner to swap that back into place of the smart-cropped file and its not as though we're deleting the main source image itself.

jeffpaul:
In regards to **Alternate Designs** 2: "I also considered adding a checkbox group on the settings page allowing the feature to be opted into on the basis of specific image sizes, but this felt too complex. I did introduce a couple of filters that could be used to turn off smart-cropping specific image sizes.", I agree with the lesser complexity of the single checkbox on the settings page and that providing filters seems the better route for flexibility.

rickalee:
Hi team,

Did some initial tests with Azure Blob Storage to ensure that images could be smart cropped even when offloaded to external service. 

Good news: Initial images under 4MB get smart cropped with Azure Blob Storage

Smart Cropped Camera

If the initial image is over 4MB and needs to go through @filesize to check server side file size it does not get smart cropped. 

Not Smart Cropped

cc @johnwatkins0 @jeffpaul

jeffpaul:
@rickalee thanks for the testing and feedback... any immediate advice for @johnwatkins0 on tweaks to ensure that images that go through the 4mb file size checks properly get through the smart cropping?

rickalee:
Add a check for file_exists when using get_attached_file() if not try wp_get_attachment_url()

Note: With Azure Blob Storage we sometimes proxy example.com/wp-content/uploads/ to Azure Blob so URLs returned from wp_get_attachment_url() may not obviously be Azure Blob. One alternative would be to add a filter to response headers to determine source e.g. `x-ms-blob-type: AppendBlob`

johnwatkins0:
Thanks, @rickalee . For testing purposes, do you use https://github.com/10up/windows-azure-storage to integrate Azure storage, or something?

rickalee:
@johnwatkins0 That plugin is correct.

johnwatkins0:
Hi, @rickalee, after taking a closer look, I think this can be solved easily. I am weighing a few options, though, and would welcome your thoughts (also CC the reviewers on this PR).

This issue is not with cropping in particular but rather the `get_largest_acceptable_image_url` function that is now used for two different Computer Vision functions (and incidentally was copied into another 10up plugin recently). The function checks the full image size and, if it's too large, walks through the intermediate images and uses the biggest one under the limit. 

For cases of remote storage, we need an alternative. Here's what I'm considering:

1. Don't provide the alternative functionality in this plugin, but provide a filter before `get_largest_acceptable_image_url` runs. Document the 4MB limit and point plugin/theme developers to the filter (e.g., in a README section specifically addressing remote storage). If the filter returns a file, run the remote request and if it fails then it fails.
2. Create an alternative function handling the remote storage case, and provide a filter allowing other theme/plugin developers to specifically opt in at the code level. Opting in would be required.
3. Do #2, but with a settings checkbox instead of a filter.
4. Create the alternative function and find a way to detect remote storage automatically. This could get complicated, because there are so many possible configurations, and multiple popular plugins handling remote storage. If we go with this option, I would suggest (cc @jeffpaul ) that this be a separate issue tackled later.

Thanks!

johnwatkins0:
Good idea, @rickalee . I'll see if that can be the solution for 5.3+, with the filter for pre-5.3.

johnwatkins0:
@rickalee , good news: The new `wp_create_image_subsizes` function in 5.3 is applied to image metadata before the `wp_generate_attachment_metadata` filter runs, so that will help with both Smart Cropping and alt generation with Computer Vision in this plugin. Unless the big image size is overridden, all images should be below the limit by the time they get here.

For pre-5.3, I added this filter that allows a URL to be passed, which will bypass the `largest_acceptable_image` function:
https://github.com/10up/classifai/blob/48eed5510371e81c246f898fc206f75267a3d538/includes/Classifai/Providers/Azure/SmartCropping.php#L168-L186

Do you think this will work? Let me know if you see any other potential pitfalls around the Azure integration. 

@jeffpaul I also added a commit to make smart-cropped images overwrite their WP-generated equivalents (with another filter to change the smart cropped file name and keep the original in place). https://github.com/10up/classifai/pull/149/commits/b8994868aadbbaf02b845af16c2aa6b3ba99308a

rickalee:
Hi team,

Additional testing showing that even images below 4MB are not being sent through Computer Vision API.

I believe it may be related to access of file system and when images are offloaded. https://github.com/10up/classifai/blob/48eed5510371e81c246f898fc206f75267a3d538/includes/Classifai/Providers/Azure/SmartCropping.php#L232

If Azure Blob Storage option: "Keep local files after uploading them to Azure Storage." is enabled. Images are cropped but get odd JSON responses of "The response is not a valid JSON response." in editor/console. 

Will make time on Thursday to dig into this more.

johnwatkins0:
Thanks, @rickalee. That should be enough information for me to replicate it in my test environment with the Azure storage plugin. I'll let you know how it goes.

johnwatkins0:
Hi, @rickalee . I did some more testing with Azure Storage turned on. At first I was getting some errors, which I found to be caused by attempting to upload images with width/height greater than 1024px, which are forbidden: https://docs.microsoft.com/en-us/rest/api/cognitiveservices/computervision/generatethumbnail/generatethumbnail#uri-parameters

After resolving that in https://github.com/10up/classifai/pull/149/commits/e81cc4b3119a96f4ee3d848827974960dc0fa572, I'm not having errors anymore. Images are uploading to Azure storage and coming back with smart-cropping. E.g., this image that I uploaded is easy to verify as  smart cropped because WP crops it much differently: https://azurethumbs.blob.core.windows.net/images/2019/10/Petrol_station_inBe%CC%81nin-crop-300x300.jpg

Is it possible the error you were experiencing was the 1024px error fixed by my last commit? If not, I'd very much appreciate replicable steps for any further issues you're having. Thanks for sticking with me on this 😃

jeffpaul:
@Ritesh-patel would be good to get your review on this as well based on your prior experience with smart cropping, so if you're able to review John's work that'd be awesome!

johnwatkins0:
@jeffpaul The need for smart cropping in WordPress has come up in some client conversations lately, including in one fast-moving project with a late-winter launch date. Is anything I can do to help move this along and maybe get it into the plugin in time?

jeffpaul:
@johnwatkins0 I'm hoping that we can finish up some other work here in January so that we shift focus to getting the ClassifAI 1.5 release out in February.  Is that timing too slow for what you're hoping for?

@rickalee have you had a chance to review John's latest updates to see if they play well with Azure blob storage such that you can 👍 this PR?

johnwatkins0:
Thanks, @jeffpaul . February would give us plenty of time.

includes/Classifai/Providers/Azure/SmartCropping.php
```diff
@@ -0,0 +1,338 @@
...
+				$size_data,
+				computer_vision_max_filesize()
+			);
+		}
+
+		if ( empty( $url ) || empty( $size_data ) || ! is_array( $size_data ) ) {
+			return false;
+		}
+
+		$data = array(
```
Ritesh-patel: We should change it to php 5 array.

dinhtungdu:
In my test, the PR works great without Azure Storage. But when I use Azure storage, there are still some issues:

- The thumbnail (150x150) doesn't get smart cropped.
- The `alt` doesn't generate with `Keep local file` unchecked.

includes/Classifai/Providers/Azure/ComputerVision.php
```diff
@@ -48,32 +51,60 @@ public function can_register() {
...
 		return true;
 	}
 
 	/**
 	 * Register the functionality.
 	 */
 	public function register() {
-		add_filter( 'wp_generate_attachment_metadata', [ $this, 'process_image' ], 10, 2 );
+		add_filter( 'wp_generate_attachment_metadata', [ $this, 'smart_crop_image' ], 10, 2 );
+		add_filter( 'wp_generate_attachment_metadata', [ $this, 'generate_image_alt_tags' ], 10, 2 );
```
dinhtungdu: In my test, changing these filters priority to `8` get image tagging and smart cropping work with Azure Storage which hooks to the same filter with priority set to `9`.

jeffpaul:
@johnwatkins0 are you able to work through the feedback from @Ritesh-patel and @dinhtungdu?  If not, I'll see if someone else can help wrap up these so the PR is ready for the v1.5 release.

johnwatkins0:
@jeffpaul Yes, I can work on it over the next few days.

johnwatkins0:
Thanks, @dinhtungdu . I have pushed the suggested changes. In my test environment using Azure storage I have uploaded an image and verified that the 150x150 thumb is smart-cropped and that the alt text is generated. Let me know if you find any other issues.

includes/Classifai/Helpers.php
```diff
@@ -302,3 +302,70 @@ function get_feature_taxonomy( $feature ) {
...
+ *
+ * @since 1.4.0
+ *
+ * @param string $full_image The path to the full-sized image source file.
+ * @param string $full_url   The URL of the full-sized image.
+ * @param array  $sizes      Intermediate size data from attachment meta.
+ * @param int    $max        The maximum acceptable size.
+ * @return string|null The image URL, or null if no acceptable image found.
+ */
+function get_largest_acceptable_image_url( $full_image, $full_url, $sizes, $max = MB_IN_BYTES ) {
```
dinhtungdu: This function return `NULL` with Azure storage enabled and configured. Tested with publicly accessible WP install and newly uploaded images after configuring Azure.

includes/Classifai/Providers/Azure/ComputerVision.php
```diff
@@ -84,18 +115,19 @@ public function get_max_filesize() {
 	 *
 	 * @return mixed
 	 */
-	public function process_image( $metadata, $attachment_id ) {
+	public function generate_image_alt_tags( $metadata, $attachment_id ) {
```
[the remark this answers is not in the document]
ryanwelcher: @dinhtungdu just to add some context here. This function handles both items because the results of `scan_image` return both the captions and tags. The actual creation for tags and alts tags are done by separate functions called by this function. This is a long-winded way of saying that we save another call to the service by doing this all at once. 

I'd be for a name change to the function to better reflect what it does though :)

tests/Classifai/HelpersTest.php
```diff
@@ -135,4 +145,97 @@ function test_it_knows_configured_feature_taxonomies() {
...
+
+		$url = get_largest_acceptable_image_url(
+			get_attached_file( $attachment ),
+			wp_get_attachment_url( $attachment, 'full' ),
+			wp_get_attachment_metadata( $attachment )['sizes'],
+			computer_vision_max_filesize()
+		);
+		$this->assertNull( $url );
+
+		remove_filter( 'classifai_computer_vision_max_filesize', $set_1kb_max_filesize );
```
ryanwelcher: Small nitpick here, if the test fails, this filter won't be removed. I'd recommend moving it up before the `$this->assertNull` call.

ryanwelcher:
@jeffpaul @dinhtungdu I think we can land this feature now and then focus on the Azure Storage integration as an immediate followup. Any concerns with that approach?

dinhtungdu:
@ryanwelcher No concern, I agree!

PR #189, opened 2020-02-27
Add/hookdocs workflow

jeffpaul:
### Description of the Change

This PR introduces the files needed to [generate hook docs and deploy them to GitHub Pages](https://github.com/10up/actions-wordpress/blob/master/hookdocs-workflow.md) for ClassifAI, similar to what was created for [Distributor](https://10up.github.io/distributor/).

### Alternate Designs

n/a

### Benefits

More fully communicates the extensibility of ClassifAI via a hook documentation site.

### Possible Drawbacks

We may not have JSDoc-formatted docblocks for our hooks, in which case we'll want to ensure those get added as well so that the hookdoc site is as complete and accurate as possible.

### Verification Process

Manually compared to set up for hookdoc generation on Distributor.

### Checklist:

### Applicable Issues

n/a, mea culpa.

### Changelog Entry

```
### Added
- Fully documented hooks and filters via new Hook Documentation site
```

### DIFF
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -3,7 +3,7 @@ name: Build Hook Docs
 on:
  push:
    branches:
-    - develop
+    - master
 
 jobs:
   build:
--- a/.github/workflows/build-docs.yml
+++ b/.github/workflows/build-docs.yml
@@ -1,0 +1,28 @@
+name: Build Hook Docs
+
+on:
+ push:
+   branches:
+    - develop
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v1
+    - name: Use Node.js 10
+      uses: actions/setup-node@v1
+      with:
+        node-version: '10.x'
+    - name: npm install, and build docs
+      run: |
+        npm install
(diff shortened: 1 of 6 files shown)


.github/workflows/build-docs.yml
```diff
@@ -0,0 +1,28 @@
+name: Build Hook Docs
+
+on:
+ push:
+   branches:
+    - develop
```
helen: Let's change this to `master` so that the hook docs don't update until we actually are ready to release a new version

package.json
```diff
@@ -11,6 +11,7 @@
     "test": "./vendor/bin/phpunit",
     "tests:e2e": "./vendor/bin/wpacceptance run",
     "makepot": "wpi18n makepot && echo '.pot file updated'"
```
helen: Missing comma :)

```suggestion
    "makepot": "wpi18n makepot && echo '.pot file updated'",
```

helen (changes requested):

.github/workflows/build-docs.yml
jeffpaul: @helen should we make a similar change to Distributor and ElasticPress?

helen (approved):

PR #192, opened 2020-03-05
Move doc deploy action to the release workflow

helen:
### Description of the Change

Moves the doc build/deployment to the same workflow as the release to stable, since both run on master and are a signal of an impending new release. Realized this as I was changing the doc workflow to go back to running on master instead of develop. Also fixes a warning about `version` being deprecated in favor of `node-version`.

### Alternate Designs

Leaving as-is, which is more resource intensive as it spins up two machines and runs `npm install` on both.

### Benefits

Faster workflow run, less maintenance.

### Possible Drawbacks

If we decide later they shouldn't run on the same trigger we'll have to split it back out.

### Verification Process

n/a, basically hoping it'll work.

### Checklist:

- [ ] I have read the [**CONTRIBUTING**](https://github.com/10up/classifai/blob/develop/CONTRIBUTING.md) document.

### Applicable Issues

See #191 

### Changelog Entry

n/a


jeffpaul (approved):
Looks good to me, good thinking on combining these as well!

PR #206, opened 2020-03-22
Prevent PHP notice if Watson settings have not been saved yet

barryceelen:
### Description of the Change

Add check to the `use_username_password()` method  in the NLU class that verifies whether the `watson_username` setting is defined before accessing it.

### Benefits

If the setting has not been saved before, the plugin triggers a PHP notice without this additional check.

### Verification Process

The function now only returns `true` if the setting has an actual value and is 'apikey'.


dinhtungdu (approved):
LGTM 👍

PR #207, opened 2020-03-23
Fix Computer Vision settings when saving for the first time

barryceelen:
### Description of the Change

This update fixes the Microsoft Computer Vision settings page checkbox options always being saved  with the value '1' if the setting had not yet been saved before.

The `sanitize_settings()` callback sets each checkbox value to:
- `no` if the option is empty (checkbox unchecked)
- `1` if the option is not empty (checkbox checked)

This logic leads to trouble when the `sanitize_settings()` callback is called twice, which happens when saving the settings for the first time. When saving a setting for the first time the `update_option()` and the `add_option()` functions are called which both run the sanitize callback.

This update also contains a minor refactor of the way the settings page error notifications are registered. It prevents the same error being rendered twice when the callback is called multiple times.

### Benefits

The settings are saved correctly when saving the options for the first time.

### Verification Process

To verify this update:

Before applying this update:

- Delete the `classifai_computer_vision` option
- Go to Image Processing settings page
- Save options
- The "Automatically Caption Images", "Automatically Tag Images" and "Enable smart cropping" checkboxes are all checked

After applying this update:
- Delete the `classifai_computer_vision` option
- Go to Image Processing settings page
- Save options
- The "Automatically Caption Images", "Automatically Tag Images" and "Enable smart cropping" checkboxes are checked or unchecked correctly


includes/Classifai/Providers/Azure/ComputerVision.php
```diff
@@ -546,8 +545,22 @@ public function sanitize_settings( $settings ) {
 			$new_settings['image_tag_taxonomy'] = $settings['image_tag_taxonomy'];
 		}
 
-		$smart_cropping_enabled                = isset( $settings['enable_smart_cropping'] ) ? '1' : 'no';
-		$new_settings['enable_smart_cropping'] = $smart_cropping_enabled;
+		if ( ! empty( $settings_errors ) ) {
+
+			$registered_settings_errors = wp_list_pluck( get_settings_errors( $this->get_option_name() ), 'code' );
+
+			foreach ( $settings_errors as $k => $v ) {
```
dinhtungdu: Can we be more explicit here for variables' names? `$code` and `$message` are easier to understand than `$k` and `$v`.

dinhtungdu (changes requested):
Great work! The checkboxes are checked and saved correctly.

I think we should investigate the duplicated error message for the first-time saving in another issue/PR because the Waston setting page has the same problem. We may add more providers and settings in the future, so it's better to fix that issue permanently. Also note that, with the suggested approach in this PR, we need to get and compare setting errors before adding new errors every time we save the setting; I don't think our setting will go super complex so it won't affect the performance.

FYI, about the sanitize callback running twice issue, there is a ticket on [Trac](https://core.trac.wordpress.org/ticket/21989) about it which hasn't had a clear decision for more than 7 years.

includes/Classifai/Providers/Azure/ComputerVision.php
barryceelen: Good call @dinhtungdu, [fixed](https://github.com/10up/classifai/pull/207/commits/d705dd2849cecb64d8568c3ed451870035a84ef7)

dinhtungdu (approved):

PR #209, opened 2020-03-27
Update JavaScript build process to work better with Gutenberg

ryanwelcher:


.eslintrc
```diff
@@ -0,0 +1,11 @@
+{
+	"extends": "@10up/eslint-config",
```
nicholasio: @ryanwelcher Can you try using the new eslint config? https://github.com/10up/eslint-config/pull/17 It's still being worked on but you can already install it with `npm install @10up/eslint-config@next` there's a Gutenberg specific config

Any feedback is appreciated.

src/js/admin.js
```diff
@@ -1,20 +1,23 @@
...
-	const $userFieldWrapper = $userField.closest( 'tr' );
-	const [$passwordFieldTitle] = document.getElementById( 'classifai-settings-watson_password' ).closest( 'tr' ).getElementsByTagName( 'label' );
+	const $userFieldWrapper = $userField.closest('tr'); // eslint-disable-line
+	const [$passwordFieldTitle] = document // eslint-disable-line
+		.getElementById('classifai-settings-watson_password')
+		.closest('tr')
+		.getElementsByTagName('label');
 
-	if ( null === $toggler ) return;
+	if ($toggler === null) return;
```
nicholasio, line 14: This if check is actually not needed. Line 6 already takes care of it.

nicholasio: This if check is not needed. See Line 6

src/js/admin.js
```diff
@@ -1,20 +1,23 @@
...
+	const $toggler = document.getElementById('classifai-waston-cred-toggle');
+	const $userField = document.getElementById('classifai-settings-watson_username');
 
-	if ( null === $toggler || null === $userField ) return;
+	if ($toggler === null || $userField === null) return;
 
-	const $userFieldWrapper = $userField.closest( 'tr' );
-	const [$passwordFieldTitle] = document.getElementById( 'classifai-settings-watson_password' ).closest( 'tr' ).getElementsByTagName( 'label' );
+	const $userFieldWrapper = $userField.closest('tr'); // eslint-disable-line
+	const [$passwordFieldTitle] = document // eslint-disable-line
```
nicholasio: If you remove Line 14 you don't need to ignore these two lines.

src/js/editor.js
```diff
@@ -1,23 +1,29 @@
-
-const { subscribe, select, dispatch } = wp.data;
+import { subscribe, select, dispatch } from '@wordpress/data'; // eslint-disable-line
```
nicholasio: I've updated the eslint config to account for "@wordpress" imports.

.eslintrc
```diff
@@ -0,0 +1,11 @@
+{
+	"extends": "@10up/eslint-config/gutenberg",
```
nicholasio: Change to `@10up/eslint-config/wordpress`

package.json
```diff
@@ -30,19 +30,34 @@
     "role": "developer"
   },
   "devDependencies": {
-    "@10up/eslint-config": "^1.0.9",
+    "@10up/eslint-config": "^2.0.0-next",
```
nicholasio: Update to `2.0.2-next`

nicholasio (changes requested):
@ryanwelcher I updated the eslint config to address some of the issues here. I've also renamed the config name from `10up/eslint-config/gutenberg` to `10up/eslint-config/wordpress` as it makes more sense.

helen:
Could I get a summary here about why we're using the 10up config instead of `wp-scripts`? I am not strongly opinionated except that I am not sure why there is a WP-specific config inside 10up's package when a WP one already exists.

jeffpaul:
@helen @ryanwelcher @nicholasio @dinhtungdu @dkotter given that we're adding a Gutenberg integration as part of #228 do we want to revisit this PR and bring into the 1.6.0 release if it'll help in relation to the work in 228?

dinhtungdu:
@jeffpaul I prefer to keep this one for 1.7 because it's not a blocker for 228.

ryanwelcher:
@dinhtungdu @jeffpaul based on the changes to our tooling since this PR was created, I would probably recommend removing the custom webpack config in favor of using everything provided by [10up-toolkit](https://github.com/10up/10up-toolkit). 

This is the [standard approach for new projects ](https://github.com/10up/wp-scaffold/blob/trunk/themes/10up-theme/package.json)and provides all of the linting tools we'd need here. 

Additionally, [we can leverage the package.json file](https://github.com/10up/wp-scaffold/blob/trunk/themes/10up-theme/package.json#L23) to manage customizations which means that we won't need a custom webpack config file

helen:
I'm closing this out for now, it's so out of date we are better off starting over again. I am not yet sure about 10up-toolkit for OSS WP plugins, as there are differences between something that happens to be openly available but still geared toward client work, and what we want to represent in a community project. We need to take a look at this across 10up plugins and then go from there :)

Issue #201, opened 2020-03-09
Resolve compatibility issue with Microsoft Azure Storage


jeffpaul:
**Describe the bug**
As an offshoot from testing on #149, we identified an [issue with compatibility with Azure blog storage](https://github.com/10up/classifai/pull/149#pullrequestreview-349112596).  This subsequent issue will track proper resolution for that issue.

**Steps to Reproduce**

**Expected behavior**

**Screenshots**

**Environment information**
 - Device: 
 - OS: 
 - Browser and version: 
 - ClassifAI version: 
 - Theme and version: 
 - Other installed plugin(s) and version(s): 

**Additional context**

jeffpaul:
@rickalee are you able to test and validate whether #217 resolves this issue or if this is still a concern that needs to be resolved?

PR #217, opened 2020-07-17
Allow filtering of the image URL prior to generating image tags and alternate text

petenelson:
### Description of the Change
In instances where an image has been offloaded to a CDN, running the `image` CLI command is failing. This command calls `generate_image_alt_tags()`, which calls `get_largest_acceptable_image_url()`, which requires an image to be physically present on the server. If the image is not present, the image URL is null and no tags or alternate text can be generated. 

### Alternate Designs

None

### Benefits

This pull request adds a filter to allow overriding of the image URL which can allow the Computer Vision API to pull the image directly from a CDN.

### Possible Drawbacks

None

### Verification Process

No verification was performed.

### Checklist:

- [X] I have read the [**CONTRIBUTING**](/CONTRIBUTING.md) document.
- [X] My code follows the code style of this project.

### Changelog Entry

* Added `classifai_generate_image_alt_tags_source_url` filter to allow overriding of the image URL within `generate_image_alt_tags()`

### DIFF
--- a/includes/Classifai/Providers/Azure/ComputerVision.php
+++ b/includes/Classifai/Providers/Azure/ComputerVision.php
@@ -187,6 +187,9 @@ class ComputerVision extends Provider {
 			'no' !== $settings['enable_image_tagging'] ||
 			'no' !== $settings['enable_image_captions']
 		) {
+
+			$image_url = null;
+
 			if ( isset( $metadata['sizes'] ) && is_array( $metadata['sizes'] ) ) {
 				$image_url = get_largest_acceptable_image_url(
 					get_attached_file( $attachment_id ),
@@ -197,6 +200,8 @@ class ComputerVision extends Provider {
 			} else {
 				$image_url = wp_get_attachment_url( $attachment_id, 'full' );
 			}
+
+			$image_url = apply_filters( 'classifai_generate_image_alt_tags_source_url', $image_url, $attachment_id );
 
 			if ( ! empty( $image_url ) ) {
 				$image_scan = $this->scan_image( $image_url );
--- a/includes/Classifai/Providers/Azure/ComputerVision.php
+++ b/includes/Classifai/Providers/Azure/ComputerVision.php
@@ -188,20 +188,20 @@ class ComputerVision extends Provider {
 			'no' !== $settings['enable_image_captions']
 		) {
 
-			$image_url = null;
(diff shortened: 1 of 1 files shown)


jeffpaul:
This may resolve #201, but testing will be needed to be certain.

includes/Classifai/Providers/Azure/ComputerVision.php
```diff
@@ -187,15 +187,20 @@ public function generate_image_alt_tags( $metadata, $attachment_id ) {
...
-					wp_get_attachment_url( $attachment_id, 'full' ),
-					$metadata['sizes'],
-					computer_vision_max_filesize()
-				);
-			} else {
-				$image_url = wp_get_attachment_url( $attachment_id, 'full' );
+
+			$image_url = apply_filters( 'classifai_generate_image_alt_tags_source_url', null, $attachment_id );
+
+			if ( ! empty( $image_url ) ) {
```
dinhtungdu: This should check if `$image_url` is empty or an invalid URL.
```suggestion
			if ( empty( $image_url ) || ! filter_var( $image_url, FILTER_VALIDATE_URL ) ) {
```

dinhtungdu (changes requested):

includes/Classifai/Providers/Azure/ComputerVision.php
```diff
@@ -187,15 +187,20 @@ public function generate_image_alt_tags( $metadata, $attachment_id ) {
...
-					wp_get_attachment_url( $attachment_id, 'full' ),
-					$metadata['sizes'],
-					computer_vision_max_filesize()
-				);
-			} else {
-				$image_url = wp_get_attachment_url( $attachment_id, 'full' );
+
+			$image_url = apply_filters( 'classifai_generate_image_alt_tags_source_url', null, $attachment_id );
+
+			if ( ! empty( $image_url ) || ! filter_var( $image_url, FILTER_VALIDATE_URL ) ) {
```
dinhtungdu: This should be `empty( $image_url )` but not `! empty( $image_url )`.

Currently, `$image_url` will be `null` if `classifai_generate_image_alt_tags_source_url` is not used. Besides, if we use that filter, the function will get the image URL again.

dinhtungdu (changes requested):

dinhtungdu (approved):
LGTM 👍

PR #228, opened 2020-10-06
Azure Image OCR

### LINKED ISSUE #111: Integrate Azure Computer Vision for OCR text generation for uploaded files
Azure's [Computer Vision](https://azure.microsoft.com/en-us/services/cognitive-services/computer-vision/) is able to read both printed and handwritten text in images.  One potential use case here would be for academics who tend to upload large PDF files or documents to WordPress for storage, ClassifAI could then utilize OCR to generate a related Post or Page for each PDF/document uploaded.

**Designs**
n/a

none

**Additional context**
none

dkotter:
### Description of the Change

Add support for [Azure image OCR](https://westcentralus.dev.cognitive.microsoft.com/docs/services/computer-vision-v3-ga/operations/56f91f2e778daf14a499f20d).

Adds a new setting that is turned off by default but can be turned on to run OCR scanning on images. When an image is initially uploaded, we run a few checks on that image. First, we check if the image matches one of the supported types (JPEG, PNG, GIF, BMP). If so, we then check if we have a previous image scan and if so, we check if we have either the handwriting or text tags set, with a high confidence level (above 90%). If all those checks pass, we then run OCR scanning.

If we get a successful response back, we then parse the text out of that response, save that text to `wp_content` for that image and then save the full response to post meta.

We also add scan/recan functionality to the media modal and single media edit screens. If a scan is run from here, we bypass our checks, as we assume if someone is manually starting a scan, they don't care about the checks. ~~Also added some better error handling to the scan/rescan buttons, so if a scan fails, we don't continue to show the loading icon but instead remove that, keep the button disabled and change the button text to say error (ideally we would add even better error handling in a separate PR).~~ _This is now in a separate PR: #231_

_TODO: last piece is to add `describedby` text into the content when needed. Need to support the block editor and the classic editor_

### Benefits

Images with text can now have that text automatically read and then inserted into the content, with the proper `describedby` tags. This helps provide more context for images, especially images that are screenshots of text (like social media posts).

### Possible Drawbacks

With this feature turned on, will cause a slight slowdown (seems like roughly 5 extra seconds), in image processing.

### Checklist:

### Applicable Issues

#111


jeffpaul:
Noting here that @dinhtungdu is going to work on scaffolding in the Gutenberg bits

src/js/editor-ocr.js
```diff
@@ -0,0 +1,148 @@
...
+			return;
+		}
+
+		setBlockIndex( getBlockIndex( currentBlock.clientId ) );
+		setImageId( currentBlock.attributes.id );
+		openModal();
+	}, 10 ) );
+
+	return isOpen && <Modal title={__( 'ClassifAI detected text in your image' )}>
+		<p>{__( 'Would you like you insert it as a paragraph under you image block?' )}</p>
```
helen:
```suggestion
		<p>{__( 'Would you like you insert it as a paragraph under this image block?' )}</p>
```

Also, do these strings need to have the `classifai` textdomain?

helen:
This is feeling pretty good to me for a first run. I think what we might want to do is add a field to `wp_prepare_attachment_for_js()` that's a bool for whether `classifai_computer_vision_ocr` is non-empty (like `classifaiHasOcr` or something) and only show the prompt in that case, not just if the description field is populated because that description field could be populated manually or from unrelated EXIF data.

src/js/editor-ocr.js
```diff
@@ -0,0 +1,148 @@
...
+		}
+
+		return (
+			<Fragment>
+				<BlockEdit {...props} />
+				<InspectorControls>
+					<PanelBody title={__( 'ClassifAI' )} initialOpen={true}>
+						<PanelRow>
+							<Button onClick={() => insertDescription( clientId, attributes.id )} isPrimary>
+								{__( 'Insert description' )}
```
helen:
```suggestion
								{__( 'Insert scanned text into content' )}
```

Maybe more like this? I don't think being stored in the description field or being known as descriptive text is how you think of it in the editorial flow.

src/js/editor-ocr.js
dinhtungdu: My bad, I forgot the textdomain, added it. Thank you!

dinhtungdu:
@helen I addressed your comment by exposing `classifai_computer_vision_ocr` in REST API response.

dinhtungdu:
@helen in https://github.com/10up/classifai/pull/228/commits/4682dddf09475563505993e5c8bc0b9dc6244107, I replaced the sidebar button with a toolbar button. The UI is better, but it's not as descriptive and explicit as the sidebar button is. Or should we use text instead of icon, for example `Insert scanned text`?

dinhtungdu:
@helen I fixed the REST API response issue. This PR is working on my live site.

helen (approved):
Nice work here, everybody!

PR #231, opened 2020-10-16
Add better error handling for manual scanning of alt text or image tags

### LINKED ISSUE #230: When manually scanning for alt text or tags, if request fails, spinner never goes away
**Describe the bug**
Currently, if you use the manual scan/rescan buttons for image alt text and/or tags, as that process is working, the button is disabled and a spinner is shown. If that request completes successfully, that spinner is removed and the button is active again.

But if something goes wrong in that request, we currently don't handle that error. So the button stays disabled and the spinner stays showing, making it seem like things are still processing even though they aren't.

**Steps to Reproduce**
1. Setup the plugin
2. Disable either the Automatically Caption Images or Automatically Tag Images options (or both)
3. Select an existing image in the media library or upload a new one
4. With the image selected, in the media modal that opens, click the button that corresponds to the setting disabled above
5. See the spinner show and never go away

**Expected behavior**
If a request fails, at the very least, we should remove the loading state so the user knows things are done. Ideally, we would also show an error message.

dkotter:
### Description of the Change

Make sure our requests to generate alt text or image tags return a WP_Error object if something goes wrong. Currently we are returning those in a few instances but not in every instance.

In our javascript that makes these requests, if the response fails, check to see if we have error details and if so, show that error message. If we don't have error details, output a default error message. Also, make sure the spinner is removed so it doesn't appear like things are still processing.

### Alternate Designs

None

### Benefits

If a request fails, we provide feedback to the user about what went wrong.

### Possible Drawbacks

None

### Verification Process

1. Setup the plugin
2. Disable either the Automatically Caption Images or Automatically Tag Images options (or both)
3. Select an existing image in the media library or upload a new one
4. With the image selected, in the media modal that opens, click the button that corresponds to the setting disabled above. Ensure an error message is shown and the spinner goes away

Screen Shot 2020-10-19 at 12 32 37 PM

### Checklist:

### Applicable Issues

#230


src/js/media.js
```diff
@@ -1,41 +1,57 @@
...
+				( error ) => {
+					const errorObj = get( error, 'responseJSON', {
+						code: 'unknown_error',
+						message: __( 'An unknown error occurred.', 'classifai' ),
+					} );
+					spinner.style.display = 'none';
+					spinner.classList.remove( 'is-active' );
+					button.textContent = __( 'Error', 'classifai' );
+					errorContainer.style.display = 'inline-block';
+					errorContainer.textContent = errorObj.message;
```
dinhtungdu: How do users retry scanning after failing (due to network issue for example)?

dinhtungdu (commented):
I just have a question about retrying scanning, otherwise, this is looking good to me!

src/js/media.js
dkotter: @dinhtungdu Thanks for the review! Just pushed a fix for this, so if an error happens, we make sure the button is no longer disabled, so it can be clicked again

dinhtungdu (approved):
LGTM 🎉

PR #233, opened 2020-10-16
Ensure manual image scanning still works if the automatic scanning is turned off

### LINKED ISSUE #232: Rescan buttons don't work if automatic tagging is turned off
**Describe the bug**
In the settings for Image Processing, there are options to turn off/on automatic alt text or tags. If these options are turned off, the scan/rescan buttons are still shown when editing a single image but if you try and use them, they won't do anything (at the moment, the spinner will just keep showing forever, which is fixed in #231 and will instead show an error message).

I think even if the automatic options are disabled, we should still ensure the manual buttons work.

**Steps to Reproduce**
1. Setup the plugin
2. Turn off the settings for either Automatically Caption Images or Automatically Tag Images (or both)
3. Edit an existing image or upload a new one
4. Select that image and in the image modal that pops up, click the button that corresponds to the setting turned off above
5. Notice the spinner shows and never goes away and alt text or tags are never added

**Expected behavior**
The manual scan buttons should work even if the automatic options are disabled

dkotter:
### Description of the Change

If either the Automatically Caption Images or Automatically Tag Images settings are turned off, we still output the manual scan/rescan buttons when editing an individual image. But those buttons don't work, as we check the aforementioned settings when making a request to get image details and if those settings are disabled, we don't request the proper image details.

This PR fixes that, so if manual scanning is triggered, this overrides the automatic setting, so even if that setting is turned off, the manual scan will still work. Automatic scanning for newly uploaded images still respect that setting though, this only applies to manual scans.

### Alternate Designs

We could remove the manual scan buttons all together if the automatic scanning is turned off

### Benefits

Manual scanning works properly even if the automatic scanning is turned off. At the moment, those buttons show but won't work properly

### Possible Drawbacks

None

### Verification Process

1. Setup the plugin
2. Turn off the settings for either Automatically Caption Images or Automatically Tag Images (or both)
3. Edit an existing image or upload a new one
4. Select that image and in the image modal that pops up, click the button that corresponds to the setting turned off above
5. Processing should work correctly and the spinner should disappear and either alt text is added or tags are added (if either can be found)

### Checklist:

### Applicable Issues

#232

### DIFF
--- a/includes/Classifai/Providers/Azure/ComputerVision.php
+++ b/includes/Classifai/Providers/Azure/ComputerVision.php
@@ -15,7 +15,7 @@ class ComputerVision extends Provider {
 	/**
 	 * @var string URL fragment to the analyze API endpoint
 	 */
-	protected $analyze_url = '/vision/v1.0/analyze';
+	protected $analyze_url = 'vision/v1.0/analyze';
 
 	/**
 	 * ComputerVision constructor.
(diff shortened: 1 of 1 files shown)


dinhtungdu (approved):
Great work! This makes a lot of sense to me.

PR #235, opened 2020-10-16
Ensure we use the largest image size possible when scanning from the single edit view

### LINKED ISSUE #234: Scanning images on the edit view doesn't always work as expected
**Describe the bug**
There are two ways to scan/rescan image information after it has already been uploaded:
1. Through the buttons added to the media modal
2. Through checkboxes that can be set when editing a single image

The issue lies in the second approach. It seems when scan options are selected from a single image edit view, we pass the image thumbnail for processing, not a larger image size. This can lead to bad results, as image processing is more accurate with a larger image to review.

**Steps to Reproduce**
This will depend heavily on the image you're testing but I tested a text heavy image and got very different results from the media modal view compared to the single image edit view
1. Setup the plugin
2. Edit an existing image or upload a new one
3. Select that image and in the image modal that pops up, click the button to rescan for alt text
4. Take a note of what that text is
5. Now go to the single edit screen for that image
6. Check the box to rescan for alt text and save the image
7. Note the alt text is now different than what was previously returned (and typically, less accurate). Also note, sometimes the alt text isn't saved at all, since the text returned is below the threshold in the settings

**Expected behavior**
Results should be the same whether you are scanning from the media modal or the single edit view

dkotter:
### Description of the Change

Currently, when initiating a scan from the single edit screen, we are passing the image thumbnail to be processed. This can lead to unexpected results, results that don't match what we get on initial image processing or when scanning from the media modal view.

This PR fixes that to ensure we are passing the largest image size possible in this situation, to match the same image size we use in other scans.

### Alternate Designs

None

### Benefits

Scan results should now be the same no matter where they are initiated.

### Possible Drawbacks

None

### Verification Process
This will depend heavily on the image you're testing but I tested a text heavy image and got very different results from the media modal view compared to the single image edit view

1. Setup the plugin
2. Edit an existing image or upload a new one
3. Select that image and in the image modal that pops up, click the button to rescan for alt text
4. Take a note of what that text is
5. Now go to the single edit screen for that image
6. Check the box to rescan for alt text and save the image
7. When the page refreshes, the alt text should still be the same

### Checklist:

### Applicable Issues
#234

### DIFF
--- a/includes/Classifai/Providers/Azure/ComputerVision.php
+++ b/includes/Classifai/Providers/Azure/ComputerVision.php
@@ -112,7 +112,12 @@ class ComputerVision extends Provider {
 	 * @param int $attachment_id Post id for the attachment
 	 */
 	public function maybe_rescan_image( $attachment_id ) {
-		$image_url  = wp_get_attachment_image_url( $attachment_id );
+		$metadata   = wp_get_attachment_metadata( $attachment_id );
+		$image_url  = get_largest_acceptable_image_url(
+			get_attached_file( $attachment_id ),
+			wp_get_attachment_url( $attachment_id ),
+			$metadata['sizes']
+		);
 		$image_scan = $this->scan_image( $image_url );
 		if ( ! is_wp_error( $image_scan ) ) {
 			// Are we updating the captions?


dinhtungdu (dismissed):
LGTM

jeffpaul:
@dkotter if you can resolve the merge conflict, then we're good to merge this one in... thanks!

PR #245, opened 2020-10-22
Remove media from language processing settings

### LINKED ISSUE #242: Remove `Media` as option to select in Language Processing
In the Language Processing settings (for IBM Watson NLU) we present an option to classify `Media`.  As best I can tell NLU only works on text and not media.

LanugageProcessingSettings

Remove the option to try and classify `Media` (aka the `Attachment` post type) in the Language Processing section.

**Designs**

Keep as-is in case there's a legitimate case for Language Processing to run on Media/Attachments?

**Additional context**
n/a

dkotter:
### Description of the Change

In the Language Processing settings, we have a section to select what post types you want processed. This pulls all public post types, which will include the Attachment post type. While in theory you could run language processing on attachments (processing takes into account the `post_title` and `post_content`, which attachments should have a `post_title` and some may have `post_content`) attachments fire different hooks than other post types, which we aren't currently accounting for. So if you select the Media option in settings, it won't actually do anything.

This PR adds a new helper function to build up the list of post types we want in our Language Processing settings. This helper function runs the results through a new filter, so these results can be filtered if needed. This new function also removes attachments from that list, so we no longer show that as an option in our settings list.

### Alternate Designs

None

### Benefits

This removes a setting that currently won't actually do anything, potentially causing user confusion. Also introduces a new filter, allowing more granular control over the options shown.

### Possible Drawbacks

None

### Verification Process

Go to the Language Processing settings and ensure there are still post type options shown and Media isn't shown anymore.

### Checklist:

### Applicable Issues

#242

### DIFF
--- a/includes/Classifai/Helpers.php
+++ b/includes/Classifai/Helpers.php
@@ -172,6 +172,34 @@ function get_watson_password() {
 }
 
 /**
+ * Get post types we want to show in the language processing settings
+ *
+ * @since 1.6.0
+ *
+ * @return array
+ */
+function get_post_types_for_language_settings() {
+	$post_types = get_post_types( [ 'public' => true ], 'objects' );
+
+	// Remove the attachment post type
+	unset( $post_types['attachment'] );
+
+	/**
+	 * Filter post types shown in language processing settings.
+	 *
+	 * @since 1.6.0
+	 * @hook classifai_language_settings_post_types
+	 *
+	 * @param {array} $post_types Array of post types to show in language processing settings.
+	 *
+	 * @return {array} Array of post types.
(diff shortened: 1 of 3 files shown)


ShahAaron:
@dkotter this needs looks good, it just needs the test removed from NaturalLanguageProcessing.php as media ( attachment ) has been removed as an option.

dkotter:
> @dkotter this needs looks good, it just needs the test removed from NaturalLanguageProcessing.php as media ( attachment ) has been removed as an option.

Thanks for catching that! This has been fixed

dinhtungdu (approved):

PR #256, opened 2020-10-28
Make sure (re)scan buttons show up immediately after uploading an image

dkotter:
### Description of the Change

Right now, if you upload an image directly to the media library and then click on that image to open the modal view, the scan/rescan buttons won't show up. You have to refresh the page before you see those.

The issue here is we have a check for the current screen and only show those buttons if the screen object doesn't exist. Not sure the historical context here but this does force a page refresh to see those settings, which is not ideal.

In testing, I don't see any issues with removing that check and it does fix the problem.

### Alternate Designs

None

### Benefits

Scan buttons show up immediately, without requiring a page refresh

### Possible Drawbacks

Whatever the original reason there was for adding this check (for instance, to solve some sort of error), that could happen again, now that the check is gone. I've not run into any issues testing this though.

### Verification Process

In the media library grid view, upload a new image and then click on that image to open the modal view. Ensure the scan buttons are showing up.

### Checklist:

### DIFF
--- a/includes/Classifai/Services/ImageProcessing.php
+++ b/includes/Classifai/Services/ImageProcessing.php
@@ -48,31 +48,30 @@ class ImageProcessing extends Service {
 	 * @param \WP_post $post        Post object for the attachment being viewed.
 	 */
 	public function add_rescan_button_to_media_modal( $form_fields, $post ) {
-		$screen   = get_current_screen();
-		$settings = get_option( 'classifai_computer_vision' );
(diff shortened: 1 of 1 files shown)


helen (approved):

jeffpaul:
@ryanwelcher if you manage to have time to review the historical changes we're removing it would be helpful to understand if there was a reasoning for them that we should still account for or if we're 👍🏼 to blaze a trail forward

PR #263, opened 2020-12-22
Fix WP 5.6 array_intersect_key error

phpbits:
This PR fixes the error below when using WordPress 5.6. This errors are only visible when on Gutenberg editor.

Screen Shot 2020-12-23 at 4 12 47 AM


dkotter:
@phpbits Code looks good but I think this will cause some other errors we will need to solve. Currently, we take that route argument and use that later on, to determine which type of processing we are doing. That check expects a string but we are now going to be passing an array, which I think will cause things to break.

See here for one place this is done: https://github.com/10up/classifai/blob/develop/includes/Classifai/Providers/Azure/ComputerVision.php#L928

jeffpaul:
This same error was reported today via email:
> I tried running ClassifAI 1.6.0 again today to try and see what I missed,
but unfortunately am getting this error when loading /wp-admin/post-new.php:

```Fatal error: Uncaught TypeError: array_intersect_key(): Argument #1 ($array) must be of type array, string given in wp-includes/rest-api/class-wp-rest-server.php:1402 Stack trace: #0 wp-includes/rest-api/class-wp-rest-server.php(1402): array_intersect_key('alt-tags', Array) #1 wp-includes/rest-api/class-wp-rest-server.php(1324): WP_REST_Server->get_data_for_route('/classifai/v1/a...', Array, 'view') #2 wp-includes/rest-api/class-wp-rest-server.php(1243): WP_REST_Server->get_data_for_routes(Array, 'view') #3 wp-includes/rest-api/class-wp-rest-server.php(1160): WP_REST_Server->get_index(Object(WP_REST_Request)) #4 wp-includes/rest-api/class-wp-rest-server.php(1007): WP_REST_Server->respond_to_request(Object(WP_REST_Request), '/', Array, NULL) #5 wp-includes/rest-api.php(479): WP_REST_Server->dispatch(Object(WP_REST_Request)) #6 wp-includes/rest-api.php(2520): rest_do_request(Object(WP_REST_Request)) #7 [internal function]: rest_preload_api_request(Array, '/') #8 wp-admin/edit-form-blocks.php(82): array_reduce(Array, 'rest_preload_ap...', Array) #9 wp-admin/post-new.php(72): require('/Users/p/Sites/...') #10 {main} thrown in wp-includes/rest-api/class-wp-rest-server.php on line 1402```

@dkotter @dinhtungdu if @phpbits approach is not the optimal path, then is there another approach you'd recommend to help resolve this error?

dkotter:
> if @phpbits approach is not the optimal path, then is there another approach you'd recommend to help resolve this error?

I think the fix we have here is correct but there's more needed to ensure this doesn't break other things within Classifai. Anywhere in the plugin we use that `route` argument, needs to be updated to account for this change from a string to an array.

jeffpaul:
@phpbits are you able to make those additional updates as @dkotter noted above?

jeffpaul:
After chatting with @phpbits I'm going to tag in @dinhtungdu to help work on the additional updates from @dkotter.

dinhtungdu:
@dkotter I updated the PR to address the data type change.

dkotter (approved):

PR #264, opened 2021-01-24
Addressing TODO from https://github.com/10up/classifai/issues/260 ComputerVision::reset_settings()

### LINKED ISSUE #260: Do the TODOs
Current count of seven different [TODOs in the code](https://github.com/10up/classifai/search?q=TODO&type=code).  Let's look to work on resolving and DOing those.

- [x] `Classifai/Providers/Azure/ComputerVision.php`: Implement reset_settings() method.
- [x] `Classifai/Providers/AWS/Comprehend.php`: Implement reset_settings() method.
- [ ] `Classifai/Providers/AWS/Comprehend.php`: Implement can_register() method.
- [x] `Classifai/Providers/AWS/Comprehend.php`: Implement sanitize_settings() method.
- [ ] `Classifai/Command/ClassifaiCommand.php`: gc
- [x] `Classifai/Providers/Provider.php`: Implement sanitize_settings() method.
- [x] `Classifai/Providers/Watson/NLU.php`: Implement can_register() method.
- [x] `Classifai/Providers/Watson/NLU.php`: Is this function, `check_license_key`, supposed to be here?
- [ ] `Classifai/Services/ServicesManager.php`: remove this requirement (see: `can_register`).

**Designs**
Probably none, but won't know until we dig into the TODOs.

Don't do the TODOs.  In which case, are they TODONTs?

**Additional context**
n/a

ActuallyConnor:
### Description of the Change

Addressing TODO from https://github.com/10up/classifai/issues/260

```
public function reset_settings() {
    // TODO: Implement reset_settings() method.
}
```

### Alternate Designs

n/a

### Benefits

Addressing issue related to resolving TODOs in the repository. Presumably this function will allow for a quick reset of the settings back to the default so that the user doesn't have to go through resetting settings manually.

### Possible Drawbacks

none identified

### Verification Process

I looked the serialized array data in the database for what is there when the plugin is activated upon a fresh install and modelled the reset_settings() method off of that data and the includes/Classifai/Providers/Watson/NLU.php reset_settings() method.

Not sure if there is more that you are looking for or if you would also like a button on this page that triggers an event to reset the settings. If there is something like that you are looking for please let me know :)

### Checklist:

### Applicable Issues

https://github.com/10up/classifai/issues/260

### Changelog Entry

Addressed TODO to implement ComputerVision::reset_settings() method

### DIFF
--- a/includes/Classifai/Providers/Azure/ComputerVision.php
+++ b/includes/Classifai/Providers/Azure/ComputerVision.php
@@ -49,10 +49,10 @@ class ComputerVision extends Provider {
 			'valid' => false,
 			'url' => '',
 			'api_key' => '',
-			'enable_image_captions' => '1',
-			'enable_image_tagging' => '1',
-			'enable_smart_cropping' => 'no',
-			'enable_ocr' => 'no',
+			'enable_image_captions' => true,
+			'enable_image_tagging' => true,
+			'enable_smart_cropping' => false,
+			'enable_ocr' => false,
 			'caption_threshold' => 75,
 			'tag_threshold' => 70,
 			'image_tag_taxonomy' => 'classifai-image-tags',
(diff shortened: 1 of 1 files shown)


dinhtungdu (commented):
The PR looks good. But I think we can go one step further. The settings are in two places which is a code smell to me. What do you think about creating a function that returns the default settings, then using that function in your function and `setup_fields_sections`?

ActuallyConnor:
Hey @dinhtungdu I created a get_default_settings() method that returns the default settings, then used that function in reset_settings() method and setup_fields_sections() method as you requested. Let me know what you think 😄

includes/Classifai/Providers/Azure/ComputerVision.php
```diff
@@ -603,16 +623,18 @@ protected function generate_image_tags( $tags, $attachment_id ) {
 	 */
 	public function setup_fields_sections() {
 		add_settings_section( $this->get_option_name(), $this->provider_service_name, '', $this->get_option_name() );
+		$settings = $this->get_default_settings();
```
dinhtungdu: We can be more explicit here, `$default_settings` sounds better to me.

includes/Classifai/Providers/Azure/ComputerVision.php
```diff
@@ -723,13 +700,57 @@ public function setup_fields_sections() {
...
+			esc_html__( 'Tag Taxonomy', 'classifai' ),
+			[ $this, 'render_select' ],
+			$this->get_option_name(),
+			$this->get_option_name(),
+			[
+				'label_for'     => 'image_tag_taxonomy',
+				'options'       => $options,
+				'default_value' => $settings['image_tag_taxonomy'],
+			]
+		);
```
dinhtungdu: Could you explain why we should move this code here?

dinhtungdu (commented):
Hi @ActuallyConnor, thanks for the update. I still have some questions, can you please take another look?

includes/Classifai/Providers/Azure/ComputerVision.php
ActuallyConnor: I decided to reorder the settings fields in the function to match the order that appears in the default settings array. I ordered the default settings array based on the order that was returned in the database when I created a fresh install of the plugin.

I think it's best if the settings field order matches the order of the default settings field array, but I am not married to the current order of things (as I said, I just copied the order based on the serialized array that was returned in the database).

In my opinion we have two routes to choose from:

1. Keep it as is since the settings fields order matches the default settings array order which matches the order of the serialized array.
2. Revert the settings fields order back to the way it was, then match the default settings array order to the original settings fields order (this would just mean that we would ignore the database serialized array settings order which is fine as well since no one will really be looking at the database serialized array settings order).

Let me know what you would prefer, I am fine with going either route.

includes/Classifai/Providers/Azure/ComputerVision.php
ActuallyConnor: Sounds good, I can implement this request.

includes/Classifai/Providers/Azure/ComputerVision.php
dinhtungdu: Hi @ActuallyConnor, I'm sorry for this very late reply. I think we should keep the order of fields as they were before. The order of fields in an associated array doesn't affect the way we're using it. And by keeping the order of settings fields, our users won't be surprised after updating the plugin. Do you want to continue working on this one (and updating the default setting variable name)? If not, please let me know, I can handle the rest.

dinhtungdu (approved):

PR #250, opened 2020-10-23
Fixes azure storage compatibility issue with classifai.

ShahAaron:
This PR resolves the compatibility issue between ClassifAI and Microsoft Azure Storage.

It would resolve #201 .

### Description of the Change

When images are added through the media gallery and Azure Blob Storage is used the images are not correctly smart cropped and the alt tags are not generated. This is due to the smart cropping and tagging happening after the image has been uploaded to Azure. By adjusting the priority of the filters to make sure smart cropping happens before the alt tags are generated it means that the image is smart cropped correctly with alt tags generated before sending the image to Azure.
### Benefits

New images are now added and smart cropped correctly before sending to Azure Blob Storage via the plugin https://github.com/10up/windows-azure-storage. All tags were generated correctly and the correct url generated in the media gallery.

### Possible Drawbacks

The smart cropping will not work locally as it needs to be from a publicly accessible url. The image will be correctly tagged and upload to Azure but not with the smart cropped image.

### Verification Process

**This only works on a publicly accessible site.**

1. Add [Azure Storage plugin](https://github.com/10up/windows-azure-storage). Activate it and go to settings->Microsoft Azure and varify the correct credentials have been added to link to the Azure Blob Container.
2. On the "Image Processing" Classifai settings page, with credentials in place, check the "Enable smart cropping" checkbox and the "Automatically Tag Images" checkbox.
3. Upload an image to the media gallery. 
4. Check the image is in Azure Storage
5. Check the image has the correct Azure url in the media gallery
6. Check Image tags have been generated correctly
7. Add the image to a test page and insure all alt tags have been automatically generated correctly for the image

### Checklist:

- [X] I have read the [**CONTRIBUTING**](/CONTRIBUTING.md) document.
- [X] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my change.
- [X] All new and existing tests passed.

PR #271, opened 2021-03-12
Update hooks priority for `wp_generate_attachment_metadata` 

thrijith:
- Update hooks priority for `wp_generate_attachment_metadata` to work with azure storage
- Move `classifai_generate_image_alt_tags_source_url` filter to helper function
- Use `get_modified_image_source_url` where rescanning is done local file

### Description of the Change

- ClassifAI is generating tags on `10` [priority](https://github.com/10up/classifai/blob/develop/includes/Classifai/Providers/Azure/ComputerVision.php#L68), while the storage plugin uploads image to storage blob and deletes locally on [priority](https://github.com/10up/windows-azure-storage/blob/develop/windows-azure-storage.php#L125-L127) `9`, by the time it is getting there to generate tags, the local image is gone and alt text is not generated, reducing priority of the hook will allow it to generate the text and then upload to storage.

- I found a related PR https://github.com/10up/classifai/pull/250 where the hook priority was increased due to similar issues, setting hook priority to be in following order `smart_crop` at `7` and `alt_text` at `8` should hopefully fix issues both issues.

- The above changes only fix issue when the image is being uploaded, it doesn't work when we try to generate/re-generate/rescan image tags from media modal for an existing image, adding the filter introduced in https://github.com/10up/classifai/pull/217 it other places will help with the same.

### Verification Process

- Tested on project that was facing this issue, applying these changes fixed the alt generation for images and image was successfully uploaded to Azure storage as well.

### Checklist:

### Applicable Issues

- Doesn't have one right now, happy to open one if needed.

### DIFF
--- a/includes/Classifai/Helpers.php
+++ b/includes/Classifai/Helpers.php
@@ -483,3 +483,29 @@ function get_largest_size_and_dimensions_image_url( $full_image, $full_url, $met
 
 	return null;
 }
+
+/**
+ * Allows returning modified image URL for a given attachment.
+ *
+ * @param int $post_id Post ID.
+ *
+ * @return mixed
+ */
+function get_modified_image_source_url( $post_id ) {
+	/**
+	 * Filter to modify image source URL in order to allow scanning images,
+	 * stored on third party storages that cannot be used by
+	 * helper function `get_largest_acceptable_image_url()` to determine `filesize()` locally.
+	 *
+	 * Default is null, return filtered string to allow classifying image on external source.
+	 *
+	 * @since 1.6.0
+	 * @hook classifai_generate_image_alt_tags_source_url
+	 *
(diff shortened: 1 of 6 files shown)


dkotter (dismissed):

dkotter:
@thrijith Thanks for this PR. Code looks good and worked fine in my testing. There is a failed test though, if you can look at that and get it fixed up?

thrijith:
> @thrijith Thanks for this PR. Code looks good and worked fine in my testing. There is a failed test though, if you can look at that and get it fixed up?

Thanks for the review @dkotter, the test failure seems unrelated to the change, I've updated the failing case it seems to be passing now, also added a change to fix some other alerts for risky tests.

dkotter (approved):

PR #211, opened 2020-06-04
Ensuring Classifiai passes VIP code scans

ryanwelcher:
- [x] I have read the [**CONTRIBUTING**](/CONTRIBUTING.md) document.
- [x] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my change.
- [ ] All new and existing tests passed.

### Applicable Issues

#182

PR #287, opened 2021-06-04
Passing VIPCS check

### LINKED ISSUE #182: Submit 1.5.0 to VIP for review
We'll want to ensure that any one on WordPress VIP are able to use ClassifAI, but we're not currently certain that's the case.

Submit ClassifAI for VIP review and work to resolve any issues identified within a quickly following release.

**Designs**

n/a

n/a

**Additional context**

n/a

dinhtungdu:
### Description of the Change
This PR is a replacement for #211 which ensures ClassifAI passes the phpcs check using WordPress-VIP-Go standard.

Note that I excluded `testst` folder in this check.

### Verification Process
Install VIPCS and run this command:
```
phpcs --standard=WordPress-VIP-Go -s --extensions=php --ignore=node_modules,vendor,tests .
```

### Checklist:

- [ ] I have read the [**CONTRIBUTING**](/CONTRIBUTING.md) document.

### Applicable Issues
#182

### DIFF
--- a/autoload.php
+++ b/autoload.php
@@ -68,8 +68,7 @@ class Psr4AutoloaderClass {
 
 		// work backwards through the namespace names of the fully-qualified
 		// class name to find a mapped file name
-		$pos = strrpos( $prefix, '\\' );
-		while ( false !== $pos ) {
+		while ( false !== $pos = strrpos( $prefix, '\\' ) ) { // phpcs:ignore WordPress.CodeAnalysis.AssignmentInCondition.FoundInWhileCondition
 
 			// retain the trailing namespace separator in the prefix
 			$prefix = substr( $class, 0, $pos + 1 );
--- a/autoload.php
+++ b/autoload.php
@@ -68,7 +68,8 @@ class Psr4AutoloaderClass {
 
 		// work backwards through the namespace names of the fully-qualified
 		// class name to find a mapped file name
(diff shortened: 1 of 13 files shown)


autoload.php
```diff
@@ -68,7 +68,8 @@ public function load_class( $class ) {
 
 		// work backwards through the namespace names of the fully-qualified
 		// class name to find a mapped file name
-		while ( false !== $pos = strrpos( $prefix, '\\' ) ) {
+		$pos = strrpos( $prefix, '\\' );
+		while ( false !== $pos ) {
```
dkotter: So I didn't test this but it seems like this wouldn't work the way we want anymore, by removing the `strrpos` out of the `while` statement. If I'm following this correctly, the way this is set up now, the `while` statement would either only execute once or would get stuck in an infinite loop, since the value of `$pos` never changes after it's first set.

dkotter (changes requested):

autoload.php
dinhtungdu: Thanks for the head-up, you're right, `$pos` doesn't get updated because `strrpos` will keep using the value of `$prefix` outside the loop. It's my mistake that I didn't test the `autoload.php`, I was using composer autoload.

dinhtungdu: @dkotter I reverted the change of this line. Tested and it works with `autoload.php`.

dkotter (approved):

PR #295, opened 2021-07-21
Fix: Generate/regenerate/rescan buttons do not work in media modal when editing a post

### LINKED ISSUE #292: [BLOCKER] Generate/regenerate/rescan buttons do not work in media modal when editing a post
**Describe the bug**
Screen Shot 2021-07-13 at 9 02 16 AM

These buttons do not do anything when clicked while editing a post, as in they do not fire off a network request or show a spinner or anything like that, you can click/activate them but it just no-ops. They do work fine in the modal when on the media library screen. I think I have noticed this before intermittently and chalked it up to caching or build oddities but it seems very consistent now.

**Steps to Reproduce**

1. Edit a post
2. Bring up the media modal by inserting an image block and selecting media library
3. Choose an image
4. Try clicking the buttons in the side panel, observing network requests in the inspector

**Expected behavior**
I would expect the buttons to work the same way they do when on the media library screen.

**Additional context**
Side note: why do we have inconsistent labels?
Screen Shot 2021-07-13 at 9 09 04 AM

dinhtungdu:
### Description of the Change

The buttons in the media only work under the Media library page. This happens because we only listen to `edit:attachment` event triggered when you open the media modal of an image on the Media library page. This PR adds a listener to another event to make the buttons work when editing a post.

### Verification Process

1. Build new assets.
2. Edit a post using the block editor, see buttons work.
3. Edit a post using the classic editor, see buttons work.
4. Edit an attachment, see buttons work.

### Checklist:

### Applicable Issues
Close #292.

### Changelog Entry

Fixed: Ensure Image Processing buttons in the media modal work when editing posts.


helen:
This does work now for existing items; however, it doesn't seem to work for new uploads within that modal "session" - they'll work if you close the modal and reopen it, though. Maybe it has to be bound a different way for that? Perhaps not a blocker since there's a workaround but wanted to note it in case we want to log it as as separate issue and release with it this way, which is already better.

dinhtungdu:
> they'll work if you close the modal and reopen it

@helen they also work if you unselect then select again. I tried to fix this issue as well but it's not simple as it seems. I will look at it again next week.

helen:
Just noting in writing that while less than great the fix is still better than before so we can figure out the final piece (not having to click away and then back to have the buttons work) later on.

helen (approved):

PR #311, opened 2021-10-04
Fix/classify post action

mustafauysal:
### Description of the Change

These changes address 2 issues with CPT.

1) Don't show the "Classify" action when the `post_type` is not supported. (`post_row_actions`, `page_row_actions` could allow to shown classify action on unregistered post type)
2) Pass `post_type` for individual classify action.  (eg: when you enable classify for the "page" and click on the row item, it doesn't work due to missing `post_type` on request.

### Benefits

Making ClassifAI work better in admin UI.

### Possible Drawbacks

Haven't seen any drawback, changes are relatively small.

### Verification Process

1) Register CPT: 
```
function setup_test_post_type() {
	$args = array(
		'public' => true,
		'label'  => esc_html__( 'FooBar', 'foobar' ),
	);
	register_post_type( 'foobar', $args );
}

add_action( 'init', 'setup_test_post_type' );
```
2) Enable ClassifAI for `foobar` only
3) Click to "Classify" action on `/wp-admin/edit.php?post_type=foobar` for a post
4) See "Classified 1 post." message.

### Checklist:

I couldn't add tests at this time, but it might be worth adding some testing 🤷🏻‍♂️

### Changelog Entry

* Fix individual classify action.

### DIFF
--- a/includes/Classifai/Admin/BulkActions.php
+++ b/includes/Classifai/Admin/BulkActions.php
@@ -122,6 +122,12 @@ class BulkActions {
 	 * @return array
 	 */
 	public function register_row_action( $actions, $post ) {
+		$post_types = get_supported_post_types();
+
+		if ( ! in_array( $post->post_type, $post_types, true ) ) {
+			return $actions;
+		}
+
 		$actions['classify'] = sprintf(
 			'<a href="%s">%s</a>',
 			esc_url( wp_nonce_url( admin_url( sprintf( 'edit.php?action=classify&ids=%d&post_type=%s', $post->ID, $post->post_type ) ), 'bulk-posts' ) ),
--- a/includes/Classifai/Admin/BulkActions.php
+++ b/includes/Classifai/Admin/BulkActions.php
@@ -72,6 +72,7 @@ class BulkActions {
 		if ( 'classify' !== $doaction ) {
 			return $redirect_to;
 		}
+
 		foreach ( $post_ids as $post_id ) {
 			$this->save_post_handler->classify( $post_id );
 		}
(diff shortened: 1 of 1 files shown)


includes/Classifai/Admin/BulkActions.php
```diff
@@ -121,9 +122,15 @@ public function bulk_action_admin_notice() {
...
+		$post_types = get_supported_post_types();
+
+		if ( ! in_array( $post->post_type, $post_types, true ) ) {
+			return $actions;
+		}
+
 		$actions['classify'] = sprintf(
 			'<a href="%s">%s</a>',
-			esc_url( wp_nonce_url( admin_url( 'edit.php?action=classify&ids=' . $post->ID ), 'bulk-posts' ) ),
+			esc_url( wp_nonce_url( admin_url( sprintf( 'edit.php?action=classify&ids=%d&post_type=%s', $post->ID, $post->post_type ) ), 'bulk-posts' ) ),
```
cadic, line 133: Looks redundant, the hook `handle_bulk_actions-edit-$post_type` works regardless of `post_type` query argument presence.

cadic (changes requested):
@mustafauysal thank you for this update. Please fix the code regarding the `post_type` argument or explain the necessity of it.

includes/Classifai/Admin/BulkActions.php
mustafauysal: @cadic It's not redundant but it can be a bit tricky to realize how it hooked on a single classification request at first glance. 

In order to see reproduce the issue, please enable classify for "page" post type only, and click on the "Classify" row action for any given page.  You will see nothing happens without passing the correct `post_type`.

`wp-admin/edit.php` needs to know `$post_type` to execute the dynamic hook (`"handle_bulk_actions-edit-$post_type"`) accordingly.

I hope this makes sense.

cadic: @mustafauysal thank you, I can confirm the action link doesn't work without `post_type`

cadic (approved):
@mustafauysal thank you, looks good!

PR #329, opened 2022-02-21
Added toggle to enable/disable language processing on update post.

### LINKED ISSUE #185: Ability to remove tags without ClassifAI immediately re-adding them back
There currently is no way within ClassifAI to remove tags on posts/pages.  If you try to do so and then save that post/page, then ClassifAI will simply re-send that post/page off and likely re-apply the removed tag via IBM Watson.  The only "option" would be to deactivate ClassifAI, make those changes, and then re-activate ClassifAI; but that's clearly sub-optimal.

We could add an on/off toggle in a metabox in the Classic Editor and potentially within a Gutenberg slotfill that would allow someone to save/publish/update without sending the content to Watson.  This seems like the simplest approach, but likely warrants some UX/design to ensure we're handling this as gracefully as possible.

**Designs**

TBD, but needed.

None, keeping as-is is not ideal.

**Additional context**

None.

iamdharmesh:
### Description of the Change

As described on #185, currently there is no way within ClassifAI to remove tags on posts/pages, ClassifAI is re-apply the removed tag via IBM Watson on save post/page. To fix this issue, this PR adds a toggle to enable/disable language processing on save the post. So, if users want to remove tags or want to add custom tags they can do after disabling the language processing on the save post.

Gutenberg | Classic Editor |
--- | --- |
image|image|

_Note: This toggle only disable language processing on save post, Classify using bulk actions or CLI will work as it is,_

Closes #185 

### Alternate Designs
Mentioned here https://github.com/10up/classifai/issues/185#issuecomment-784611986

### Possible Drawbacks
not known any

### Verification Process
Disable language processing on save the post and verify that removed tags are not getting added back on save post/page afterward.

### Checklist:

### Credits

Props @iamdharmesh @jeffpaul


includes/Classifai/Admin/SavePostHandler.php
```diff
@@ -89,6 +89,11 @@ public function did_save_post( $post_id ) {
 
 		// Process posts in allowed post statuses, supported items and only if features are enabled
 		if ( in_array( $post_status, $post_statuses, true ) && in_array( $post_type, $supported, true ) && \Classifai\language_processing_features_enabled() ) {
+			// Check if process content on save is disabled for specific post.
```
dkotter:
```suggestion
			// Check if processing content on save is disabled.
```

src/js/gutenberg-plugin.js
```diff
@@ -0,0 +1,89 @@
...
+		size={ 20 }
+	/>
+);
+
+/**
+ * ClassifAIToggle Component.
+ *
+ */
+const ClassifAIToggle = () => {
+	// Use the datastore to retrieve the all the meta for this post.
```
```suggestion
	// Use the datastore to retrieve all the meta for this post.
```

includes/Classifai/Providers/Watson/NLU.php
```diff
@@ -172,13 +179,37 @@ public function get_settings( $index = false ) {
...
 			CLASSIFAI_PLUGIN_VERSION,
 			true
 		);
+
+		if ( empty( $post ) ) {
+			return;
+		}
+
+		wp_enqueue_script(
+			'classifai-gutenberg-plugin', // Handle.
```
dkotter:
```suggestion
			'classifai-gutenberg-plugin',
```

includes/Classifai/Providers/Watson/NLU.php
```diff
@@ -736,4 +767,103 @@ private function get_formatted_latest_response() {
...
+				null,
+				'side',
+				'low',
+				array( '__back_compat_meta_box' => true )
+			);
+		}
+	}
+
+	/**
+	 * Render Metabox content.
```
dkotter:
```suggestion
	 * Render metabox content.
```

dkotter (dismissed):

iamdharmesh:
@dkotter @jeffpaul E2E Tests are failing for WP minimum env (#5.2) here. It's failing because support for adding a panel (slotfill) in the editor sidebar is added in WP 5.3. So, it breaks the gutenberg. is it fine to proceed here with updating the minimum WP version to 5.3 in plugin OR do we have to add an additional condition in code to add slotfill only for WP >= **5.3**? (Same question for AutoShare for Twitter as well, for PR I raised today)

Language processing Tests -- Can create post and taxonomy terms get created by ClassifAI (with 75 threshold) (failed)


Thanks

dkotter:
> is it fine to proceed here with updating the minimum WP version to 5.3 in plugin

I think this is fine, I'm all for bumping our minimums more regularly and 5.3 has been out since November 2019, so I don't see any issues with that (could maybe even bump that higher)

dkotter:
I guess one more thing here. I see our E2E tests are using WP 5.2 as our minimum but in other places we say our minimum is 5.0. We should ensure we update our minimums in all locations so they match

jeffpaul:
I'm fine bumping WP minimum to 5.3 (but also probably as far as 5.8 perhaps) so @iamdharmesh pick what's best/easiest here and bump in all places as Darin noted above.  When making that minimum bump you'll want to check for any conditionals/checks we may for older versions that could be removed.  We'll also want to add a note in the changelogs stating that the WP minimum was bumped from/to so that it's explicitly called out for folks.

dkotter:
Yeah, just finished reviewing #343 and it relies on features from WP 5.6, so seems we should at least bump to that

iamdharmesh:
Thanks for the feedback @dkotter @jeffpaul.

I have bumped the WP minimum to 5.6 and added a changelog in the PR description.

@jeffpaul `phpunit` check is removed from workflow but it's still showing Required and Expected. can you please help to remove it from the required? 

Thanks.

dkotter (approved):

PR #310, opened 2021-09-08
Added function to retrieve filterable post statuses for classification.

jamesmorrison:
### Description of the Change

Adds a filter to allow post statuses other than `publish` to be allowed to classify content.

### Benefits

Users will be able to save content in other post statuses - e.g. `draft` and this content will be classified.

### Possible Drawbacks

If `publish` is removed from the array, content will not be classified on publication.

### Verification Process

I ran the following code in a MU plugin to modify the post statuses:

```
add_action(
	'after_setup_theme',
	function() {
		add_filter(
			'classifai_post_statuses_for_post_type_or_id',
			function( $post_statuses, $post_type, $post_id ) {
				return array(
					'publish',
					'draft',
					'future',
				);
			},
			10,
			3
		);
	}
);
```

The resulting data upon saving a draft post was:

```
{"post_statuses":["publish","draft","future"],"post_status":"draft","post_type":"post","post_id":2285}
```

### Checklist:

- [x] I have read the [**CONTRIBUTING**](/CONTRIBUTING.md) document.
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [x] I have added tests to cover my change.
- [ ] All new and existing tests passed.

### Applicable Issues

https://github.com/10up/classifai/issues/309

### Changelog Entry

Added: `classifai_post_statuses` filter; allows post statuses for content classification to be changed as required (but would apply to all post types).

Added: `classifai_post_statuses_for_post_type_or_id` filter; allows post statuses for content classification to be changed as required based on post type / post ID.

PR #330, opened 2022-03-04
Enhancement/266 A filter to manipulate the post statuses when classifying with Watson

### LINKED ISSUE #266: Add filter for processing on alternate post status

We're using Watson language processing as an element in classifying whether or not a document is sufficiently topical to be published. For this to be effective, we want to run processing on posts prior to publishing them. 

In https://github.com/10up/classifai/blob/51584310501073b7fb008139d2922e75b2fda50c/includes/Classifai/Command/ClassifaiCommand.php add a filter on post statuses to allow classification to run in contexts other than published posts. 

**Designs**

Alternate: Operate Watson on a private post type and switch post types if classification meets established criteria. Would prefer to us post status, however. 

**Additional context**

faisal-alvi:
### Description of the Change

This PR introduces a filter to manipulate the post statuses of posts need to be classified with Watson.

Closes #266.

### Checklist:

- [ ] I have read the [**CONTRIBUTING**](/CONTRIBUTING.md) document.

### Changelog Entry

> Added - A filter to manipulate post statuses to classify with Watson

### Credits

Props @

### DIFF
--- a/includes/Classifai/Command/ClassifaiCommand.php
+++ b/includes/Classifai/Command/ClassifaiCommand.php
@@ -437,9 +437,22 @@ class ClassifaiCommand extends \WP_CLI_Command {
 	 * @return array
 	 */
 	private function get_posts_to_classify( $opts = [] ) {
+		/**
+		 * Filter post statuses when classifying with Watson.
+		 *
+		 * @since x.x.x
+		 * @hook classifai_post_statuses_for_watson
+		 *
+		 * @param array $post_statuses Array of post statuses to be classified with language processing.
+		 * @param array $opts Options from WP CLI.
+		 *
+		 * @return array Array of post statuses.
+		 */
+		$post_statuses = apply_filters( 'classifai_post_statuses_for_watson', array( 'publish' ), $opts );
+
 		$query_params = [
 			'post_type'      => ! empty( $opts['post_type'] ) ? $opts['post_type'] : 'any',
-			'post_status'    => 'publish',
+			'post_status'    => $post_statuses,
 			'fields'         => 'ids',
 			'posts_per_page' => -1,
 		];


dkotter:
@faisal-alvi Code looks fine here but I guess I'm a little confused on the actual request. The original issue mentions they want to classify things prior to publish to be help determine what content should be published.

I think this is a fine request but the file they pointed to (and where we made the change) is in our WP-CLI command. This code will not fire as part of the normal publication process in WordPress, it will have to be manually run. If we do want the change here, I'd suggest instead of a filter we just add `post_status` as a supported argument to our WP-CLI command (like we already support with `post_type`).

But probably worth clarifying with the original reporter on if this is what they really want (changes to the WP-CLI command). Seems to me we'd want to support other post statuses other than `publish` when items are actually being saved in the admin, not through manually triggering a WP-CLI command. But I may not fully understand their current workflow, which I think would be helpful to understand.

jeffpaul:
@joedolson thoughts on the above?

joedolson:
Thanks, @jeffpaul - My use case does need usage within the admin. The use case is that items are submitted into WP posts via REST API in unpublished statuses, run through ClassifAI language processing, then reviewed by moderators to determine whether they meet expectations for publication. The ClassifAI processing is part of the processing for identifying the editor responsible for review, so it needs to happen before publishing.

jeffpaul:
@joedolson would the filter for `post_status` as Darin noted work for you or would an alternate approach be a more ideal implementation?

joedolson:
As long as the filter doesn't depend on WP CLI, it should work. It looks like the current code at https://github.com/10up/classifai/blob/97541047ac827e4ac14308b69c2890c116b06348/includes/Classifai/Admin/SavePostHandler.php#L78 would meet my needs. 

That filter comment is wrong, though; that filter is new, not since 1.0.0.

dkotter:
@joedolson @jeffpaul That filter was added in #310, which has not been released yet (we will need to update the `@since` comments once released). But I do think that solves the reported issue here and we can probably close this PR out

PR #341, opened 2022-05-03
Add filter to modify the registration conditions for the SavePostHandler class.

### LINKED ISSUE #123: RSS Feed ingestion not triggering Watson NLU tagging
**Describe your question**
In chatting with @dingman, their site scenario is based more on content ingestion via RSS feeds using the [WordPress Automatic Plugin](https://codecanyon.net/item/wordpress-automatic-plugin/1904470) and a cron job.  This posting of content does not appear to trigger  sending the content to Watson for Language Processing as no tags appear on content regardless how low the thresholds are, but manually updating the content gets the tags to appear.  Potential issue here is that the WP Automatic plugin doesn't use the same hooks that ClassifAI does, so can we confirm to @dingman what hook(s) ClassifAI uses to compare with WP Automatic?

s3rgiosan:
### Description of the Change

This PR introduces a filter to modify the registration conditions for the `SavePostHandler` class.

Closes #123

### Checklist:

### Changelog Entry

> Added - `classifai_should_register_save_post_handler` filter; allows modifying the registration conditions for the `SavePostHandler` class (props [@s3rgiosan](https://github.com/s3rgiosan) via https://github.com/10up/classifai/pull/341).

### Credits

Props @s3rgiosan


includes/Classifai/Admin/SavePostHandler.php
```diff
@@ -21,26 +21,27 @@ public function register() {
...
-			return true;
-		} elseif ( defined( 'PHPUNIT_RUNNER' ) && PHPUNIT_RUNNER ) {
-			return false;
-		} elseif ( defined( 'WP_CLI' ) && WP_CLI ) {
-			return false;
-		} else {
-			return false;
+
+		$should_register = false;
+		if ( is_admin() || $this->is_rest_route() ) {
```
dkotter: I think we need to add more conditions here. Right now, this class will be registered any time we're in the admin, even if things haven't been configured yet. I think we need to keep the configuration checks (the `classifai_configured` and `classifai_watson_nlu` options)

dkotter (changes requested):

includes/Classifai/Admin/SavePostHandler.php
s3rgiosan: @dkotter It would help if we did something like this?

```
	/**
	 * Check to see if we can register this class.
	 */
	public function can_register() {

		$should_register = false;
		if ( $this->is_configured() && ( is_admin() || $this->is_rest_route() ) ) {
			$should_register = true;
		}

		/**
		 * Filter whether ClassifAI should register this class.
		 *
		 * @since 1.8.0
		 * @hook classifai_should_register_save_post_handler
		 *
		 * @param  {bool} $should_register Whether the class should be registered.
		 * @return {bool} Whether the class should be registered.
		 */
		$should_register = apply_filters( 'classifai_should_register_save_post_handler', $should_register );

		return $should_register;
	}

	/**
	 * Check if ClassifAI is properly configured.
	 *
	 * @return bool
	 */
	public function is_configured() {
		return get_option( 'classifai_configured' ) && ! empty( get_option( 'classifai_watson_nlu' )['credentials']['watson_url'] );
	}
```

dkotter: Yep, that is way more clean, I like it!

s3rgiosan: @dkotter I've pushed the requested changes, and updated the unit tests.

dkotter (approved):

PR #343, opened 2022-05-11
Added "Recommended Content" Block

### LINKED ISSUE #69: Integrate Azure Personalizer for personalized content
Increasing the relevance of recommended content 

Leverage [Project Custom Decision](https://labs.cognitive.microsoft.com/en-us/project-custom-decision) from Microsoft Cognitive Services Labs.  Project Custom Decision (formerly Custom Decision Service) is an experimental API offered as part of Microsoft Azure Cognitive Services.  It uses Reinforcement Learning to personalize content for each visitor, and experiments with new choices to see how different options perform and adjust to emerging trends.

**Designs**
n/a

I have yet to research alternatives in existence.

**Additional context**
I believe we've got some sample code that could help, I'll look to find that before we'd begin on this work.

iamdharmesh:
### Description of the Change
This PR Adds the Initial "Recommended Content" Block backed by the Azure Personalizer. Major changes of PR are listed below.
- Added `@wordpress/scripts`
- Format js and fixed reported eslint errors.
- Integration with Azure Personalizer API.
- Added "Recommended Content" Block.

**Note:** This is yet initial implementation based on personalizer documentation and we may still have to check on the below stuff to close #69
- Include User location data in `contextFeatures` for better personalization. 
- <del>Currently, block only suggest 1 recommended content as personalizer only return 1 rewarded action. We should check on the possibility/workaround to display multiple contents.</del> (We are showing multiple items now, as decided in internal slack discussion)
- Define suggested configuration to make the plugin work more effectively (config like `model update frequency`, `reward wait time`, etc...).
- Verify content personalization with huge content and different use cases

Some Useful links for Azure Personalizer:
- [Personalizer Documentation](https://docs.microsoft.com/en-us/azure/cognitive-services/personalizer/)
- [How Personalizer works](https://docs.microsoft.com/en-us/azure/cognitive-services/personalizer/how-personalizer-works)

##### Personalizer Settings:
image

##### Recommended Content Block:
image

Closes #69

### Alternate Designs
We may choose any other recommended content service, other than Azure Personalizer.

### Possible Drawbacks
Not identified any.

### Verification Process
1. Save Azure Personalizer settings. 
2. Verify it gets authenticated and saved. (wrong key we should show error on save)
3. Add "Recommended Content" Block
4. Verify Filters and various setting options work properly.
5. Verify Block content displays properly in front-end and send the reward to azure on click-on content.

### Checklist:

### Credits

Props @iamdharmesh @jeffpaul @Sidsector9


jeffpaul:
> Currently, block only suggest 1 recommended content as personalizer only return 1 rewarded action. We should check on the possibility/workaround to display multiple contents.

If we're unable to find a way to get multiple posts recommended back from Azure Personalizer, then we should consider a slight renaming from ~`Recommended Content` (and related copy) to something more akin to ~`Read This Next` (since it would only be 1 post recommended).

jeffpaul:
@dkotter pinging you here as reminder on code review / feedback on this PR / new functionality.

includes/Classifai/Blocks.php
```diff
@@ -0,0 +1,90 @@
...
+	};
+
+	add_action( 'enqueue_block_assets', $n( 'blocks_styles' ) );
+	add_filter( 'block_categories_all', $n( 'blocks_categories' ), 10, 2 );
+
+	register_blocks();
+}
+
+/**
+ * Registers blocks that are located within the includes/blocks directory
```
dkotter:
```suggestion
 * Registers blocks that are located within the includes/blocks directory.
```

includes/Classifai/Blocks.php
```diff
@@ -0,0 +1,90 @@
...
+ *
+ * @return void
+ */
+function register_blocks() {
+	// Require custom blocks
```
dkotter:
```suggestion
	// Require custom blocks.
```

includes/Classifai/Blocks.php
```diff
@@ -0,0 +1,90 @@
...
+	require_once CLASSIFAI_PLUGIN_DIR . '/includes/Classifai/Blocks/recommended-content-block/register.php';
+
+	// Call block Register function for each block.
```
dkotter:
```suggestion
	// Call register function for each block.
```

includes/Classifai/Blocks/recommended-content-block/edit.js
```diff
@@ -0,0 +1,149 @@
...
+		displayPostExcept,
+		addLinkToFeaturedImage,
+	} = attributes;
+
+	const blockProps = useBlockProps();
+	const { postTypesTaxonomiesMap, postTypesSelectOptions } = usePostTypes();
+	const onPostTypeChange = (newValue) => {
+		const updateQuery = { contentPostType: newValue };
+		// We need to dynamically update the `taxQuery` property,
+		// by removing any not supported taxonomy from the query.
```
dkotter:
```suggestion
		// by removing any not supported taxonomies from the query.
```

includes/Classifai/Blocks/recommended-content-block/frontend.js
```diff
@@ -0,0 +1,67 @@
+import './frontend.scss';
+
+/* global classifai_personalizer_params */
+jQuery(document).ready(function () {
```
dkotter: Do we need `jQuery` here or can this be replaced by vanilla JS? I'm assuming this only loads on the front-end, where jQuery isn't loaded by default (like it is in the admin). Seems the big thing we're using it for is the ajax request but since we're using `fetch` in another place, could we also utilize that there?

includes/Classifai/Blocks/recommended-content-block/index.js
```diff
@@ -0,0 +1,32 @@
...
+
+/* Uncomment for CSS overrides in the admin */
+// import './index.css';
+
+/**
+ * Register block
+ */
+registerBlockType(block.name, {
+	title: __('Recommended Content', 'classifai'),
+	description: __('Display Recommended content by Azure Personalizer', 'classifai'),
```
dkotter:
```suggestion
	description: __('Display Recommended content from Azure Personalizer', 'classifai'),
```

includes/Classifai/Blocks/recommended-content-block/markup.php
```diff
@@ -0,0 +1,20 @@
+<?php
+/**
+ * Recommended Content block markup
+ *
+ * @package Classifai\Blocks
+ *
+ * @var array    $attributes         Block attributes.
+ * @var string   $content            Block content.
+ * @var WP_Block $block              Block instance.
+ * @var array    $context            BLock context.
```
dkotter: Some extra spacing here it seems, plus a capital `L` that we can lowercase

package.json
```diff
@@ -37,23 +40,24 @@
...
   "devDependencies": {
     "@10up/cypress-wp-utils": "github:10up/cypress-wp-utils#build",
-    "@10up/eslint-config": "^1.0.9",
-    "@babel/core": "^7.5.5",
-    "@babel/preset-env": "^7.5.5",
-    "@babel/preset-react": "^7.0.0",
     "@wordpress/env": "^4.5.0",
-    "babel-eslint": "10.0.3",
-    "babel-loader": "^8.0.6",
+    "10up-toolkit": "^4.0.0",
```
dkotter: So this is probably fine though I know we've tried to stay away from 10up specific toolkits in our OSS plugins. May be worth using `wp-scripts` here instead

includes/Classifai/Providers/Azure/Personalizer.php
```diff
@@ -0,0 +1,608 @@
...
+	 * Can the functionality be initialized?
+	 *
+	 * @return bool
+	 */
+	public function can_register() {
+		$options = get_option( $this->get_option_name() );
+		if ( isset( $options['authenticated'] ) && false === $options['authenticated'] ) {
+			return false;
+		}
+		if ( empty( $options ) ) {
```
dkotter: I'd probably combine this `if` statement with the one above it, just to simplify things

includes/Classifai/Providers/Azure/Personalizer.php
```diff
@@ -0,0 +1,608 @@
...
+					);
+				}
+			}
+		}
+
+		return $new_settings;
+	}
+
+	/**
+	 * Get Recent posts based on given arguments
```
dkotter:
```suggestion
	 * Get Recent posts based on given arguments.
```

includes/Classifai/Providers/Azure/Personalizer.php
```diff
@@ -0,0 +1,608 @@
...
+				'label_for'     => 'api_key',
+				'input_type'    => 'password',
+				'default_value' => $default_settings['api_key'],
+				'description'   => __( 'Azure Cognitive Service Personalizer Key.', 'classifai' ),
+			]
+		);
+	}
+
+	/**
+	 * Sanitization
```
dkotter:
```suggestion
	 * Sanitize settings.
```

includes/Classifai/Providers/Azure/Personalizer.php
```diff
@@ -0,0 +1,608 @@
...
+	/**
+	 * Register the functionality.
+	 */
+	public function register() {
+		// Setup Blocks
+		Blocks\setup();
+	}
+
+	/**
+	 * Setup fields
```
dkotter:
```suggestion
	 * Setup fields.
```

includes/Classifai/Providers/Azure/Personalizer.php
```diff
@@ -0,0 +1,608 @@
...
+
+		if ( isset( $attributes['displayPostDate'] ) && $attributes['displayPostDate'] ) {
+			$markup .= sprintf(
+				'<time datetime="%1$s" class="wp-block-classifai-recommended-content__post-date">%2$s</time>',
+				esc_attr( get_the_date( 'c', $post ) ),
+				esc_html( get_the_date( '', $post ) )
+			);
+		}
+
+		if ( isset( $attributes['displayPostExcept'] ) && $attributes['displayPostExcept'] ) {
```
dkotter: Typo here, should be `displayPostExcerpt`

dkotter (changes requested):

dkotter:
I've left a few comments on the code but haven't actually tested this out yet.

But prior to that, I do think this warrants a conversation on what we actually want to achieve here. As noted, we can only return a single piece of content, so not sure a full-fledged Recommended Content block makes sense here. Jeff mentioned this could be used as a `Up Next` sort of feature, and that probably works better. The issue I'm seeing though is this feature is meant to be used on a small content base (maximum of 50 items). We are putting that limit in place at the code-level but I wonder how that impacts results on sites that have hundreds or thousands of posts.

An intriguing idea that was mentioned as a use-case in the documentation is `determine what product to suggest to shoppers`. I could see us building an integration with WooCommerce for a feature like that, which may be more useful than a `Read Next` section. Another use-case mentioned is `figure out the optimal position for an advertisement` which also could have wide usage.

So I don't want to de-rail things here and this may be useful enough to release and try out, but I wonder if there are better options for this Personalizer service, realizing we can only show a single item and we can only have a limited set of content (max 50). Also `Has a minimum of ~1k/day content-related events for Personalizer to be effective`, so will need to mention this will only work properly on sites that get decent traffic.

package.json
iamdharmesh: Sure, I was not aware of this. I followed this from https://github.com/10up/publisher-media-kit/, I will migrate this to `wp-scripts`.

includes/Classifai/Services/Personalizer.php
```diff
@@ -0,0 +1,134 @@
...
+		);
+	}
+
+	/**
+	 * Register the rest API endpoints
+	 */
+	public function init() {
+		parent::init();
+		add_action( 'rest_api_init', [ $this, 'register_endpoints' ] );
+		add_action( 'wp_ajax_render_recommended_content', [ $this, 'ajax_render_recommended_content' ] );
```
dkotter: We'll probably want to prefix these actions so we don't risk a conflict if some other plugin also uses the `render_recommended_content` action

includes/Classifai/Providers/Azure/Personalizer.php
```diff
@@ -0,0 +1,608 @@
...
+
+		if ( isset( $attributes['displayPostDate'] ) && $attributes['displayPostDate'] ) {
+			$class .= ' has-dates';
+		}
+
+		if ( isset( $attributes['displayAuthor'] ) && $attributes['displayAuthor'] ) {
+			$class .= ' has-author';
+		}
+
+		$wrapper_attributes = get_block_wrapper_attributes( array( 'class' => $class ) );
```
dkotter: A couple things here:

1. When testing things locally, I'm getting PHP warnings originating from here: `Trying to access array offset on value of type null in "wp-includes/class-wp-block-supports.php" lines 94 and 96`
2. This function was added in WP 5.6.0 which is above our minimum requirements. If we need to use this, we'll need to bump our minimums to 5.6

includes/Classifai/Blocks/recommended-content-block/register.php
```diff
@@ -0,0 +1,48 @@
...
+ * Render callback method for the block
+ *
+ * @param array  $attributes The blocks attributes
+ * @param string $content    Data returned from InnerBlocks.Content
+ * @param array  $block      Block information such as context.
+ *
+ * @return string The rendered block markup.
+ */
+function render_block_callback( $attributes, $content, $block ) {
+	// Render block in Gutenberg Editor
```
dkotter:
```suggestion
	// Render block in Gutenberg Editor.
```

includes/Classifai/Blocks/recommended-content-block/register.php
```diff
@@ -0,0 +1,48 @@
...
+		[
+			'render_callback' => $n( 'render_block_callback' ),
+		]
+	);
+}
+
+/**
+ * Render callback method for the block
+ *
+ * @param array  $attributes The blocks attributes
```
dkotter:
```suggestion
 * @param array  $attributes The blocks attributes.
```

includes/Classifai/Blocks/recommended-content-block/register.php
```diff
@@ -0,0 +1,48 @@
...
+ * @param string $content    Data returned from InnerBlocks.Content
```
dkotter:
```suggestion
 * @param string $content    Data returned from InnerBlocks.Content.
```

dkotter:
Following up on my [last comment](https://github.com/10up/classifai/pull/343#issuecomment-1172507197), we discussed this more internally and while I think there might be better use cases for Personalizer (especially due to some of the limitations) I do think it's worth getting this particular feature released and see if we can gather some real-world feedback.

A couple additional thoughts/questions:

1. We'll want to update the README.md file with information about this new feature (and probably the caveats). We'll want to call out this as being a new feature and I think we'll want steps on how to get this setup, since you need to create a Personalizer service instead of a Computer Vision service.
2. Do we want to rename this feature or are we fine with `Recommended Content`? I think we'll definitely want to update the block description to make it clear that it only renders a single item
3. Should we output some sort of heading within the block markup? Right now we just render the one content item based on the settings chosen. Wondering if we want some sort of `Recommended` heading to make a clear distinction between this and the regular content. Or do we just want to rely on users adding that themselves?
4. I'm curious how this works on sites that get little to no traffic. For instance on my local, how does it decide which piece of content to show? Is it just random out of the 50 items we send? Seems I get the same post each time (even after clearing localstorage). And I notice in the admin preview, anytime I change a setting it re-renders the block and seems to show me a new post each time. I could see this confusing people, not sure if we need some sort of way to make it clear that the content shown in the admin is just an example and may not be the same as the item that gets shown on the front-end?

dkotter:
Sorry, one other question. When someone clicks on a recommended content item, we send that information to Personalizer. Is that the only time we send information to them? I guess just curious if there's additional information/tracking we should be sending to them so we get better results.

includes/Classifai/Providers/Azure/Personalizer.php
iamdharmesh: Nice catch. I had copied this from `ComputerVision`, updated it there as well.

includes/Classifai/Blocks/recommended-content-block/frontend.js
iamdharmesh: Done. Thanks for a note here.

package.json
iamdharmesh: Migrated to `wp-scripts` now.

includes/Classifai/Services/Personalizer.php
iamdharmesh: Done.

includes/Classifai/Providers/Azure/Personalizer.php
iamdharmesh: We are not using this now. however, we already updated wp minimum to 5.6 in other PR.

iamdharmesh:
@dkotter Thanks for a detailed review here (Sorry for some formatting issues, not sure why those issues are not detected by phpcs). I have updated the codebase as per the review feedback. 



> 1. We'll want to update the README.md file with information about this new feature (and probably the caveats). We'll want to call out this as being a new feature and I think we'll want steps on how to get this setup, since you need to create a Personalizer service instead of a Computer Vision service.

Sure, @jeffpaul can you please help to azure account details. So I can write down the steps for creating an account and grabbing the keys from there. we may also have to provide some information to users on the suggested configuration to make the plugin work more effectively as I mentioned in PR description. 

> Do we want to rename this feature or are we fine with `Recommended Content`? I think we'll definitely want to update the block description to make it clear that it only renders a single item

Yes, agree. We should update the name to something like `Read This Next` as suggested by @jeffpaul, any suggestion from your side @dkotter ?

>3. Should we output some sort of heading within the block markup? Right now we just render the one content item based on the settings chosen. Wondering if we want some sort of `Recommended` heading to make a clear distinction between this and the regular content. Or do we just want to rely on users adding that themselves?

We can add a heading as a part of the block, but I followed what core blocks do here, like `Latest posts`, `Query Loop` etc... it only render a list of posts without heading. Also, keeping the heading separate from the block and allow users to add heading will be better as per me, because it will give them more control on modifying the style, size, etc... of the heading.

>4. I'm curious how this works on sites that get little to no traffic. For instance on my local, how does it decide which piece of content to show? Is it just random out of the 50 items we send? Seems I get the same post each time (even after clearing localstorage). And I notice in the admin preview, anytime I change a setting it re-renders the block and seems to show me a new post each time. I could see this confusing people, not sure if we need some sort of way to make it clear that the content shown in the admin is just an example and may not be the same as the item that gets shown on the front-end?

I am not sure on behind scene process of how the personalizer decided the rank of the article and didn't find any docs on that as well. maybe we can add notice at some place in block?

>  When someone clicks on a recommended content item, we send that information to Personalizer. Is that the only time we send information to them? I guess just curious if there's additional information/tracking we should be sending to them so we get better results.

Personalizer only has `reward` API to send data about user interaction (which also only accept 1 parameter). So, apart from reward API we don't have any other API to send user data.
API Doc: https://westus2.dev.cognitive.microsoft.com/docs/services/personalizer-api/operations/Rank

This demo https://personalizercontentdemo.azurewebsites.net/ has a similar use case (shows a personalized article) with some information on how it works, API request, and response information.

Please let me know if you need any further information on this.

Thanks

README.md
```diff
@@ -36,7 +36,7 @@
 
 ## Requirements
 
-* PHP 7.0+
+* PHP 7.2+
```
dkotter, line 45: Curious on why we need to bump to 7.2?

includes/Classifai/Providers/Azure/Personalizer.php
```diff
@@ -0,0 +1,604 @@
...
+		$content = $this->get_recommended_content( $attributes );
+
+		if ( empty( $content ) || empty( $content->rewardActionId ) ) { // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
+			return $content;
+		}
+
+		// phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
+		$post_id   = $content->rewardActionId;
+		$event_id  = $content->eventId; // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
+		$post      = get_post( $post_id );
```
dkotter: While unlikely, there's a chance here that no post object is found here, so we should add a check to ensure we actually have a proper post object before we continue

includes/Classifai/Providers/Azure/Personalizer.php
```diff
@@ -0,0 +1,604 @@
...
+		$post_link = esc_url( get_permalink( $post ) );
```
dkotter: We don't need the `esc_url` here as it's used later

includes/Classifai/Providers/Azure/Personalizer.php
```diff
@@ -0,0 +1,604 @@
...
+		$class = 'wp-block-classifai-recommended-content';
+
+		if ( isset( $attributes['displayPostDate'] ) && $attributes['displayPostDate'] ) {
+			$class .= ' has-dates';
+		}
+
+		if ( isset( $attributes['displayAuthor'] ) && $attributes['displayAuthor'] ) {
+			$class .= ' has-author';
+		}
+		return sprintf(
```
dkotter: Wondering if we want a filter somewhere around here, if someone wanted to completely change the HTML output? Maybe even earlier on in this method and we can short-circuit early if the filter returns some value?

includes/Classifai/Providers/Azure/Personalizer.php
```diff
@@ -0,0 +1,604 @@
...
+					$features[ $taxonomy->label ][ $term->name ] = 1;
+				}
+			}
+		}
+
+		return $features;
+	}
+
+	/**
+	 * Get user agent for personilizer contextFeatures.
```
dkotter:
```suggestion
	 * Get user agent for personalizer contextFeatures.
```

includes/Classifai/Providers/Azure/Personalizer.php
```diff
@@ -0,0 +1,604 @@
...
+		// User Agent Parsing
+		$parser    = Parser::create();
+		$ua_result = $parser->parse( $user_agent );
+
+		return array(
+			'_ua'           => $ua_result->originalUserAgent, // phpcs:ignore WordPress.NamingConventions.ValidVariableName.UsedPropertyNotSnakeCase
+			'_DeviceFamily' => $ua_result->device->family,
+			'_OSFamily'     => $ua_result->os->family,
+			'DeviceBrand'   => $ua_result->device->brand,
+			'DeviceModel'   => $ua_result->device->model,
```
dkotter, line 592: Do we need any escaping here?

includes/Classifai/Providers/Azure/Personalizer.php
```diff
@@ -0,0 +1,604 @@
...
+					'Ocp-Apim-Subscription-Key' => $settings['api_key'],
+					'Content-Type'              => 'application/json',
+				],
+				'body'    => wp_json_encode( $rank_request ),
+			]
+		);
+
+		if ( ! is_wp_error( $result ) ) {
+			$response = json_decode( wp_remote_retrieve_body( $result ) );
+			set_transient( 'classifai_azure_personalizer_rank_response', $response, DAY_IN_SECONDS * 30 );
```
dkotter: So we cache the response here but doesn't appear we ever use this?

includes/Classifai/Services/Personalizer.php
```diff
@@ -0,0 +1,134 @@
...
+	/**
+	 * Render Recommended content over AJAX.
+	 *
+	 * @return void
+	 */
+	public function ajax_render_recommended_content() {
+		check_ajax_referer( 'classifai-recommended-block', 'security' );
+
+		if ( ! isset( $_POST['contentPostType'] ) || empty( $_POST['contentPostType'] ) ) {
+			esc_html_e( 'No Results found.', 'classifai' );
```
dkotter:
```suggestion
			esc_html_e( 'No results found.', 'classifai' );
```

includes/Classifai/Services/Personalizer.php
```diff
@@ -0,0 +1,134 @@
...
+			if ( isset( $this->provider_classes[0] ) ) {
+				$response = $this->provider_classes[0]->personalizer_send_reward( $event_id );
+			}
+		}
+
+		return $response;
+	}
+
+	/**
+	 * Maybe Clear save transient for recent actions.
```
dkotter:
```suggestion
	 * Maybe clear transients for recent actions.
```

dkotter (changes requested):

includes/Classifai/Providers/Azure/Personalizer.php
```diff
@@ -0,0 +1,604 @@
...
+	 */
+	protected function get_recent_actions( $attributes ) {
+		$post_type      = $attributes['contentPostType'];
+		$key_attributes = array(
+			'terms' => isset( $attributes['taxQuery'] ) ? $attributes['taxQuery'] : array(),
+		);
+		$transient_key  = 'classifai_actions_' . $post_type . md5( maybe_serialize( $key_attributes ) );
+		$actions        = get_transient( $transient_key );
+		if ( false === $actions ) {
+			$query_args = array(
```
dkotter, line 193: I'm not seeing this handled anywhere else but we will want to ensure that we don't ever recommend the current item that is being displayed. Not sure if that means we exclude that here (`post__not_in`?) or remove it later but I think with the way this is set up right now, you could end up showing a recommended piece of content on that same piece of content itself.

dkotter (changes requested):

includes/Classifai/Providers/Azure/Personalizer.php
iamdharmesh: Nice catch. This is leftover code, it was initially added to show it in site health info. but later I displayed service status info, removed now.

README.md
iamdharmesh: The lib ([`ua-parser/uap-php`](https://github.com/ua-parser/uap-php)) we are using for UA parser requires PHP >= 7.2

includes/Classifai/Providers/Azure/Personalizer.php
iamdharmesh: I think it is fine to use this without escaping, as these data come from parser lib and we are not storing or showing it anywhere, just sending request to the personalizer with this data.

includes/Classifai/Providers/Azure/Personalizer.php
iamdharmesh: Added filter on markup and block attributes.

includes/Classifai/Providers/Azure/Personalizer.php
iamdharmesh: Code updated to display multiple tickets now. So, not using this code anymore.

dkotter (dismissed):

dkotter:
This is looking good to me, thanks for all the work here @iamdharmesh! Looks like some merge conflicts that need resolved.

Also, I think still needing updates to our readme around this new feature and how to get it set up. I think you may be waiting on something from @jeffpaul, at least according to [this comment](https://github.com/10up/classifai/pull/343#issuecomment-1177521049).

iamdharmesh:
Thanks for the review @dkotter, I am made some minor improvements and resolved conflicts.
> Also, I think still needing updates to our readme around this new feature and how to get it set up. I think you may be waiting on something from @jeffpaul, at least according to https://github.com/10up/classifai/pull/343#issuecomment-1177521049.

I have updated readme based on docs available online, maybe @jeffpaul can help to verify it once. 

Do you think we should release this feature as Alpha/Beta for now by adding notice somewhere? also, We have some further stuff (mentioned in PR description) to work on it once this get merged.

Thanks.

dkotter (dismissed):

dkotter:
I think this is good to go as a first version of this.

> Do you think we should release this feature as Alpha/Beta for now by adding notice somewhere

We could note this in the README, that this is a feature that we're still gathering feedback on, but not sure if that's critical

jeffpaul:
@iamdharmesh I just did a first though thorough pass on copy and such, but would love to test this out a bit more closely.  Mind spinning this PR out into a test site for me to see if there are other places to tweak the UX/copy/etc?

jeffpaul (approved):
I made some tweaks to copy throughout and added a screenshot and content to the readme, all minor changes and otherwise this looks great.  Fantastic work @iamdharmesh!

PR #268, opened 2021-02-26
Add option to do language processing for existing content in Classic Editor

thrijith:
https://github.com/10up/classifai/issues/202
### Description of the Change

Allows generating tags for published post content without having to save the post to do so.

### Benefits

- Tag generation can be done on existing post without having to `Update` the post, doesn't trigger normal post save actions.

### Possible Drawbacks

- Block editor requires a refresh once tags are generated, don't see an option that updates the metabox value programmatically.
- Trying to update records based on generated terms, breaks the editor using following snippet.
- 
```
wp.data.dispatch( 'core' ).editEntityRecord(
            'postType',
            currentPost.type,
            currentPost.id,
            { [e]: response[e] } // [e] is taxonomy name, response[e] are ids of the new terms
        );
```

### Verification Process

- Tested language processing in Classic Editor and Block Editor

### Checklist:

- [x] I have read the [**CONTRIBUTING**](/CONTRIBUTING.md) document.
- [x] My code follows the code style of this project.
- [x] My change requires a change to the documentation.
- [x] I have updated the documentation accordingly.
- [ ] I have added tests to cover my change.
- [ ] All new and existing tests passed.

### Applicable Issues

Fixes https://github.com/10up/classifai/issues/202
### Changelog Entry

- Add support to process language for existing content

PR #366, opened 2022-09-09
Added "Classify Post" button to process existing content.

### LINKED ISSUE #202: Process existing content via Language Processing without having to update content
We referenced this concept in #121 but didn't get to it in v1.5 and this relates a bit to the work we're working to confirm approach on in #185, but it would be nice to have options similar to what landed in v1.5 via #135 wherein we could request content tagging via our Language Processing features without having to update an individual piece of content (which is what is currently required to trigger the language processing).

- [ ] button on individual Post / Page screen to process content via Language Processing

**Designs**

TBD

n/a

**Additional context**

n/a

iamdharmesh:
### Description of the Change
This PR is a continuation of PR #268 (derives some code from PR #268) and implements the requested functionality in #202, it adds support for the process content of the existing posts. It adds the "Classify Post" button in the ClassifAI sidebar panel. The button loads conditionally, it displays only when `Process content on update` is disabled. Please check the attached video for more details on this.

https://user-images.githubusercontent.com/10613171/190631497-e3f35833-5a21-4d7b-a4c2-0c800574090b.mp4

Closes #202 

### How to test the Change
1. Create a few posts without activating the ClassifAI plugin.
2. Activate the ClaissifAI plugin and language processing.
3. Go to edit post page and open the ClassifAI sidebar panel.
4. Disable the `Process content on update` option
5. Make sure a "Classify Post" button is visible now. (`Post` text will be changed for the post type. for Page, it will be "Classify Page")
6. Click it, Verify that terms are generated properly for each taxonomy.
7. Test "Classify Post" button for the classic editor as well. 

### Changelog Entry

> Added - "Classify Post" button to process existing content.

### Credits
Props @thrijith @iamdharmesh @dkotter @dinhtungdu @jeffpaul 

### Checklist:


includes/Classifai/Providers/Watson/NLU.php
```diff
@@ -815,13 +817,22 @@ public function render_classifai_meta_box( $post ) {
...
 		<p>
 			<label for="_classifai_process_content">
-				<input type="checkbox" value="yes" id="_classifai_process_content" name="_classifai_process_content" <?php checked( $classifai_process_content, 'yes' ); ?>/>
+				<input type="checkbox" value="yes" id="_classifai_process_content" name="_classifai_process_content" <?php checked( $classifai_process_content, 'yes' ); ?> />
 				<?php esc_html_e( 'Process content on update', 'classifai' ); ?>
 			</label>
 		</p>
+		<a href="<?php echo esc_url( wp_nonce_url( admin_url( 'admin-post.php?action=classifai_classify_post&post_id=' . $post->ID ), 'classifai_classify_post_action', 'classifai_classify_post_nonce' ) ); ?>" class="button button-classify-post">
+			<?php printf( esc_html__( 'Classify %s', 'classifai' ), esc_html( $post_type_label ) ); ?>
+		</a>
```
peterwilsoncc: This always displays the button in the classic editor, regardless of the setting of the checkbox. If the button can be hidden so it's consistent with the block editor, that would be helpful.

Screen Shot 2022-09-19 at 10 15 03 am

includes/Classifai/Services/LanguageProcessing.php
```diff
@@ -13,4 +15,91 @@ class LanguageProcessing extends Service {
...
+					'id' => array(
+						'required'          => true,
+						'type'              => 'integer',
+						'sanitize_callback' => 'absint',
+						'description'       => esc_html__( 'Post ID to generate tags.', 'classifai' ),
+					),
+				),
+				'permission_callback' => function () {
+					return current_user_can( 'edit_posts' );
+				},
```
peterwilsoncc: This needs to be a little more complex.

The correct code to determine if a user can edit a single post is `current_user_can( 'edit_post', $post_id )`.

However, the button is acutally used to assign terms so it might need to be a little more complex to determine if the user is allowed to do so. There is a meta capability `current_user_can( 'assign_term', $term_id )` for the purpose but the use case here is a little odd. Technically the user is not assigning the term, the bot is. 

The ultimate question is: should the user be have permission to edit the post, assign the terms or both? @dkotter might have a view here. The answer to this will also depend on how the tags are assigned -- if you save via the REST API post endpoint, then you should match the permissions it uses.

src/js/gutenberg-plugin.js
```diff
@@ -54,6 +55,116 @@ const ClassifAIToggle = () => {
 	);
 };
 
+/**
```
peterwilsoncc, line 58: It would be good to show a notice in the block editor similar to that shown in the classic editor when the button is clicked.

See the [notices API](https://developer.wordpress.org/block-editor/reference-guides/data/data-core-notices/) in the block editor handbook. There are helpers for creating various types of notices, success, failure, etc.

Notices can be either along the top or in snackbar format (similar to the "Post published" text WP Core uses).

includes/Classifai/Services/LanguageProcessing.php
```diff
@@ -13,4 +15,91 @@ class LanguageProcessing extends Service {
...
+
+			$post_type     = get_post_type( $post_id );
+			$post_status   = get_post_status( $post_id );
+			$supported     = \Classifai\get_supported_post_types();
+			$post_statuses = \Classifai\get_supported_post_statuses();
+
+			// Check if processing allowed.
+			if ( ! in_array( $post_status, $post_statuses, true ) || ! in_array( $post_type, $supported, true ) || ! \Classifai\language_processing_features_enabled() ) {
+				return new \WP_Error( 'not_enabled', esc_html__( 'Language Processing not enabled for current post.', 'classifai' ) );
+			}
```
peterwilsoncc: Some, if not all, of this can be in the permission check.

As the permission check is becomming complex, you might need to move it in to a method rather than use an enclosure.

src/js/gutenberg-plugin.js
```diff
@@ -1,10 +1,11 @@
 /* eslint-disable no-unused-vars */
-import { ReactComponent as icon } from '../../assets/img/editor-icon.svg';
+import { ReactComponent as icon } from '../../assets/img/block-icon.svg';
+import { handleClick } from './helpers';
```
peterwilsoncc, line 3: `lodash` is now a dependency of this file and should be added to the script registration.

https://github.com/10up/classifai/blob/1f668e6c383e9bd69d72f63c15ed9c2277013ce9/includes/Classifai/Providers/Watson/NLU.php#L198-L204

includes/Classifai/Admin/SavePostHandler.php
```diff
@@ -214,6 +215,19 @@ public function show_error_if() {
 			</div>
 			<?php
 		}
+
+		// Display classify post success message for manually classified post.
+		// phpcs:ignore WordPress.Security.NonceVerification.Recommended
+		$classified = isset( $_GET['classified'] ) ? intval( wp_unslash( $_GET['classified'] ) ) : 0;
```
peterwilsoncc: `classified` should be added to the `removable_query_args` filter.

```
add_filter( 'removable_query_args', function( $removable_query_args ) {
   $removable_query_args[] = 'classified';
  return $removable_query_args;
} );
```

to avoid the potential for a collision, it might be worth being more specific with something like `classifai_reclassify` -- it's best to namespace all the things, not just function names :)

peterwilsoncc (changes requested):
Generally the code looks good but I've added a few notes inline.

The most important thing is the permission check.

includes/Classifai/Services/LanguageProcessing.php
dkotter: This is a tricky one. Right now, if Language Processing is turned on, we automatically classify content on save, without ever checking what permissions a user has, I think because the user is not actually assigning terms, the AI/bot is doing that.

We also are currently using [`current_user_can( 'edit_posts' )`](https://github.com/10up/classifai/blob/develop/includes/Classifai/Providers/Watson/NLU.php#L213) to determine if we should show that Classifai panel in Gutenberg, which I'm assuming is why were doing the same here.

This might be worth a broader discussion on permissions within Classifai but feels like if we want to keep the same behavior, we don't necessarily care about a user's permission to assign terms, just if they have permission to edit the piece of content. So I think `edit_post` should suffice here.

includes/Classifai/Services/LanguageProcessing.php
iamdharmesh: Done. :+1:

includes/Classifai/Services/LanguageProcessing.php
iamdharmesh: I updated code to use `current_user_can( 'edit_post', $post_id )`, same at display Classifai panel as well.

iamdharmesh:
@peterwilsoncc Thanks for the detailed review, I have made suggested changes. please help with review changes. 

Thanks.

includes/Classifai/Admin/SavePostHandler.php
```diff
@@ -248,4 +271,36 @@ public function is_rest_route() {
...
+				$result     = $this->classify( $post_id );
+				$classified = array();
+				if ( ! is_wp_error( $result ) ) {
+					$classified = array( 'classifai_classify' => 1 );
+				}
+				wp_safe_redirect( esc_url_raw( add_query_arg( $classified, get_edit_post_link( $post_id, 'edit' ) ) ) );
+				exit();
+			}
+		} else {
+			die( esc_html__( 'You don\'t have access to doing this operations.', 'classifai' ) );
```
peterwilsoncc:
```suggestion
			wp_die( esc_html__( 'You don\'t have permission to perform this operation.', 'classifai' ) );
```

`wp_die()` is a little nicer looking than `die()`

src/js/gutenberg-plugin.js
```diff
@@ -54,6 +55,130 @@ const ClassifAIToggle = () => {
...
+					postId
+				);
+			}
+
+			// Display success notice.
+			dispatch( 'core/notices' ).createSuccessNotice(
+				`${ postTypeLabel } ${ __(
+					'classified successfully.',
+					'classifai'
+				) }`,
```
peterwilsoncc: This isn't translatable as it presumes a language structure that may not apply to some languages.

```suggestion
				sprintf(
					/** translators: %s is post type label. */
					__( '%s classified successfully.', 'classifai' ),
					postTypeLabel
				) }`,
```

Please make sure `sprintf` is imported from `wp.i18n` and check my code, I haven't verified it for typos. 

To reduce the number of strings for translators, this duplicates the PHP version of the string:

https://github.com/10up/classifai/blob/919f3b915d39e12009c7d85888371055e6fb0a22/includes/Classifai/Admin/SavePostHandler.php#L232-L235

src/js/gutenberg-plugin.js
```diff
@@ -54,6 +55,130 @@ const ClassifAIToggle = () => {
...
+	const enabled = 'no' === processContent ? 'no' : 'yes';
+	if ( 'yes' === enabled ) {
+		return null;
+	}
+
+	const postId = wp.data.select( 'core/editor' ).getCurrentPostId();
+	const postTypeLabel =
+		wp.data.select( 'core/editor' ).getPostTypeLabel() ||
+		__( 'Post', 'classifai' );
+	const buttonText = `${ __( 'Classify', 'classifai' ) } ${ postTypeLabel }`;
```
peterwilsoncc: Similar to my earlier comment, this should use `sprintf()`

```suggestion
	/** translators: %s Post type label */
	const buttonText = sprintf( __( 'Classify %s', 'classifai' ), postTypeLabel );
```

Again, it shares the string with the existing PHP version.

https://github.com/10up/classifai/blob/919f3b915d39e12009c7d85888371055e6fb0a22/includes/Classifai/Providers/Watson/NLU.php#L835

includes/Classifai/Providers/Watson/NLU.php
```diff
@@ -815,13 +817,24 @@ public function render_classifai_meta_box( $post ) {
...
 		<p>
 			<label for="_classifai_process_content">
-				<input type="checkbox" value="yes" id="_classifai_process_content" name="_classifai_process_content" <?php checked( $classifai_process_content, 'yes' ); ?>/>
+				<input type="checkbox" value="yes" id="_classifai_process_content" name="_classifai_process_content" <?php checked( $classifai_process_content, 'yes' ); ?> />
 				<?php esc_html_e( 'Process content on update', 'classifai' ); ?>
 			</label>
 		</p>
+		<div class="classifai-clasify-post-wrapper" style="display: none;">
+			<a href="<?php echo esc_url( wp_nonce_url( admin_url( 'admin-post.php?action=classifai_classify_post&post_id=' . $post->ID ), 'classifai_classify_post_action', 'classifai_classify_post_nonce' ) ); ?>" class="button button-classify-post">
+				<?php printf( esc_html__( 'Classify %s', 'classifai' ), esc_html( $post_type_label ) ); ?>
```
peterwilsoncc:
```suggestion
				<?php
				/** translators: %s Post type label */
				printf( esc_html__( 'Classify %s', 'classifai' ), esc_html( $post_type_label ) );
				?>
```

peterwilsoncc (changes requested):
The capability and classic editor changes look great, thank you!

I've added a few notes about translations to ensure the strings are tranlatable to languages that place post type labels elsewhere. @ravinderk may have suggestions too, he has offered to assist in Distributor and, unlike me, speaks more than one language.

Please test my suggestions as I haven't checked them for typos.

src/js/gutenberg-plugin.js
iamdharmesh: Thanks a lot for noticing this.

peterwilsoncc (approved):
Thanks, much appreciated!

PR #374, opened 2022-09-30
feat/278: Add settings fields to enable storing image captions in non-alt text image fields.

### LINKED ISSUE #278: Option to store Image Captions in alternate fields
I received the following feedback via email:
> The option to automatically Caption Images should have the option to place the result somewhere else, like in the Description instead. I and many others have a very SEO streamlined approach to the Alt. Text and there is no scenario where your AI is going to be good enough for me to take over that function completely. For me, it is a guiding light as to what people might see that I can use for inspiration.

We might consider updating the Image Processing settings for `Automatically Caption Images` to select where that data is stored, with the `Alternative Text` field as the default and `Caption` and `Description` fields as other options.  Possibly even allowing for this to be a multi-select checkbox, but not certain on that part yet.

**Designs**

**Additional context**

Sidsector9:
### Description of the Change

This PR adds settings to add image captions to the following places:

Screenshot 2022-11-27 at 11 58 27 PM

Closes #278 

### How to test the Change

1. Test for a new installation.
2. Test for old installation with previous setup.

### Changelog Entry

> Added - Ability to set alt text as image caption and image description.

### Credits

Props @jeffpaul @Sidsector9 

### Checklist:


jeffpaul:
@Sidsector9 is there more work needed here or can we assign OSP to get this into review?

Sidsector9:
@jeffpaul few minor things I need to take care of. I'll prepare the final PR next week.

includes/Classifai/Services/ImageProcessing.php
```diff
@@ -39,7 +39,14 @@ public function init() {
...
 	public function enqueue_media_scripts() {
-		wp_enqueue_script( 'media-script', CLASSIFAI_PLUGIN_URL . '/dist/media.js', array( 'jquery', 'media-editor', 'lodash', 'wp-i18n' ), CLASSIFAI_PLUGIN_VERSION, true );
+		wp_enqueue_script( 'classifai-media-script', CLASSIFAI_PLUGIN_URL . '/dist/media.js', array( 'jquery', 'media-editor', 'lodash', 'wp-i18n' ), CLASSIFAI_PLUGIN_VERSION, true );
+		wp_localize_script(
+			'classifai-media-script',
+			'classifaiMediaVars',
+			array(
+				'enabledAltTextFields' => $this->provider_classes[0]->get_alt_text_settings() ? $this->provider_classes[0]->get_alt_text_settings() : array(),
+			)
+		);
```
peterwilsoncc: For settings, it's recommended to use `wp_add_inline_script()` instead.

The method `wp_localize_script()` uses [does some type casting](https://github.com/WordPress/wordpress-develop/blob/360f6ed32d12c80a80b61dd2dc781432483ebb9a/src/wp-includes/class-wp-scripts.php#L511-L521) so it's not guaranteed that the data will always be in the form expected.

includes/Classifai/Providers/Azure/ComputerVision.php
```diff
@@ -63,6 +67,29 @@ private function get_default_settings() {
 		];
 	}
 
+	/**
+	 * Returns an array of fields enabled to be set to store image captions.
+	 * Returns `false` if no fields are selected.
+	 *
+	 * @return array|boolean
+	 */
+	public function get_alt_text_settings() {
```
peterwilsoncc, line 76: This needs to account for data stored in the old format.

If I've enabled alt text in an old version a couple of things happen:

* the option is no longer checked when I change to this branch
* This notice is thrown six times: `Undefined index: enable_image_captions` in `includes/Classifai/Providers/Provider.php` (three times each on lines 224 and 225)

lost-setting

peterwilsoncc (changes requested):
I've added a couple of notes inline, there's a bug for sites with data saved in the old format.

It would be good to add a test for the old format option of `enable_image_captions` in computer vision phpunit tests. I think it would start with the following:


```php
	public function test_option_reformatting() {
		add_filter(
			'pre_get_option_classifai_computer_vision',
			function() {
				return array (
					'valid' => false,
					'url' => '',
					'api_key' => '',
					'enable_image_captions' => '1',
					'enable_image_tagging' => '1',
					'enable_smart_cropping' => 'no',
					'enable_ocr' => 'no',
					'enable_read_pdf' => 'no',
					'caption_threshold' => 75,
					'tag_threshold' => 70,
					'image_tag_taxonomy' => 'classifai-image-tags',
				);
			}
		);

		// Get the alt text formatting.
		$settings = $this->get_computer_vision()->get_alt_text_settings();

		// Some assertion to confirm ALT text is enabled.
		// Some assertion to confirm the data is an array.
		// Some assertion to confirm the defaults are as expected for the two new settings.
	}
```

peterwilsoncc:
Sorry, noticed some additional errors:

I was sometimes getting a warning too:

```
Invalid argument supplied for foreach()
    content/plugins/classifai/includes/Classifai/Providers/Azure/ComputerVision.php:84
    Classifai\P\A\ComputerVision->get_alt_text_settings()
    content/plugins/classifai/includes/Classifai/Services/ImageProcessing.php:47
    Classifai\Services\ImageProcessing->enqueue_media_scripts()
    wp-includes/class-wp-hook.php:308
    do_action('admin_enqueue_scripts')
    wp-admin/admin-header.php:118
```

Generate alt text no longer seems appropriate here. "Generate descriptive text" could be better. This applies to the threshold option too.



Screen Shot 2022-11-28 at 3 14 54 pm

tests/Classifai/Providers/Azure/ComputerVisionTest.php
```diff
@@ -98,4 +102,53 @@ public function test_set_image_meta_data() {
...
+			'enable_image_tagging'  => '1',
+			'enable_smart_cropping' => 'no',
+			'enable_ocr'            => 'no',
+			'enable_read_pdf'       => 'no',
+			'caption_threshold'     => 75,
+			'tag_threshold'         => 70,
+			'image_tag_taxonomy'    => 'classifai-image-tags',
+		);
+
+		/** Test with `enable_image_captions` set to `1` */
```
peterwilsoncc:
```suggestion
		// Test with `enable_image_captions` set to `1`
```

A couple of things here:

* `/**` indicated a docblock rather than a comment, however....
* the WPCS for single line comments is to use `//` rather than `/*`

tests/Classifai/Providers/Azure/ComputerVisionTest.php
```diff
@@ -98,4 +102,53 @@ public function test_set_image_meta_data() {
...
+		$this->assertSame(
+			$image_captions_settings,
+			array(
+				'alt'         => 'alt',
+				'caption'     => 0,
+				'description' => 0,
+			)
+		);
+
+		/** Test with `enable_image_captions` set to `no` */
```
peterwilsoncc:
```suggestion
		// Test with `enable_image_captions` set to `no`
```

includes/Classifai/Providers/Azure/ComputerVision.php
```diff
@@ -63,6 +67,37 @@ private function get_default_settings() {
...
+		$settings       = $this->get_settings();
+		$enabled_fields = array();
+
+		if ( ! isset( $settings['enable_image_captions'] ) ) {
+			return false;
```
peterwilsoncc:
```suggestion
			return array();
```

I noticed that when no fields are set, this method sometimes returns `false` and sometimes an empty array. It would be good to be consistent.

This might allow you to simplify `generate_alt_tags()` a little as the method could assume an array and lose the repeated `is_array()` check.

To reproduce:
* delete the option from the database
* var_dump shows false
* save the image processing options page without enabling the descriptive text for any option
* var_dump of this method shows `array()`

includes/Classifai/Providers/Provider.php
```diff
@@ -195,6 +195,60 @@ public function render_select( $args ) {
...
+
+		$checkbox_options = array(
+			'alt'         => esc_html__( 'Alt text', 'classifai' ),
+			'caption'     => esc_html__( 'Image caption', 'classifai' ),
+			'description' => esc_html__( 'Image description', 'classifai' ),
+		);
+
+		foreach ( $checkbox_options as $option_value => $option_label ) {
+			if ( ! is_array( $setting_index['enable_image_captions'] ) ) {
+				$default_value = '1' === $setting_index['enable_image_captions'] ? 'alt' : '';
```
peterwilsoncc: These two lines throw a notice if you visit the page before the options are saved. To reproduce:

* Delete the `classifai_computer_vision` option from the database
* Visit the options page
* A notice will be thrown

Saving the option then caused different errors, continue from above:
* save the options without making any changes
* notices are thrown on line 233 for each item in the loop:
   * Undefined index: alt
   * Undefined index: caption
   * Undefined index: description

peterwilsoncc (commented):
Added a few notes below. I'm still seeing notices and warnings under certain circumstances that it would be good to figure out. 

I think a few additional `isset` checks are needed, or always using the new method when it's possible. Using the new method may not be possible if you need to account for different ai providers though.

includes/Classifai/Providers/Azure/ComputerVision.php
```diff
@@ -770,8 +827,8 @@ public function setup_fields_sections() {
 		);
 		add_settings_field(
 			'enable-image-captions',
-			esc_html__( 'Generate alt text', 'classifai' ),
-			[ $this, 'render_input' ],
+			esc_html__( 'Generate descriptive text', 'classifai' ),
```
peterwilsoncc, line 830: `Alt text confidence threshold` will need to become `Descriptive text confidence threshold` in the next option.

.eslintrc.json
```diff
@@ -10,7 +10,8 @@
 		"Cypress": "readonly",
 		"cy": "readonly",
 		"ajaxurl": "readonly",
-		"FormData": "readonly"
+		"FormData": "readonly",
+		"classifaiMediaVars": "readonly"
```
peterwilsoncc, line 14: `classifaiMediaVars` was removed in 1bfe51cc0a1b71187cefad44aefc28f18181a064

peterwilsoncc (commented):
A couple more notes inline.

PHP notices thrown in previous testing have been resolved, thank you! It would be good to add a test for the deleted option as that was causing problems earlier.

.eslintrc.json
Sidsector9: @peterwilsoncc the PR wasn't up for re-review. I re-added `classifaiMediaVars` in  [ba3162a](https://github.com/10up/classifai/pull/374/commits/ba3162a273704b85f6efb4e105343d95572f6f56)

peterwilsoncc (commented):
The code all looks good but one thing I noticed is that the feature branch no longer automatically adds the alt text when an image is uploaded via the block editor

I tested this after saving the image processing page using the new settings.

**Develop branch**

develop-branch

**Feature branch**

feature-branch

Sidsector9:
@peterwilsoncc The descriptive texts are generating for me in the block editor.

I tested drag and drop as well as uploading via the image block using the upload button.

Please watch this screencast:


https://user-images.githubusercontent.com/17757960/214793234-885d05cf-3892-47c5-af2b-454fd753c0db.mov

peterwilsoncc (approved):
@Sidsector9 Sorry, not sure what went wrong last time. It seems to be working now.

Possibly I confused WP / the bot by using ngrok -- setting the ngrok tunnel's domain as the site's primary domain worked. I'll keep that in mind for next time.

Issue #392, opened 2023-01-16
[SPIKE] Personalization Research


jeffpaul:
### Is your enhancement related to a problem? Please describe.

Alongside our Recommended Content feature using Azure Personalizer, we should do some market research to see how other products/projects handle user experience personalization, identify what might make for a good fit inside ClassifAI, and then identify what available AI APIs might support this (and where feasible have a free tier to allow folks to test/explore the functionality before committing to a paid service/tier).

### Designs

n/a

### Describe alternatives you've considered

Some existing products/services to review as existing in this space:
- [Altis](https://www.altis-dxp.com/accelerate/personalisation/)
- [Sunbreak](https://sunbreak.io/integrations/personalization-for-wordpress/)

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

faisal-alvi:
A few recommendations worth exploring:

1. [Amazon Personalize](https://aws.amazon.com/personalize/): Amazon Web Services (AWS) offers Amazon Personalize, which provides similar functionality for creating personalized recommendations.
2. [Google Cloud Recommendations AI](https://cloud.google.com/recommendations): Google Cloud's Recommendations AI is designed to help you build personalized product recommendation systems.
3. [Algolia](https://www.algolia.com/): Algolia offers a powerful search and discovery API that can be used for building personalized search and recommendation systems.
4. [Recombee](https://www.recombee.com/): Recombee is a recommendation service with APIs for personalization, recommendation, and user behavior analysis.

jeffpaul:
Perfect timing on the additional options @faisal-alvi as Microsoft recently announced that [Personalizer no longer accepts new customers and will be retired in October 2026](https://learn.microsoft.com/en-us/azure/ai-services/personalizer/).  As @dkotter mentioned elsewhere, [Azure Intelligent Recommendations](https://learn.microsoft.com/en-us/industry/retail/intelligent-recommendations/overview) may be a worthwhile option to consider as well.

jeffpaul:
I did some quick reading into Google Cloud Recommendation AI today and that seems like it could be a solid option.  Some interesting bits in [What is Discovery for Media?](https://cloud.google.com/discovery-engine/media/docs/introduction) and [Build a Recommendation System using Google Cloud Recommendation AI](https://towardsdatascience.com/build-a-recommender-system-using-google-cloud-c0929f0c3080).

That second item has some rather relevant integrations it mentions, all of which would seem to be worthwhile within ClassifAI:
[...]

PR #397, opened 2023-02-13
Content Recommend block setting update

### LINKED ISSUE #379: Recommended Content settings iteration

The current Recommended Content block takes input from a user via several controls, two of which are the Categories and Tags multiselect boxes.  Let's look into an iteration here that:

1. Sets the `Post type` field to the same post type for the currently edited post type (e.g., if editing on a Page, then set the `Post type` to `Page`).
2. Set the `Categories` and `Tags` to be the same as the currently edited post type (e.g., if editing on a Page that's set as `Categories`=`Sports`, `Tags`=`[Soccer, Technology]`, then set those categories and tags as the default ones for the Recommended Content block).
3. Add a `Use this Post's Categories and Tags` toggle that defaults to enabled and sets those fields based on #2 above.  This makes those fields uneditable (or alternatively hides them).  Disabling this toggle makes the `Categories` and `Tags` fields editable.

Props to @jakemgold for the inspiration on this UX improvement.

### Designs

Use This Posts Categories And Tags setting enabled and showing full view of pre-set fields:
UseThisPostsCategoriesAndTags_EnabledFullView

Use This Posts Categories And Tags setting enabled but hiding the pre-set fields / collapsing them away from view:
UseThisPostsCategoriesAndTags_EnabledCollapsed

UseThisPostsCategoriesAndTags_Disabled:
UseThisPostsCategoriesAndTags_Disabled

n/a

### Code of Conduct

jayedul:
### Description of the Change
Content Recommend block setting updated

Closes #379 

### Changelog Entry

> Changed - Post type and taxonomy selection enhanced

### Credits

Props @jayedul, @jakemgold 

### Checklist:


peterwilsoncc:
I did a quick test of this and think it will need to follow the guide for [block deprecation](https://developer.wordpress.org/block-editor/reference-guides/block-api/block-deprecation/) in the block editor handbook.

When the attributes change/the block is altered the deprecation code needs to be added to safely migrate from one form to another.

At the moment, blocks created on the default branch (develop) fail on this new branch -- I presume because the attributes and/or saved data are not in the same format.

testing-branch

peterwilsoncc:
I've also noticed an issue with the new format on the front end:

1. On `develop` create a recommended content block and publish the post (ensure it has results)
2. Visit the post on the front end, the post results display as expected.
   Screen Shot 2023-02-23 at 10 46 47 am
3. Switch to the feature branch for this PR
4. Run `npm run build`
5. No results are displayed
   Screen Shot 2023-02-23 at 10 48 23 am

jayedul:
Post not found issue is fixed. However I couldn't reproduce block crashing issue. Here's a video showing it works here after changing develop to fix branch. Which version of WordPress did you test with? @peterwilsoncc 

https://user-images.githubusercontent.com/12506700/224363249-47d064b8-51cd-49ae-97b5-bcf1805edb36.mp4

jayedul:
BTW, I guess e2e test is failing here because of cypress-wp-utils library. 
Most likely this fix will resolve it. https://github.com/10up/cypress-wp-utils/pull/89/files#diff-a807a69cfe758f8aabc6ea3a6f836df2dceaff610de1bd4384a6ed0afca4dc63R79

jeffpaul:
I'll defer to Peter's re-review here, though once that's approved I will want to do a live test on this PR to validate it updates the settings as expected before we merge this in

peterwilsoncc:
@jayedul I think I was testing against master.

I'll document the full steps. Make sure to run the build step each time you switch branches.

1. checkout `develop`
2. `npm run build`
3. Create a new post
4. Add a recommend content block
5. In the block settings add some tags to relate content by
   Screen Shot 2023-03-14 at 11 09 23 am
6. Save Post
7. Checkout the feature branch
8. `npm run build`
9. Reload the edit post interface
10. The block crashes due to the modifications 

I suspect it's due to the change from `{"taxQuery":{"post_tag":[3,17,18]}}` to `{"contentPostType":"post","taxQuery":{"post_tag":[4]},"defaultTermsSetFor":["post_tag","post_tag","post_tag"]}` in the attribute.

jayedul:
It's made ready for functional review, though eslint test is failing. Locally fixed code by eslint command fails here, which is contradictory.

jayedul:
This ticket can be re-reviewed.

includes/Classifai/Blocks/recommended-content-block/edit.js
```diff
@@ -96,6 +102,65 @@ const RecommendedContentBlockEdit = ( props ) => {
 		},
 	];
 
+	const getPostInfo = ( blockIndex ) => {
+		window.jQuery.getJSON(
```
dkotter, line 126: Kind of confused on why we're making a custom ajax request here using jQuery. Can't we use something like [getEntityRecords](https://developer.wordpress.org/block-editor/reference-guides/data/data-core/#getentityrecords) to get taxonomy information?

includes/Classifai/Blocks/recommended-content-block/block.json
```diff
@@ -18,6 +17,14 @@
       "type": "boolean",
       "default": false
     },
+    "usePostTerms": {
+      "type": "boolean",
+      "default": false
```
dkotter, line 22: If I read the issue correctly, it sounds like we want this enabled by default, not turned off by default

includes/Classifai/Blocks/recommended-content-block/inspector-controls/taxonomy-controls.js
```diff
@@ -4,12 +4,14 @@
...
  */
-import { FormTokenField } from '@wordpress/components';
+import { __ } from '@wordpress/i18n';
+import { FormTokenField, ToggleControl } from '@wordpress/components';
 import { useSelect } from '@wordpress/data';
 import { store as coreStore } from '@wordpress/core-data';
 import { getEntitiesInfo, useTaxonomies } from '../utils';
 
 const termsPerPage = 100;
+export const syncTaxonomies = [ 'category', 'post_tag' ];
```
dkotter: I know the original issue mentioned just categories and tags but I think we allow setting any taxonomies that are assigned to the current post type. Ideal would be to not hardcode these values but support all taxonomies the current post type supports

includes/Classifai/Blocks/recommended-content-block/inspector-controls/taxonomy-controls.js
```diff
@@ -112,14 +122,39 @@ const TaxonomyControls = ( { onChange, query } ) => {
 			[]
 		);
 	};
+
+	const UseTermToggle = () => {
+		return (
+			<ToggleControl
+				label={ __(
+					"Use this Post's Categories and Tags",
```
dkotter: Ideally we don't hardcode the word `Post` here, as this could be another type of content (page or CPT)

includes/Classifai/Blocks/recommended-content-block/markup.php
```diff
@@ -15,8 +15,16 @@
...
-$attr_key = md5( maybe_serialize( $attributes ) );
-$block_id = 'classifai-recommended-block-' . $attr_key;
+// Backward compatibillity for default post type.
+// We can not keep default post type in block.json as it makes impossible to distinguish whether it is previously selected or default one in the editor.
+// Because we've introduced auto selection of post type based on currently edit screen for the first time block component mount.
+if ( empty( $attributes['contentPostType'] ) ) {
+	$attributes['contentPostType'] = 'post';
+}
+
+$attributes = apply_filters( 'classifai_recommended_block_attributes', $attributes );
```
dkotter, line 34: This looks to be a new filter. We should add a filter docblock around this

includes/Classifai/Blocks/recommended-content-block/register.php
```diff
@@ -23,6 +23,47 @@ function register() {
...
+}
+
+/**
+ * Get post info to set default terms based on.
+ *
+ * @return void
+ */
+function get_post_info() {
+	if ( empty( $_GET['classifai_editor_nonce'] ) || ! wp_verify_nonce( sanitize_text_field( $_GET['classifai_editor_nonce'] ), 'classifai_editor_nonce' ) ) {
+		wp_send_json_error( array( 'message' => 'Invalid noncce' ) );
```
dkotter:
```suggestion
		wp_send_json_error( array( 'message' => 'Invalid nonce' ) );
```

dkotter: Probably should wrap this in a translation function as well

includes/Classifai/Blocks/recommended-content-block/register.php
```diff
@@ -23,6 +23,47 @@ function register() {
...
+		exit;
+	}
+
+	$post_id = ! empty( $_GET['post_id'] ) ? (int) $_GET['post_id'] : 0;
```
dkotter: I think we want to use `wp_unslash` on all these `$_GET` variables since WordPress may add slashes (see https://github.com/WordPress/wordpress-develop/blob/c421c7ead6c72ebb448c7a21f6bef3acfaab6a92/src/wp-includes/load.php#L1123-L1132)

dkotter: Probably also want to verify that this post ID actually belongs to a post object

includes/Classifai/Blocks/recommended-content-block/register.php
```diff
@@ -23,6 +23,47 @@ function register() {
...
+	$blocks  = array_values(
+		array_filter(
+			$blocks,
+			function( $block ) {
+				return ! empty( $block['blockName'] ) && 'classifai/recommended-content-block' === $block['blockName'];
+			}
+		)
+	);
+
+	$attributes = isset( $blocks[ $index ] ) ? $blocks[ $index ] : null;
```
dkotter:
```suggestion
	$attributes = $blocks[ $index ] ?? : null;
```

includes/Classifai/Blocks/recommended-content-block/register.php
```diff
@@ -46,3 +87,40 @@ function render_block_callback( $attributes, $content, $block ) {
...
 }
+
+/**
+ * Reset category and taxonomy if usePostTerms is true
+ *
+ * @param array $attributes Block attributes
+ * @return array
+ */
+function use_default_post_terms( $attributes ) {
+	if ( ! empty( $attributes['usePostTerms'] ) && true === $attributes['usePostTerms'] ) {
```
dkotter: I think we can invert the conditions here and return `$attributes` early if the conditions don't match, rather than wrap all this code in a conditional

dkotter (changes requested):

includes/Classifai/Blocks/recommended-content-block/block.json
jayedul: This value controls the switch. By default currently associated terms will be selected automatically on `Content Recommend` block mount. Then user can modify the term selection. However if they enable switch (usePostTerms technically), then term input field will disappear and current post terms will be used in filtering no matter what is in input field. 

Screenshot 2023-04-19 at 7 46 52 AM

jeffpaul:
@dkotter @peterwilsoncc back to you for re-review

includes/Classifai/Blocks/recommended-content-block/block.json
dkotter: So this isn't currently working for me, unless I'm doing something wrong. I have a post that has 2 categories and 1 tag set but when I add the Recommended Content block, neither of those inputs are filled out.

That said, I still think the desired behavior is the `Use assigned terms` should be selected by default. Then if you deselect that, we can either leave the taxonomy options empty or at that point, maybe we set those to match the current values of the post? Any thoughts on how you expect this to work @jeffpaul?

dkotter: @jayedul Talked to Jeff on this and the direction we came to is that the `Use assigned terms` should be selected by default, which is opposite of what we have now. And then if someone deselects that option, we should then pre-fill in whatever terms the post has set (this doesn't seem to be working right now)

jeffpaul:
@10up/open-source-practice pinging the team in hopes of getting this through another round of code review/testing feedback

iamdharmesh:
@jeffpaul It seems there are some [open](https://github.com/10up/classifai/pull/397#discussion_r1165972006) feedback that need to be addressed. Also, we have to manage cases like using recommended block outside of edit posts/pages. For example, adding a block in the footer template, where we don't have terms assigned to auto-select as described [here](https://github.com/10up/classifai/pull/397#discussion_r1195383709). I will try to update PR to address these issues.

jeffpaul:
For cases where there aren't terms to auto-select then the terms should be empty.

jeffpaul:
@ravinderk as you're able during your OSS time, could you please look to handle the code review feedback and any open tasks on this PR to help move it along for final review/merge?  Thanks!

dkotter:
Due to [Personalizer going away](https://github.com/10up/classifai/issues/392#issuecomment-1759809106), thinking we probably just want to close this out.

jeffpaul:
:sad-panda:  I agree.

PR #399, opened 2023-02-27
[WIP] Create Azure/Language provider and connect it to LP Service

### LINKED ISSUE #159: Auto-populate missing meta tags and descriptions

**Designs**
n/a

n/a

**Additional context**
n/a

cadic:
### Description of the Change

Adds Azure/Language provider within LanguageProcessing service

The API keys screen is in Classifai > Language Processing > Microsoft Azure.

The provider will have 2 types of requests:
1. Sync request returns the result immediately in run time. Used for [Key Phrase Extraction](https://learn.microsoft.com/en-us/rest/api/language/2022-10-01-preview/text-analysis-runtime/analyze-text?tabs=HTTP#successful-key-phrase-extraction-request) 
2. Async request sets up the job, which results to be requested later. Will be used for [Abstractive Summarization](https://learn.microsoft.com/en-us/rest/api/language/2022-10-01-preview/text-analysis-runtime/submit-job?tabs=HTTP#successful-submit-abstractive-summarization-task)

The API version is currently `2022-10-01-preview`.

The Abstractive Summarization feature is not available publicly and needs to be [applied for](https://aka.ms/applyforgatedsummarizationfeatures) before we proceed with it.

Closes #159

### Credits

### Checklist:

### DIFF
--- a/includes/Classifai/Providers/Azure/Language.php
+++ b/includes/Classifai/Providers/Azure/Language.php
@@ -1,0 +1,341 @@
+<?php
+/**
+ * Azure Cognitive Tools - Language
+ */
+
+namespace Classifai\Providers\Azure;
+
+use Classifai\Providers\Provider;
+use Classifai\Blocks;
+use WP_Error;
+
+class Language extends Provider {
+
+	/**
+	 * @var string URL fragment to perform the syncronous text analysis request
+	 */
+	protected $endpoint = '/language/:analyze-text?api-version=2022-10-01-preview';
+
+	/**
+	 * @var string URL fragment to access asyncronous text analysis features
+	 */
+	protected $async_endpoint = '/language/analyze-text/jobs?api-version=2022-05-15-preview';
+
+	/**
+	 * Language constructor.
+	 *
+	 * @param string $service The service this class belongs to.
+	 */
+	public function __construct( $service ) {
+		parent::__construct(
+			'Microsoft Azure',
+			'Language',
+			'Language',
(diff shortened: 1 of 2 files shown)


jeffpaul:
@cadic might be good to possibly build off the work @dkotter has in his OpenAI/ChatGPT PR to add this functionality and focus more on ensuring someone can select either OpenAI or Azure for this feature.

jeffpaul:
FYI that I would still like to see this get out of WIP as this would be a good UX experiment for how we handle the same functionality within the plugin via two different service providers.  I might be more pressed to handle this when we have a specific request to have a different service provider for this functionality, it may be best to leave as WIP for now.

In discussion with @dkotter, it may also be better holding on this PR until Azure brings this feature out of beta preview into general public availability.

jeffpaul:
@Sidsector9 this might be a good PR to pick up (or perhaps close and open a subsequent related one) to use as an example of adding a new service provider with the v3 release; thoughts?

Sidsector9:
Since we're adding this to core, it would be a good example for internal developers only. Third-party devs will still need to use hooks that will ship out with v3.

jeffpaul:
@Sidsector9 sorry, I had intended for this to be built post-v3 stabilizing in `develop` (or even released) so that the PR here (or one like it if we want to start anew) could be an example of how to extend ClassifAI post-v3's hooks being available.

Sidsector9:
Closing in favor of PR #786

PR #351, opened 2022-07-03
feat/193: add a previewer in admin settings

Sidsector9:
### Description of the Change

This PR adds a feature to display preview of the taxonomy types for a post.

1. Adds a select dropdown with AJAX-based search input field to select/search a post.
2. Displays the score/relevance and the label in results.
3. Displays color-based bar with a width indicating score for visual clarity.

Closes #193 

### Screencast

https://user-images.githubusercontent.com/17757960/177052864-842194a2-630f-4702-b2fa-5898b9a34558.mov

### Checklist:

- [ ] I have read the [**CONTRIBUTING**](https://github.com/10up/.github/blob/trunk/CONTRIBUTING.md) document.
- [ ] My code follows the code style of this project.
- [ ] My change requires a change to the documentation.
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my change.
- [ ] All new and existing tests passed.

### Changelog Entry

> Added - Preview for Language Processing settings changes

### Credits

Props @Sidsector9

PR #402, opened 2023-03-03
Remove some code in the preview section that isn't needed

dkotter:
### Description of the Change

In #351, a previewer was added to the admin. This makes a request to the `Classifier::get_body` method, passing in some options. The problem here is that this method expects the options array to look like this:

```php
[
    'text' => 'Some Text'
    'language' => 'en',
    'features' => [
        'categories' => [],
        'concepts' => [],
        // etc
    ],
]
```

But all we are passing in is a single array containing features part, which looks like:
```php
[
    'categories' => [],
    'concepts' => [],
    // etc
]
```

These values end up being ignored because they don't follow the format we expect. In addition, these values are the same as the default values in the `Classifier::get_body` method, so there's no real reason I can see to pass those in the first place. I discovered this while using the preview section to quickly verify a new feature and found out the values I was passing in weren't being recognized. 

This PR removes those and also changes the default `limit` value to match how we set the limit in other places (using our constant first). An alternative approach here is to change the format of the data we pass in to match what we expect but because we just pass in default values, doesn't seem to add any value currently.

### How to test the Change

Ensure the preview area still functions as expected

### Changelog Entry

> Fixed - Removed some unnecessary code in the preview feature

### Credits

Props @dkotter

### Checklist:

### DIFF
--- a/includes/Classifai/Admin/PreviewClassifierData.php
+++ b/includes/Classifai/Admin/PreviewClassifierData.php
@@ -28,17 +28,8 @@ class PreviewClassifierData {
 		$classifier = new \Classifai\Watson\Classifier();
 		$normalizer = new \Classifai\Watson\Normalizer();
 
-		$features['categories'] = (object) [];
-		$features['concepts']   = (object) [];
-		$features['entities']   = (object) [];
-		$features['keywords']   = [
-			'emotion'   => false,
-			'sentiment' => false,
-			'limit'     => defined( 'WATSON_KEYWORD_LIMIT' ) ? WATSON_KEYWORD_LIMIT : 10,
-		];
-
 		$text_to_classify        = $normalizer->normalize( $post_id );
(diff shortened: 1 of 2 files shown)


jeffpaul:
@dkotter as long as the end result is still "user selects a sample post and can preview the results that Watson returns based on the enabled features & thresholds" then that's fine by me

dkotter:
@jeffpaul Yeah, this doesn't change any functionality, everything should still work exactly the same. Just removes code that wasn't actually being used because it was in the wrong format. Could fix the format so it is used but since it's exactly the same as the default data, seemed better to just remove it.

jeffpaul:
@Sidsector9 would be great to wrap up review here so this can get merged before end of month when I'd like to get out the next major release

Sidsector9 (approved):
Approving as the tests are failing not because of the changes in this PR.

PR #403, opened 2023-03-06
feat/115: Add Microsoft Azure Text to speech capability

### LINKED ISSUE #115: Integrate Azure Text to Speech for "read-to-me" functionality on blog posts
Azure's [Text to Speech](https://azure.microsoft.com/en-us/services/cognitive-services/text-to-speech/) is able to convert text to audio in near real time, play it back, and save it as a file for later use.  This could be leveraged to assist with accessibility improvements for text-only/text-heavy content so that it could be available as audio content.

**Designs**
n/a

This functionality exists via AWS Labs and WP Engine's [Amazon Polly plugin](https://wordpress.org/plugins/amazon-polly/), but would mimic that utilizing Azure instead of AWS.

**Additional context**
none

Sidsector9:
### Description of the Change

Closes #115

### How to test the Change

1. Visit `/wp-admin/admin.php?page=language_processing&tab=azure_text_to_speech`.
2. Setup the endpoint and key. 
3. Select post types which should support this feature.
4. Select a voice.

Screenshot 2023-03-31 at 2 20 41 PM

5. Add a new Post/Page, whichever has support for the feature.
6. The feature will be "on" by default if the post type supports it.

Screenshot 2023-03-31 at 2 22 37 PM

7. Save the post and 

classifai mylocal_wp-admin_post php_post=581 action=edit

8. New controls such as **Refresh** and **Preview** will be visible. Refresh can be used if you decide to change the voice.
9. Go to the frontend, the post will have a play button added using `the_content` filter.

Screenshot 2023-03-31 at 2 26 14 PM

### Changelog Entry

> Added - Text to Speech capability using Microsoft Azura Text to Speech 

### Credits

Props @jeffpaul @Sidsector9 

### Checklist:


src/js/gutenberg-plugin.js
```diff
@@ -296,6 +186,195 @@ const ClassifAIGenerateTagsButton = () => {
...
+
+	// Fetches the latest audio file to avoid disk cache.
+	let cacheBustingUrl = `${ sourceUrl }?ver=${ timestamp }`;
+
+	return (
+		<>
+			<ToggleControl
+				label={ __( 'Generate audio for this post.' ) }
+				help={ isFeatureSupported
+					? __( 'ClassifAI will generate audio for the post when it is published or updated.' )
```
ravinderk: @Sidsector9 did we leave these string untranslatable intentionally?

Sidsector9: @ravinderk  The PR is not up for review yet, it is marked as draft. Please do not pick it up for review until it is ready.

includes/Classifai/Services/LanguageProcessing.php
```diff
@@ -208,4 +245,27 @@ public function generate_post_excerpt_permissions_check( WP_REST_Request $reques
...
+	/**
+	 * Check if a given request has access to generate audio for the post.
+	 *
+	 * @param WP_REST_Request $request Full data about the request.
+	 * @return WP_Error|bool
+	 */
+	public function speech_synthesis_permissions_check( $request ) {
+		$post_id = $request->get_param( 'id' );
+
+		if ( ! empty( $post_id ) ) {
```
iamdharmesh: I think we should also add a check here to verify if the user has permission to edit the post. what do you think?

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,553 @@
...
+	}
+
+	/**
+	 * Returns the default settings.
+	 */
+	private function get_default_settings() {
+		return [
+			'url'           => '',
+			'api_key'       => '',
+			'voices'        => array(),
```
iamdharmesh, line 425:
```suggestion
			'voices'        => '',
```
To prevent this initial notice:
Screenshot 2023-03-31 at 3 02 20 PM

includes/Classifai/Admin/SavePostHandler.php
```diff
@@ -167,6 +170,121 @@ public function classify( $post_id ) {
...
+	 *
+	 * @param int $post_id Post ID.
+	 * @return bool|int|WP_Error
+	 */
+	public function synthesize_speech( $post_id ) {
+		$normalizer   = new Normalizer();
+		$settings     = \Classifai\get_azure_text_to_speech_settings();
+		$post         = get_post( $post_id );
+		$post_content = $normalizer->normalize_content( $post->post_content, $post->post_title, $post_id );
+		$voice        = $settings['voice'];
```
iamdharmesh:
```suggestion
		$voice        = isset( $settings['voice'] ) ? $settings['voice'] : '';
```

includes/Classifai/Services/LanguageProcessing.php
```diff
@@ -167,6 +186,24 @@ public function generate_post_excerpt( WP_REST_Request $request ) {
...
+	 * @param WP_REST_Request $request Full data about the request.
+	 * @return boolean
+	 */
+	public function synthesize_speech_from_text( $request ) {
+		$post_id           = $request->get_param( 'id' );
+		$save_post_handler = new SavePostHandler();
+		$attachment_id     = $save_post_handler->synthesize_speech( $post_id );
+
+		if ( is_wp_error( $attachment_id ) ) {
+			return false;
```
iamdharmesh: Should we return any error message here? to inform users about text-to-speech generation is failed? what do you think?

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,553 @@
...
+						<span class="dashicons dashicons-controls-pause"></span>
+					</div>
+					<div class='classifai-post-audio-heading'>
+						<?php
+							$list_to_post_text = apply_filters( 'classifai_listen_to_this_post_text', esc_html__( 'Listen to this post', 'classifai' ) );
+							echo esc_html( $list_to_post_text );
+						?>
+					</div>
+				</div>
+				<audio id="classifai-post-audio-player" src="<?php echo esc_url( home_url( $audio_attachment_url ) ); ?>"></audio>
```
iamdharmesh:
```suggestion
				<audio id="classifai-post-audio-player" src="<?php echo esc_url( $audio_attachment_url ); ?>"></audio>
```

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,553 @@
...
+		if ( ! empty( $voices_options ) ) {
+			add_settings_field(
+				'voice',
+				esc_html__( 'Voice', 'classifai' ),
+				[ $this, 'render_select' ],
+				$this->get_option_name(),
+				$this->get_option_name(),
+				[
+					'label_for'     => 'voice',
+					'options'       => $voices_options,
```
iamdharmesh, line 195: Currently, we are showing the list of all voices, could we limit it by language? Site in the English language should show only English voices. what are your thoughts on this?

iamdharmesh (changes requested):
Hi @Sidsector9,

Thanks for the great work here 🎉 . Overall, PR looks good to me. I have added a few minor notes to check and discuss otherwise functionality works as expected and as described in the PR description.

Apart from the notes on code, could you please help to add a readme section for "**Set Up Text to Speech Language Processing**" and a screenshot for this newly added feature as we have for other features?

Thanks again for all your efforts on this.

includes/Classifai/Providers/Azure/TextToSpeech.php
Sidsector9: @iamdharmesh as discussed on Slack, we can leave this as it for now as it doesn't have a major impact.

includes/Classifai/Admin/SavePostHandler.php
```diff
@@ -167,6 +170,121 @@ public function classify( $post_id ) {
...
+	 *
+	 * @param int $post_id Post ID.
+	 * @return bool|int|WP_Error
+	 */
+	public function synthesize_speech( $post_id ) {
+		$normalizer   = new Normalizer();
+		$settings     = \Classifai\get_azure_text_to_speech_settings();
+		$post         = get_post( $post_id );
+		$post_content = $normalizer->normalize_content( $post->post_content, $post->post_title, $post_id );
+		$voice        = isset( $settings['voice'] ) ? $settings['voice'] : '';
```
dkotter: Can simplify this line a bit:
```suggestion
		$voice        = $settings['voice']  ?? '';
```

includes/Classifai/Admin/SavePostHandler.php
```diff
@@ -167,6 +170,121 @@ public function classify( $post_id ) {
...
+		// Return error if creation of attachment fails.
+		if ( is_wp_error( $attachment_id ) ) {
+			return new \WP_Error(
+				'azure_text_to_speech_resource_creation_failure',
+				esc_html__( 'Audio creation failed.', 'classifai' )
+			);
+		} else {
+			update_post_meta( $post_id, TextToSpeech::AUDIO_ID_KEY, $attachment_id );
+			update_post_meta( $post_id, TextToSpeech::AUDIO_TIMESTAMP_KEY, time() );
+		}
```
dkotter: No real need for this `else` statement, especially since we `return` in the statement above.
```suggestion

		update_post_meta( $post_id, TextToSpeech::AUDIO_ID_KEY, $attachment_id );
		update_post_meta( $post_id, TextToSpeech::AUDIO_TIMESTAMP_KEY, time() );
```

includes/Classifai/Admin/SavePostHandler.php
```diff
@@ -167,6 +170,121 @@ public function classify( $post_id ) {
...
+
+		// Upload the audio stream as an .mp3 file.
+		$file_data = wp_upload_bits(
+			$audio_file_name,
+			null,
+			$response_body
+		);
+
+		// Insert the audio file as attachment.
+		$attachment_id = wp_insert_attachment(
```
dkotter, line 287: Should we check that `$file_data` didn't return an error before proceeding here? Looking at the return of `wp_upload_bits`, it will always return an `array` but there could be an error message in that `array`. Thinking we should check if that exists before proceeding here

includes/Classifai/Admin/SavePostHandler.php
```diff
@@ -167,6 +170,121 @@ public function classify( $post_id ) {
...
+			array(
+				'guid'           => $file_data['file'],
+				'post_title'     => $audio_file_name,
+				'post_mime_type' => $file_data['type'],
+			),
+			$file_data['file']
+		);
+
+		// Return error if creation of attachment fails.
+		if ( is_wp_error( $attachment_id ) ) {
```
dkotter: Looking at the function definition for `wp_insert_attachment`, it doesn't return a `WP_Error` object unless you pass in `true` for the 4th argument, which we're not doing here. So if something were to go wrong, it will return the value zero instead, so I think this check needs to be:
```suggestion
		if ( ! $attachment_id ) {
```

includes/Classifai/Admin/SavePostHandler.php
```diff
@@ -167,6 +170,121 @@ public function classify( $post_id ) {
...
+			return new \WP_Error(
+				'azure_text_to_speech_resource_creation_failure',
+				esc_html__( 'Audio creation failed.', 'classifai' )
+			);
+		} else {
+			update_post_meta( $post_id, TextToSpeech::AUDIO_ID_KEY, $attachment_id );
```
dkotter: I know we're just saving the attachment ID that was generated above but probably still worth having escaping around this value

includes/Classifai/Admin/SavePostHandler.php
```diff
@@ -167,6 +170,121 @@ public function classify( $post_id ) {
...
 	}
 
+	/**
+	 * Synthesizes speech from the post title and content.
+	 *
+	 * @param int $post_id Post ID.
+	 * @return bool|int|WP_Error
+	 */
+	public function synthesize_speech( $post_id ) {
+		$normalizer   = new Normalizer();
```
dkotter, line 194: So this may be overkill but as I've added a few new features to Classifai over the last few months, I try to ensure we are always checking if a feature is turned on and the current user has permissions before proceeding (as we don't currently do this everywhere, which I have a WIP branch to fix that).

Right now, this method is only called when our custom REST endpoint is hit, which has a permissions callback that does properly check permissions. But what if someone were to call this method directly? Even though it means we end up double checking permissions in the majority of use cases, I'd suggest we add similar checks here as well, both a `current_user_can` check and ensuring the post type is supported.

includes/Classifai/Admin/SavePostHandler.php
```diff
@@ -167,6 +170,121 @@ public function classify( $post_id ) {
...
+			'body'    => $request_body,
+			'headers' => array(
+				'Ocp-Apim-Subscription-Key' => $settings['api_key'],
+				'Content-Type'              => 'application/ssml+xml',
+				'X-Microsoft-OutputFormat'  => 'audio-16khz-128kbitrate-mono-mp3',
+			),
+		);
+
+		$remote_url = sprintf( '%s%s', $settings['url'], TextToSpeech::API_PATH );
+		$response   = wp_remote_post( $remote_url, $request_params );
```
dkotter, line 237: If someone doesn't have proper credentials in place, this request will fail. What are your thoughts on adding a check here to ensure we have authorized credentials before making this API request, allowing us to show a more proper error message to the user?

includes/Classifai/Helpers.php
```diff
@@ -596,7 +596,39 @@ function attachment_is_pdf( $post ) {
...
 
 /**
- * Get asset info from extracted asset files.
+ * Returns the settings for Azure's Text to Speech Provider.
+ * Found at `/wp-admin/admin.php?page=language_processing&tab=azure_text_to_speech`
+ *
+ * @todo Move this to a more generic method during refactoring of the plugin.
+ * @return array
+ */
+function get_azure_text_to_speech_settings() {
```
dkotter: Instead of having this custom helper function specific to Azure Text to Speech, could you use the `get_plugin_settings` helper we already have, since it now allows you to pass in both the service and provider?
```php
$settings = get_plugin_settings( 'language_processing', 'Text to Speech' )
```

includes/Classifai/Helpers.php
```diff
@@ -596,7 +596,39 @@ function attachment_is_pdf( $post ) {
...
+	return $provider->get_settings();
+}
+
+/**
+ * Returns supported post types for Azure Text to Speech.
+ *
+ * @todo Move this to a more generic method during refactoring of the plugin.
+ * @return array
+ */
+function get_supported_post_types_for_azure_text_to_speech() {
```
dkotter: Looks like this is only used in our REST permission callback. Instead of adding a new helper method, can we just add this code to that permissions callback? Or maybe move this function into the `TextToSpeech` class?

includes/Classifai/Services/LanguageProcessing.php
```diff
@@ -75,6 +76,24 @@ public function register_endpoints() {
 				'permission_callback' => [ $this, 'generate_post_excerpt_permissions_check' ],
 			]
 		);
+
+		register_rest_route(
+			'classifai/v1',
+			'synthesize-speech/(?P<id>\d+)',
+			[
+				'methods'             => \WP_REST_Server::READABLE,
```
dkotter:
```suggestion
				'methods'             => WP_REST_Server::READABLE,
```

includes/Classifai/Services/LanguageProcessing.php
```diff
@@ -75,6 +76,24 @@ public function register_endpoints() {
...
+				'callback'            => array( $this, 'synthesize_speech_from_text' ),
+				'args'                => array(
+					'id' => array(
+						'required'          => true,
+						'type'              => 'integer',
+						'sanitize_callback' => 'absint',
+						'description'       => esc_html__( 'Post ID to generate tags.', 'classifai' ),
```
dkotter:
```suggestion
						'description'       => esc_html__( 'ID of post to run text to speech conversion on.', 'classifai' ),
```

includes/Classifai/Services/LanguageProcessing.php
```diff
@@ -167,6 +186,34 @@ public function generate_post_excerpt( WP_REST_Request $request ) {
 		return rest_ensure_response( $provider->rest_endpoint_callback( $post_id, 'excerpt' ) );
 	}
 
+	/**
+	 * Generates text to speech for a post using REST.
+	 *
+	 * @param WP_REST_Request $request Full data about the request.
+	 * @return boolean
+	 */
+	public function synthesize_speech_from_text( $request ) {
```
dkotter:
```suggestion
	public function synthesize_speech_from_text( WP_REST_Request $request ) {
```

includes/Classifai/Services/LanguageProcessing.php
```diff
@@ -167,6 +186,34 @@ public function generate_post_excerpt( WP_REST_Request $request ) {
...
+			$error_message = $attachment_id->get_error_message();
+
+			return array(
+				'success' => false,
+				'code'    => $error_code,
+				'message' => $error_message,
+			);
+		}
+
+		return array(
```
dkotter: I would suggest wrapping the two return statements here in `rest_ensure_response`

includes/Classifai/Services/LanguageProcessing.php
```diff
@@ -167,6 +186,34 @@ public function generate_post_excerpt( WP_REST_Request $request ) {
...
+		$attachment_id     = $save_post_handler->synthesize_speech( $post_id );
+
+		if ( is_wp_error( $attachment_id ) ) {
+			$error_code    = $attachment_id->get_error_code();
+			$error_message = $attachment_id->get_error_message();
+
+			return array(
+				'success' => false,
+				'code'    => $error_code,
+				'message' => $error_message,
```
dkotter: Can simplify this by removing the lines above:
```suggestion
				'code'    => $attachment_id->get_error_code(),
				'message' => $attachment_id->get_error_message(),
```

includes/Classifai/Services/LanguageProcessing.php
```diff
@@ -208,4 +255,27 @@ public function generate_post_excerpt_permissions_check( WP_REST_Request $reques
...
+	public function speech_synthesis_permissions_check( $request ) {
+		$post_id = $request->get_param( 'id' );
+
+		if ( ! empty( $post_id ) && current_user_can( 'edit_post', $post_id ) ) {
+			$post_type = get_post_type( $post_id );
+			$supported = \Classifai\get_supported_post_types_for_azure_text_to_speech();
+
+			// Check if processing allowed.
+			if ( ! in_array( $post_type, $supported, true ) ) {
+				return new \WP_Error( 'not_enabled', esc_html__( 'Azure Speech synthesis is not enabled for current post.', 'classifai' ) );
```
dkotter:
```suggestion
				return new WP_Error( 'not_enabled', esc_html__( 'Azure Speech synthesis is not enabled for current post.', 'classifai' ) );
```

includes/Classifai/Services/LanguageProcessing.php
```diff
@@ -208,4 +255,27 @@ public function generate_post_excerpt_permissions_check( WP_REST_Request $reques
 		return true;
 	}
 
+	/**
+	 * Check if a given request has access to generate audio for the post.
+	 *
+	 * @param WP_REST_Request $request Full data about the request.
+	 * @return WP_Error|bool
+	 */
+	public function speech_synthesis_permissions_check( $request ) {
```
dkotter:
```suggestion
	public function speech_synthesis_permissions_check( WP_REST_Request $request ) {
```

includes/Classifai/Services/LanguageProcessing.php
```diff
@@ -167,6 +186,34 @@ public function generate_post_excerpt( WP_REST_Request $request ) {
 		return rest_ensure_response( $provider->rest_endpoint_callback( $post_id, 'excerpt' ) );
 	}
 
+	/**
+	 * Generates text to speech for a post using REST.
+	 *
+	 * @param WP_REST_Request $request Full data about the request.
+	 * @return boolean
```
dkotter: Right now this always returns an `array`, not a `boolean`. But if we utilize `rest_ensure_response` here, we should change this to:

```suggestion
	 * @return \WP_REST_Response|WP_Error
```

src/js/gutenberg-plugin.js
```diff
@@ -184,6 +187,231 @@ const ClassifAIGenerateTagsButton = () => {
 	);
 };
 
+let errorCode = '';
+
+/**
+ * Calls the server-side method that synthesis speech for a post.
```
```suggestion
 * Calls the server-side method that synthesizes speech for a post.
```

src/js/gutenberg-plugin.js
```diff
@@ -184,6 +187,231 @@ const ClassifAIGenerateTagsButton = () => {
...
+ *
+ * @param {number} postId The Post ID
+ * @return {boolean} Returns true on success.
+ */
+const synthesizeSpeech = async ( postId ) => {
+	// Endpoint URL.
+	const synthesizeSpeechUrl = `${ wpApiSettings.root }classifai/v1/synthesize-speech/${ postId }`;
+
+	// Stores status of the synthensis process.
+	const isProcessing = wp.data.select( postAudioStore ).getIsProcessing();
```
dkotter: Minor thing but instead of calling `wp.data.` in multiple places here (for `select` and `dispatch`), we can destructure those above so we can just call `select` or `dispatch`

src/js/gutenberg-plugin.js
```diff
@@ -184,6 +187,231 @@ const ClassifAIGenerateTagsButton = () => {
...
+	// Indicates whether speech synthesis is enabled for the current post.
+	const isSynthesizeSpeech = 'yes' === useSelect( ( select ) => select( 'core/editor' ).getEditedPostAttribute( 'classifai_synthesize_speech' ) );
+
+	// Post type of the current post.
+	const postType = useSelect( ( select ) => select( 'core/editor' ).getCurrentPostType() );
+
+	// Says if the post is currently being saved.
+	const isSavingPost = useSelect( ( select ) => select( 'core/editor' ).isSavingPost() );
+
+	// Says whether if speech synthesis is in progress.
```
```suggestion
	// Says whether speech synthesis is in progress.
```

src/js/gutenberg-plugin.js
```diff
@@ -184,6 +187,231 @@ const ClassifAIGenerateTagsButton = () => {
...
+		wp.data.dispatch( 'core/editor' ).savePost();
+	}
+
+	// Fetches the latest audio file to avoid disk cache.
+	const cacheBustingUrl = `${ sourceUrl }?ver=${ timestamp }`;
+
+	return (
+		<>
+			<ToggleControl
+				label={ __( 'Generate audio for this post.' ) }
```
```suggestion
				label={ __( 'Generate audio for this post.', 'classifai' ) }
```

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,553 @@
+<?php
+/**
+ * IBM Watson NLU
```
dkotter: This isn't accurate

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,553 @@
...
+ */
+
+namespace Classifai\Providers\Azure;
+
+use Classifai\Providers\Provider;
+use Classifai\Blocks;
```
dkotter: Doesn't seem this is used anywhere?

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,553 @@
...
+use WP_Http;
+
+use function Classifai\get_post_types_for_language_settings;
+
+class TextToSpeech extends Provider {
+
+	const API_PATH = 'cognitiveservices/v1';
```
dkotter, line 29: Might be nice to have descriptions on what each of these constants are for

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,553 @@
...
+			$service
+		);
+	}
+
+	/**
+	 * Enqueue the editor scripts.
+	 */
+	public function enqueue_editor_assets() {
+		global $post;
+		wp_enqueue_script(
```
dkotter, line 83: For these enqueues, I'd suggest using the new `get_asset_info` helper function if possible (this is utilized in the ChatGPT class as an example)

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,553 @@
...
+
+	/**
+	 * Sanitization callback for settings.
+	 *
+	 * @param array $settings The settings being saved.
+	 * @return array
+	 */
+	public function sanitize_settings( $settings ) {
+		$new_settings = $this->get_settings();
+		$new_settings = array();
```
dkotter: Not sure I follow why we call `get_settings` but then immediately set the value to an empty `array`?

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,553 @@
...
+			add_settings_error(
+				$this->get_option_name(),
+				'classifai-azure-text-to-speech-auth-empty',
+				esc_html__( 'One or more credentials required to connect to the Azure Text to Speech service is empty.', 'classifai' ),
+				'error'
+			);
+		}
+
+		// Sanitize the post type checkboxes
+		$post_types = get_post_types( [ 'public' => true ], 'objects' );
```
dkotter: Couple things here. Looks like we're using `get_post_types_for_language_settings` to render out the checkbox options, seems like we should then utilize that here as well for consistency.

Also, I wasn't aware of this until Peter brought it up recently but instead of checking for `public => true`, he suggests using the `is_post_type_viewable` function (see https://github.com/10up/mention-notifier/pull/6#discussion_r1142811966). If we standardize to using the `get_post_types_for_language_settings` here, we should make that change there

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,553 @@
...
+		);
+	}
+
+	/**
+	 * Can the functionality be initialized?
+	 *
+	 * @return bool
+	 */
+	public function can_register() {
+		$options = get_option( $this->get_option_name() );
```
dkotter: For consistency with other providers, can we change this:
```suggestion
		$settings = $this->get_settings();
```

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,553 @@
...
+			'azure_text_to_speech',
+			$service
+		);
+	}
+
+	/**
+	 * Enqueue the editor scripts.
+	 */
+	public function enqueue_editor_assets() {
+		global $post;
```
dkotter, line 81: We should probably add a check here to ensure this feature has been enabled for this post type before loading the JS files, so we don't end up having files loaded unnecessarily.

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,553 @@
...
+		return $new_settings;
+	}
+
+	/**
+	 * Connects to Azure's Text to Speech service.
+	 *
+	 * @param array $args Overridable args.
+	 */
+	public function connect_to_service( $args = array() ) {
+		$options = get_option( $this->get_option_name(), array() );
```
dkotter: Can this be changed to `$this->get_settings()`?

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,553 @@
...
+
+		return $new_settings;
+	}
+
+	/**
+	 * Connects to Azure's Text to Speech service.
+	 *
+	 * @param array $args Overridable args.
+	 */
+	public function connect_to_service( $args = array() ) {
```
```suggestion
	public function connect_to_service( array $args = array() ) {
```

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,553 @@
...
+			$new_settings['voice'] = '';
+		}
+
+		return $new_settings;
+	}
+
+	/**
+	 * Connects to Azure's Text to Speech service.
+	 *
+	 * @param array $args Overridable args.
```
dkotter, line 269: Should add a return statement here

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,553 @@
...
+				'error'
+			);
+
+			return false;
+		}
+
+		$response_body = wp_remote_retrieve_body( $response );
+		$voices        = json_decode( $response_body );
+
+		set_transient( 'classifai-azure-speech-to-text-voices', $voices, MONTH_IN_SECONDS );
```
dkotter: Do we need any sanitization around `$voices`?

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,553 @@
...
+		}
+
+		foreach ( $voices_transient as $voice ) {
+			// phpcs is disabled because it throws error for camel case.
+			// phpcs:disable
+			$options[ "{$voice->ShortName}|{$voice->Gender}" ] = sprintf(
+				'%1$s (%2$s/%3$s)',
+				$voice->LocaleName,
+				$voice->DisplayName,
+				$voice->Gender
```
dkotter: I think escaping around these would be great

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,553 @@
...
+	public function get_provider_debug_information( $settings = null, $configured = null ) {
+		if ( is_null( $settings ) ) {
+			$settings = $this->sanitize_settings( $this->get_settings() );
+		}
+
+		$authenticated = 1 === intval( $settings['authenticated'] ?? 0 );
+
+		return [
+			__( 'Authenticated', 'classifai' )  => $authenticated ? __( 'Yes', 'classifai' ) : __( 'No', 'classifai' ),
+			__( 'API URL', 'classifai' )        => $settings['url'] ?? '',
```
```suggestion
			__( 'Authenticated', 'classifai' ) => $authenticated ? __( 'Yes', 'classifai' ) : __( 'No', 'classifai' ),
			__( 'API URL', 'classifai' )       => $settings['url'] ?? '',
```

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,553 @@
...
+			'authenticated' => false,
+		];
+	}
+
+	/**
+	 * Add `classifai_synthesize_speech` to rest API for view/edit.
+	 */
+	public function add_synthesize_speech_meta_to_rest_api() {
+		$settings             = \Classifai\get_azure_text_to_speech_settings();
+		$supported_post_types = isset( $settings['post_types'] ) ? $settings['post_types'] : array();
```
```suggestion
		$supported_post_types = $settings['post_types'] ?? array();
```

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,553 @@
...
+			'voice'         => '',
+			'authenticated' => false,
+		];
+	}
+
+	/**
+	 * Add `classifai_synthesize_speech` to rest API for view/edit.
+	 */
+	public function add_synthesize_speech_meta_to_rest_api() {
+		$settings             = \Classifai\get_azure_text_to_speech_settings();
```
dkotter: Can we use `$this->get_settings()` here instead?

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,553 @@
...
+	 * @return string
+	 */
+	public function render_post_audio_controls( $content ) {
+		global $post;
+
+		if ( ! $post ) {
+			return $content;
+		}
+
+		if ( apply_filters( 'classifai_disable_post_to_audio_block', false, $post->ID ) ) {
```
dkotter, line 510: Can we add a filter docblock here?

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,553 @@
...
+
+		wp_enqueue_style(
+			'classifai-post-audio-player-css',
+			CLASSIFAI_PLUGIN_URL . '/dist/post-audio-controls.css',
+			array(),
+			CLASSIFAI_PLUGIN_VERSION,
+			'all'
+		);
+
+		$audio_attachment_id  = (int) get_post_meta( $post->ID, self::AUDIO_ID_KEY, true );
```
```suggestion
		$audio_attachment_id = (int) get_post_meta( $post->ID, self::AUDIO_ID_KEY, true );
```

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,553 @@
...
+	/**
+	 * Adds audio controls to the post that has speech sythesis enabled.
+	 *
+	 * @param string $content Post content.
+	 * @return string
+	 */
+	public function render_post_audio_controls( $content ) {
+		global $post;
+
+		if ( ! $post ) {
```
dkotter, line 496: Do we need an extended check here to ensure this post type is supported, in case someone had this supported at one point but removed support later?

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,553 @@
...
+			return $content;
+		}
+
+		if ( apply_filters( 'classifai_disable_post_to_audio_block', false, $post->ID ) ) {
+			return $content;
+		}
+
+		wp_enqueue_script(
```
dkotter, line 520: Should probably move these enqueues after the check for an audio attachment ID, so we're not loading these assets if we don't actually have audio to play

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,553 @@
...
+		?>
+			<div>
+				<div class='classifai-listen-to-post-wrapper'>
+					<div class="class-post-audio-controls">
+						<span class="dashicons dashicons-controls-play"></span>
+						<span class="dashicons dashicons-controls-pause"></span>
+					</div>
+					<div class='classifai-post-audio-heading'>
+						<?php
+							$list_to_post_text = apply_filters( 'classifai_listen_to_this_post_text', esc_html__( 'Listen to this post', 'classifai' ) );
```
dkotter: Can we add a docblock for this filter?

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,553 @@
...
+
+		echo '</ul>';
+	}
+
+	/**
+	 * Generic text input field callback
+	 *
+	 * @param array $args The args passed to add_settings_field.
+	 */
+	public function render_checkbox_group( array $args = array() ) {
```
dkotter: So we have this method in the parent class, will that not work here?

dkotter (changes requested):

dkotter:
@Sidsector9 Nice work here! Sorry for the amount of review items, most are pretty minor. I've been spending lots of time in Classifai lately so trying to get the codebase to be consistent.

Also, if you could add some E2E tests here, that would be great (or let me know if you'd like some help on that). Let me know if you have any questions.

includes/Classifai/Admin/SavePostHandler.php
Sidsector9: @dkotter if the credentials are incorrect, the audio controls won't render.

src/js/gutenberg-plugin.js
Sidsector9: @dkotter I have destructured them locally within the functions they are used due to clash with `select` function received inside `useSelect()`

includes/Classifai/Services/LanguageProcessing.php
```diff
@@ -75,6 +77,24 @@ public function register_endpoints() {
...
+			'classifai/v1',
+			'synthesize-speech/(?P<id>\d+)',
+			[
+				'methods'             => WP_REST_Server::READABLE,
+				'callback'            => array( $this, 'synthesize_speech_from_text' ),
+				'args'                => array(
+					'id' => array(
+						'required'          => true,
+						'type'              => 'integer',
+						'sanitize_callback' => 'absint',
```
ravinderk, line 94: @Sidsector9 WordPress handles sanitization of `type=number` request param ([Code](https://github.com/WordPress/WordPress/blob/996fe7077e2da39a1c7a795e414c2f8d411a7831/wp-includes/rest-api.php#L2791))

Do you think we can remove `sanitize_callback` param?

dkotter: Two things I'd mention:

1. This is my personal opinion so feel free to ignore but I like being verbose in these things rather than rely on the default (even if we run the same code as the default), just so it's clear at a glance that we are sanitizing things here
2. The default sanitization here is `return (int) $value` which ensures the `$value` is always an integer but will allow negative numbers. Using `absint` ensures we never have a negative number here

ravinderk: @dkotter I agree with you. It is ok to use `absint`.

Sidsector9: @dkotter in this specific case, @ravinderk's suggestion will work fine because the regex `\d+` will not allow negative numbers and it will return a "No route was found" error.

Would you recommend removing the `absint` sanitisation checks?

dkotter: It doesn't matter to me either way. I'd personally leave it just so it's clear and obvious what is going on but I'm not opposed to removing it for simplicities sake

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,673 @@
...
+	 */
+	public function __construct( $service ) {
+		parent::__construct(
+			'Microsoft Azure',
+			self::FEATURE_NAME,
+			'azure_text_to_speech',
+			$service
+		);
+
+		add_action( 'admin_init', array( $this, 'maybe_connect_on_init' ) );
```
dkotter: I'm assuming this is in the constructor as opposed to in the `register` method because if the connection expires, the `register` method never runs because `can_register` will return false, is that correct? Could we at least move this into the `can_register` method so we don't have hooks being initialized in the constructor?

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,673 @@
...
+	 */
+	public function connect_to_service( array $args = array(), $ignore_transient = false ) {
+		$creditials = $this->get_settings( 'credentials' );
+
+		$default = array(
+			'url'     => isset( $creditials['url'] ) ? $creditials['url'] : '',
+			'api_key' => isset( $creditials['api_key'] ) ? $creditials['api_key'] : '',
+		);
+
+		$default = array_merge( $default, $args );
```
dkotter: If desired, could use `wp_parse_args` here to handle merging defaults

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,673 @@
...
+				$request_url,
+				'',
+				3,
+				1,
+				20,
+				$request_params
+			);
+		} else {
+			// phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.wp_remote_get_wp_remote_get -- use of `vip_safe_wp_remote_get` is done when available.
+			$response = wp_remote_get(
```
dkotter, line 313: Do we need to set a higher timeout value here? The default is 5 seconds, not sure if we ever come close to exceeding that.

includes/Classifai/Providers/Azure/TextToSpeech.php
dkotter: Looks like we have a new `$sanitized_voices` param but not seeing that used anywhere

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,673 @@
...
+	public function get_provider_debug_information( $settings = null, $configured = null ) {
+		if ( is_null( $settings ) ) {
+			$settings = $this->sanitize_settings( $this->get_settings() );
+		}
+
+		$authenticated = 1 === intval( $settings['authenticated'] ?? 0 );
+
+		return [
+			__( 'Authenticated', 'classifai' ) => $authenticated ? __( 'Yes', 'classifai' ) : __( 'No', 'classifai' ),
+			__( 'API URL', 'classifai' )       => $settings['url'] ?? '',
```
dkotter: We could add the voice data here if we think that is useful for debugging purposes

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,673 @@
...
+			array_filter(
+				$supported_post_types,
+				function( $post_type ) {
+					return ! is_null( $post_type );
+				}
+			)
+		);
+
+		register_rest_field(
+			$supported_post_types,
```
dkotter, line 453: Do we need to check that we have at least one supported post type before running this code? Not sure if it throws an error if we have an empty array

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,673 @@
...
+
+		echo '</ul>';
+	}
+
+	/**
+	 * Generic text input field callback
+	 *
+	 * @param array $args The args passed to add_settings_field.
+	 */
+	public function render_input( $args ) {
```
dkotter: So we have this method in the parent class and seems to be pretty similar to what we have here. If there's specific changes made here that are needed, I'd suggest trying to make the parent method flexible enough to account for those instead of duplicating this method here

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -0,0 +1,673 @@
...
+	}
+
+	/**
+	 * Render the post types checkbox array.
+	 *
+	 * @param array $args Settings for the input
+	 *
+	 * @return void
+	 */
+	public function render_post_types_checkboxes( $args ) {
```
dkotter: I think you can use the `render_checkbox_group` parent method here instead. See https://github.com/10up/classifai/pull/437/files#diff-2eb150fd29e6aa3e88614679d3738c2afe4e1341e78cc71bf3a36dc5f39e1d77R168 for an example

src/js/gutenberg-plugin.js
```diff
@@ -184,6 +187,217 @@ const ClassifAIGenerateTagsButton = () => {
...
+	const isSynthesizeSpeech = 'yes' === useSelect( ( select ) => select( 'core/editor' ).getEditedPostAttribute( 'classifai_synthesize_speech' ) );
+
+	// Post type of the current post.
+	const postType = useSelect( ( select ) => select( 'core/editor' ).getCurrentPostType() );
+
+	// Says whether speech synthesis is in progress.
+	const isProcessingAudio = useSelect( ( select ) => select( postAudioStore ).getIsProcessing() );
+
+	// Figure out if speech synthesis is supported by the current post.
+	if ( classifaiTextToSpeechData && classifaiTextToSpeechData.supportedPostTypes.includes( postType ) ) {
```
dkotter: I'm getting an error here due to this not being defined: `ReferenceError: classifaiTextToSpeechData is not defined`

This crashes the block editor. This happens when I don't have proper authentication in place and as such, the `register` method doesn't fire and `enqueue_editor_assets` never runs

src/js/gutenberg-plugin.js
```diff
@@ -231,9 +424,19 @@ const ClassifAIPlugin = () => {
...
+					isPosTypeSupported &&
+					isPostStatusSupported && (
+						<>
+							<ClassifAIToggle />
+							<ClassifAIGenerateTagsButton />
+						</>
+					)
+				}
 			</>
+			<ClassifAITSpeechSynthesisToggle audioId={ audioId } />
```
dkotter: So the issue with this new approach is this ClassifAI panel will always show, even if all features are turned off. Maybe that's okay? But it is a change to how we approach this right now

src/js/gutenberg-plugin.js
```diff
@@ -184,6 +187,217 @@ const ClassifAIGenerateTagsButton = () => {
...
+	} else if ( isPreviewing ) {
+		audioIcon = 'controls-pause';
+	}
+
+	return (
+		<>
+			<ToggleControl
+				label={ __( 'Generate audio for this post.', 'classifai' ) }
+				help={ isFeatureSupported
+					? __( 'ClassifAI will generate audio for the post when it is published or updated.', 'classifai' )
```
dkotter: Would be great if we didn't hardcode to using `post` here. For instance, if I activate this feature for pages only, it still says `post` in this panel which may be confusing.

includes/Classifai/Providers/Azure/TextToSpeech.php
dkotter: Would also be great to not hardcode the post type here, as this isn't accurate if someone uses this on a post type other than `post`

dkotter (changes requested):

dkotter:
@Sidsector9 Gave this another review as well as a round of testing. Great work here, things work pretty much like I'd expect. I've left a few more notes on things I found while doing testing but this is pretty close.

Would be great to get documentation added here in the readmes (I fumbled around trying to figure out the correct API URL for this) and get some E2E tests in place before we merge this in.

includes/Classifai/Providers/Azure/TextToSpeech.php
Sidsector9: @dkotter it fails due to timeout while testing with large text content, that's why for safety I increased the value.

dkotter: Sorry, I worded this wrong. It doesn't appear that we are changing the default timeout value here, so it will use 5 seconds. In other places, we've changed this to be at least 30 seconds, wondering if we need to do the same here?

src/js/gutenberg-plugin.js
Sidsector9: @dkotter Yes, I have intentionally kept it this way.
Since we have multiple services under the same panel, the conditional check just to show/hide the panel will be way too lengthy and complicated. This is why I am always showing the Speech synthesis controls but enabled/disabled based on supported post type.

src/js/gutenberg-plugin.js
Sidsector9: Okay I was able to reproduce this after some time. This happens when NLU is also enabled which loads this file which causes it to break.

jeffpaul:
@Sidsector9 it would be good to test this with the Twenty Twenty-Three and Twenty Twenty-Two default themes to ensure things render as expected on the frontend (and if not, add compatibility to ensure at least those two themes work as expected).

.github/workflows/lint.yml
```diff
@@ -36,14 +36,10 @@ jobs:
...
+          **/*.json
 
     - name: Run JS linting
-      run: |
-        if [[ "${{ steps.changed-files.outputs.any_changed }}" == 'true' && "${{ github.event_name }}" == "pull_request" ]]; then
-          npx wp-scripts lint-js ${{ steps.changed-files.outputs.all_changed_files }}
-        elif [[ "${{ github.event_name }}" == "push" ]]; then
-          npm run lint:js
-        fi
+      run: npm run lint:js
```
dkotter: Curious why this change was needed? The goal here was to only run linting on JS files that changed in a PR, whereas now it will run linting on all JS files.

includes/Classifai/Helpers.php
```diff
@@ -5,6 +5,7 @@
 use Classifai\Providers\Provider;
 use Classifai\Services\Service;
 use Classifai\Services\ServicesManager;
+use Classifai\Providers\Azure\TextToSpeech;
```
dkotter: Doesn't seem like this is needed anymore

includes/Classifai/Providers/Azure/TextToSpeech.php
dkotter: @Sidsector9 Is this still an issue?

dkotter (changes requested):

dkotter:
@Sidsector9 Just a couple small things left I flagged. Then we'll need to update the README file with details on this new feature and this should be good to go

.github/workflows/lint.yml
Sidsector9: So, I added `*.json` to the list in case any linting or prettier rules dependency is updated, then the action will lint using the updated rules. Due to this, `npx wp-scripts lint-js ${{ steps.changed-files.outputs.all_changed_files }}` runs on the `*.json` when it is part of the diff due to which the linting fails.

Sidsector9: I handled this in the eslint config but it still leaves a warning. [390a0ba](https://github.com/10up/classifai/pull/403/commits/390a0babc43566c2c6f8d5f1deb17ba791f1927b)

Sidsector9:
> @Sidsector9 Is this still an issue?
@dkotter no it's not, I tested it on a fresh installation.

Sidsector9:
@dkotter I have updated the README by referring to the existing documentation. I couldn't confirm the steps with my own Azure account as I've exhausted the trial. Can you confirm this with the 10up account?

dkotter (approved):

rickalee:
@dkotter @Sidsector9 If we wanted to generate audio versions for all past posts, what would be the best method to achieve that? Don't see a quick method to generate via hook we could use for a custom CLI script.

dkotter:
@rickalee At the moment, we don't have a built-in way to handle this in bulk. We do have existing `WP-CLI` integrations, as well as integrations with WordPres core's bulk action handling, which we should look to implement that here (I've opened #460 to track this).

That said, if you're looking to do this on your own before those changes are made, there's two approaches you could fairly easily take:

1. We did add a custom REST endpoint that is used to generate speech. You can write a script that iterates over multiple items and hits this endpoint for each item: `example.com/wp-json/classifai/v1/synthesize-speech/POSTID`
2. You can copy the code from that REST endpoint instead. Same idea as above, you would iterate over multiple items and run something like:
```php
$save_post_handler = new SavePostHandler();
foreach ( $post_ids as $post_id ) {
	$attachment_id = $save_post_handler->synthesize_speech( $post_id );
}
```

PR #405, opened 2023-03-08
Automatic excerpt generation using OpenAI's ChatGPT API

### LINKED ISSUE #159: Auto-populate missing meta tags and descriptions

dkotter:
## Description of the Change

This PR adds an integration with OpenAI as a new provider in the Language Processing service, specifically integrating with ChatGPT. The specific integration being added here is utilizing ChatGPT to provide a summary of a piece of content and then storing that summary in the excerpt field.

### Setup

Setup with this provider only requires an API key. There's validation done on the settings page, anytime settings are saved, to verify if the API key is valid. This can give you one of three errors:

| No API Key | Invalid API Key | Rate limit reached |
| ----------- | --------------- | ------------------ |
| No API key entered | Invalid API key entered | Rate limit reached |

If the API key is valid, you won't get an error message and you'll get a success message instead. You then have a few settings to change. The most important is turning on the `Generate excerpt` option. If a valid API key is added but this setting isn't on, no integration will happen. The `Allowed roles` setting allows you to control which user roles see the `Generate excerpt` button (and are allowed to access the REST API endpoint). The `Excerpt length` setting controls how many words the final excerpt will be. This defaults to the excerpt length that WordPress has set (that can be changed by the core `excerpt_length` filter). The `Temperature value` is one of the config options the [API supports](https://platform.openai.com/docs/api-reference/chat/create). There are other options there but I decided not to bring those over for now.

ChatGPT settings

### Excerpt integration

I debated a few different approaches on how to actually integrate with the excerpt generation. My priorities were the following:

1. Limit how often we hit the API since OpenAI charges by usage
2. Only try generating an excerpt once the content is (mostly) finalized. No sense in generating an excerpt on an in-progress post that doesn't have much content
3. Have the ability to see what the generated excerpt will be **before** publishing
4. Have the ability to add your own excerpt, modify the generated excerpt or remove the excerpt all together without those changes being overwritten

I initially was thinking of automatically generating an excerpt on save (draft, publish, ...) but this goes against point 1 and 2. I then considered automatically generating only on publish but that goes against 3 (and possibly 4). I eventually landed on the approach of adding a `Generate excerpt` button in the `Excerpt` panel that will send content to OpenAI when clicked and populate the excerpt with whatever value is returned.

This solves all the points above, as you are able to choose how often the API is hit and when the excerpt is generated (only when the button is pressed). And if you don't want to generate the excerpt, you don't have to. It does make the process more manual, as you have to click the button but I think that's a fine trade-off. I am open to other ideas on the best integration here though (I had considered adding something to the pre-publish panel but I don't think that's enough by itself. May be worth adding in addition to what else we have here).

| Excerpt panel | Excerpt loading in |

### API integration

When the `Generate excerpt` button is clicked, a request is made to a new REST endpoint (`wp-json/classifai/v1/generate-excerpt/POST_ID`). This endpoint verifies the current user has permission to edit the post, we are properly authenticated with OpenAI and the `Generate excerpt` setting is on.

Assuming all that passes, a new `Tokenizer` class has been added that will try and determine how many tokens the content has and how many tokens the final excerpt will be. The ChatGPT API has a limit of 4096 tokens per request and this includes both the data you send and the data that is sent back. Unfortunately tokens are equivalent to words or characters (roughly 4 characters is 1 token) but we do some basic calculations, erring on the side of being too aggressive, to ensure our request doesn't go over the limit.

A new `APIRequest` class has been added here as well (followed the approach in the Watson `APIRequest` class) to make it easier to integrate with the API, not only for this feature but for any other OpenAI features we may add in the future (the `Tokenizer` class should also be reusable for future integrations).

The request is sent and then the response is parsed and returned, whether we get a successful response including our excerpt or we get an error. If it's an error, that will be shown to the user. If success, we set the returned value as our excerpt.

### Reviewer notes

- I couldn't find a way to add a custom button to the core Excerpt panel so I removed that panel all together and replaced it with our own, copying most of the code from Gutenberg and adding in our custom handling
- There's no WP-CLI integration in this PR. I think it makes sense to add that in a followup PR
- I added new details about OpenAI into the readme files and also updated those in a few places to make it more clear that we have multiple Language Processing providers now
-  I added a new image to the readme as well and noticed the existing images were not optimized, so I ran those through an optimization step as well as reduced the dimensions on one image. I can revert that last change if we want that image to stay super large but currently [displays weird in the readme](https://github.com/10up/classifai#features)
- `get_plugin_settings` was updated to account for multiple providers instead of just always using the first provider. There are other places in the code that should be updated to account for this but I'm planning to tackle that in a followup PR

## How to test the Change

A valid OpenAI API key is needed to fully test this feature. OpenAI does offer a free $5 credit for new users so if you haven't signed up before, you can [sign up](https://platform.openai.com/signup) and get an API key (ping me in Slack if you want to use my API key for testing).

1. Log in to your OpenAI account and go to your [API key section](https://platform.openai.com/account/api-keys). Generate a new API key there and copy it
2. Go to `ClassifAI > Language Processing > OpenAI` and paste in your API key
3. Turn on the `Generate excerpt` setting. The other settings can be left default. Save changes and ensure no error message is shown
4. Create a new post, ensuring it has at least a few paragraphs of content
5. Open the `Excerpt` panel, ensure you can see the `Generate excerpt` button then click on that
6. Ensure an excerpt gets populated and no errors are shown
7. Can run through these same tests with no API key entered, an invalid API key entered and/or the `Generate excerpt` option is off, ensuring proper error messages are shown and functionality is removed

### Changelog Entry

> Added - Automatic excerpt generation using OpenAI's ChatGPT API

### Credits

Props @dkotter, @jeffpaul, @zamanq 

### Checklist:


dkotter:
Note that E2E tests are failing here but they seem to have been failing for a bit (all recent PRs are failing as well). I'm going to look to see what needs fixed on those and tackle that separately from this PR

src/js/post-excerpt/panel.js
```diff
@@ -0,0 +1,77 @@
...
+				className="editor-post-excerpt__textarea"
+				onChange={ ( value ) => onUpdateExcerpt( value ) }
+				value={ excerpt }
+			/>
+			<ExternalLink
+				href={ __(
+					'https://wordpress.org/support/article/settings-sidebar/#excerpt'
+				) }
+			>
+				{ __( 'Learn more about manual excerpts' ) }
```
ravinderk: @dkotter can we add a text domain to the function? and other places too

src/js/post-excerpt/panel.js
```diff
@@ -0,0 +1,77 @@
...
+			<TextareaControl
+				__nextHasNoMarginBottom
+				label={ __( 'Write an excerpt (optional)' ) }
+				className="editor-post-excerpt__textarea"
+				onChange={ ( value ) => onUpdateExcerpt( value ) }
+				value={ excerpt }
+			/>
+			<ExternalLink
+				href={ __(
+					'https://wordpress.org/support/article/settings-sidebar/#excerpt'
```
ravinderk: @dkotter Does this link need to be translatable?

src/js/post-excerpt/panel.js
```diff
@@ -0,0 +1,77 @@
+/**
+ * WordPress dependencies
+ */
+const { __ } = wp.i18n;
```
ravinderk, line 4: @dkotter, this is good, but I always prefer to use `import` packages when possible because this gives better code completion in the editor.

src/js/post-excerpt/panel.js
dkotter: This code as copied over from Gutenberg (see https://github.com/WordPress/gutenberg/blob/4b4c4befb34d815634b85cbee23cad169ab0e073/packages/editor/src/components/post-excerpt/index.js) and as such, I've left off the text domain on these strings. This means we don't have to maintain translations for these strings as they will load from WordPress Core instead. See:

Screen Shot 2023-03-09 at 8 17 04 AM

src/js/post-excerpt/panel.js
dkotter: As mentioned in my comment above, this code was copied from Gutenberg. Looking at WordPress Core translations, it looks like they do have language specific URLs they swap in here, see https://translate.wordpress.org/projects/wp/6.1.x/fr/default/?filters%5Bterm%5D=https%3A%2F%2Fwordpress.org%2Fsupport%2Farticle%2Fsettings-sidebar%2F%23excerpt&filters%5Bterm_scope%5D=scope_originals&filters%5Bstatus%5D=current_or_waiting_or_fuzzy_or_untranslated&filters%5Buser_login%5D=&filter=Apply+Filters&sort%5Bby%5D=priority&sort%5Bhow%5D=desc

includes/Classifai/Providers/OpenAI/ChatGPT.php
```diff
@@ -0,0 +1,510 @@
...
+	 * Authenticate our credentials.
+	 *
+	 * @param string $api_key Api Key.
+	 *
+	 * @return bool|WP_Error
+	 */
+	protected function authenticate_credentials( string $api_key = '' ) {
+		// Check that we have credentials before hitting the API.
+		if ( empty( $api_key ) ) {
+			return new WP_Error( 'auth', esc_html__( 'Please enter your API key', 'classifai' ) );
```
jeffpaul: Could adjust to "Please enter your OpenAI ChatGPT API key." by including the service provider & feature as well as including the trailing period

includes/Classifai/Providers/OpenAI/ChatGPT.php
```diff
@@ -0,0 +1,510 @@
...
+			'enable-excerpt',
+			esc_html__( 'Generate excerpt', 'classifai' ),
+			[ $this, 'render_input' ],
+			$this->get_option_name(),
+			$this->get_option_name(),
+			[
+				'label_for'     => 'enable_excerpt',
+				'input_type'    => 'checkbox',
+				'default_value' => $default_settings['enable_excerpt'],
+				'description'   => __( 'A button will be added to the excerpt panel that can be used to generate an excerpt.', 'classifai' ),
```
jeffpaul, line 157: Even more automated would be to generate that excerpt on publish (and perhaps on update) versus requiring someone clicks a button.  On one hand we automate things completely while on the other you limit the number of calls to OpenAI.  I'm curious for thoughts from folks on best initial approach to take here?

includes/Classifai/Providers/OpenAI/ChatGPT.php
```diff
@@ -0,0 +1,510 @@
...
+			$this->get_option_name(),
+			$this->get_option_name(),
+			[
+				'label_for'     => 'length',
+				'input_type'    => 'number',
+				'max'           => 5,
+				'min'           => 1,
+				'step'          => 1,
+				'default_value' => $default_settings['length'],
+				'description'   => __( 'How many sentences should the excerpt be?', 'classifai' ),
```
jeffpaul: Is there any opportunity to limit this to characters instead of sentences with OpenAI or is their prompt the number of sentences only?  I ask as some folks' front end designs might only support up to a certain number of characters for rendering an excerpt and am curious if we would be able to better support that sort of precision versus a random total character count based on number of sentences?

includes/Classifai/Providers/OpenAI/ChatGPT.php
```diff
@@ -0,0 +1,510 @@
...
+			]
+		);
+
+		add_settings_field(
+			'temperature',
+			esc_html__( 'Temperature value', 'classifai' ),
```
jeffpaul: I recognize that "Temperature value" and the "sampling temperature" and following helper text closely match what OpenAI is expecting, but I wonder if we might make this a bit easier to comprehend by making this less technical to something like "Excerpt precision" and "On a scale from 1 to 10 how precise do you want the excerpt to be?  10 is very focused and precise while 1 is less precise and random?" and then we convert that 1-10 value into the 0-2 increments?

includes/Classifai/Providers/OpenAI/ChatGPT.php
```diff
@@ -0,0 +1,510 @@
...
+		if ( ! $post_id || ! get_post( $post_id ) ) {
+			return new WP_Error( 'post_id_required', esc_html__( 'Post ID is required to generate an excerpt.', 'classifai' ) );
+		}
+
+		$settings = $this->get_settings();
+
+		// These checks (and the one above) happen in the REST permission_callback,
+		// but we run them again here in case this method is called directly.
+		if ( empty( $settings ) || ( isset( $settings['authenticated'] ) && false === $settings['authenticated'] ) || ( isset( $settings['enable_excerpt'] ) && 'no' === $settings['enable_excerpt'] ) ) {
+			return new WP_Error( 'not_enabled', esc_html__( 'Excerpt generation is turned off or authentication failed.', 'classifai' ) );
```
jeffpaul: Perhaps "Excerpt generation is disable or OpenAI ChatGPT authentication failed." though I guess if we eventually integrate another service for excerpt generation that we wouldn't want to call out OpenAI if something like Azure is actually in use.

includes/Classifai/Providers/OpenAI/ChatGPT.php
jeffpaul: After reading your thoughts in the PR description @dkotter I'm thinking an alternate might be generating the excerpt automatically on publish (or perhaps doing that heading into the pre-publish panel and displaying the info there and allowing someone to change the excerpt or re-generate a new excerpt from OpenAI.  Once the post is published, having the "Generate Excerpt" or "Re-generate Excerpt" (if an excerpt already exists) could allow for someone to tweak the excerpt if they're not liking what's present.

jeffpaul:
@fabiankaegy per Darin's comment of:
> I couldn't find a way to add a custom button the core Excerpt panel so I removed that panel all together and replaced it with our own, copying most of the code from Gutenberg and adding in our custom handling

...are you aware of a way to add a button into that panel or is the approach here the best given the current state of the editor?

jeffpaul:
> There's no WP-CLI integration in this PR. I think it makes sense to add that in a followup PR

I agree, the same holds for being able to have a Bulk Edit menu option for "Generate excerpt(s)" is something that's fine to handle in a followup PR.  @vikrampm1, mind opening an issue to document that please?

jeffpaul (commented):
1.) I couldn't tell in a quick pass here, but are we showing the new Excerpt sidebar and Generate Excerpt button ONLY IF that setting is enabled or are we doing that for any case?  I would recommend we only render that IF the generate excerpt setting is enabled.
2.) In the second error message screenshot, let's ensure we link out on that URL instead of showing it as plaintext.
3.) In the third error message screenshot, let's include a link to the OpenAI site so something like "You have exceeded your current OpenAI quote, please check your plan and billing details: <HTTP link here>."

I've left some other comments/questions throughout, would be good to chat more thoroughly on those but overall this is VERY exciting and a good first integration with OpenAI to help automate editorial workflows!

includes/Classifai/Providers/OpenAI/ChatGPT.php
dkotter: Updated in [9542a02](https://github.com/10up/classifai/pull/405/commits/9542a02ebf8fd899babb02794e209748dc76903e)

includes/Classifai/Providers/OpenAI/ChatGPT.php
dkotter: Updated in [9542a02](https://github.com/10up/classifai/pull/405/commits/9542a02ebf8fd899babb02794e209748dc76903e)

includes/Classifai/Providers/OpenAI/ChatGPT.php
dkotter: Probably worth talking this through to align on best workflow for this. I will say I don't think we want to have it automatically generated on publish, as that doesn't give you a chance to review the excerpt before that goes live. I like the idea of utilizing the pre-publish panel but I think we would still need to have a button that is clicked to send off the API request, so workflow won't be much different

includes/Classifai/Providers/OpenAI/ChatGPT.php
dkotter: We have full control over the prompt we send. Right now that looks like:
`Summarize the following text into 2 short sentences: {post_content}`

There isn't an option on the API side to limit by characters, words or sentences. There is a `max_tokens` param we can use to limit how many tokens the response contains but there's not a super clean way to convert words or characters to tokens (it's roughly 4 characters per token).

I'm open to the idea of changing this, as the responses I usually get from the API are fairly long. I guess I'm concerned that most people won't have a clue on how many characters they want. Maybe we change this from sentences to words? WordPress uses 55 words per excerpt as a default with a filter around that value. Thinking we can pull that same filtered value to use as the default here.

includes/Classifai/Providers/OpenAI/ChatGPT.php
dkotter: I debated even adding this setting in the first place. It's one of the options the [API supports](https://platform.openai.com/docs/api-reference/chat/create) but it's fairly confusing on what it does, not sure it's worth having. I guess I lean towards just removing this rather than trying to change this to make more sense and have to handle some sort of conversion behind the scenes

dkotter:
@jeffpaul 

> 1.) I couldn't tell in a quick pass here, but are we showing the new Excerpt sidebar and Generate Excerpt button ONLY IF that setting is enabled or are we doing that for any case? I would recommend we only render that IF the generate excerpt setting is enabled.

We only load our custom Excerpt panel (which includes the button) if proper authentication is in place AND the excerpt option is turned on

> 2.) In the second error message screenshot, let's ensure we link out on that URL instead of showing it as plaintext.

So this error comes from OpenAI and we just render it as-is. I'll see if there's an easy way to parse that URL out to add an anchor tag around it.

> 3.) In the third error message screenshot, let's include a link to the OpenAI site so something like "You have exceeded your current OpenAI quote, please check your plan and billing details: ."

This is another error message OpenAI sends but we can tack on a URL to the end of this

dkotter:
Both those error messages now have proper anchor links in them (and screenshots have been updated)

includes/Classifai/Providers/OpenAI/ChatGPT.php
dkotter: Actually, after testing, the `max_tokens` param limits how many tokens the response returns but doesn't mean it returns a full response within that limit. So if this limit is set low, the return you get will just cut off in the middle of a sentence, which is not great. I'll experiment a bit more with crafting a better prompt to see if that can be used to more easily control excerpt length

dkotter: All right, I've changed this from setting the excerpt length in sentences to words, defaulting to the excerpt length value in WordPress. Note though that in testing, the results returned from the API tend to exceed this number. It was fairly consistently 40-50 words more than requested and after experimenting with the prompt we send, it's now more consistently around 10-15 extra words.

I've updated the description on the setting to note this but I guess we need to decide if being able to fine-tune a little better by using words is worth the frustration when the results returned will more than likely exceed that number. When requesting sentences, it always gave me the amount of sentences I requested but the results were usually longer than I would want for an excerpt. Now I get results closer to what I would expect for an excerpt but still longer than what I requested.

src/js/post-excerpt/panel.js
dkotter: While I tend to agree, I set this up to match what we're doing in other JS files in this project. I like to try and keep things consistent but open to feedback if you think we should change this here.

fabiankaegy:
> @fabiankaegy per Darin's comment of:
> 
> > I couldn't find a way to add a custom button the core Excerpt panel so I removed that panel all together and replaced it with our own, copying most of the code from Gutenberg and adding in our custom handling
> 
> ...are you aware of a way to add a button into that panel or is the approach here the best given the current state of the editor?

Sadly I'm not aware of any great options here. The only other idea is manually injecting a new react root into the dom at that position using the `createRoot` (`render`) function. But that is fragile if core ever changes the markup of the sidebar since you are doing JS DOM manipulations here.

includes/Classifai/Providers/OpenAI/ChatGPT.php
zamanq: Hi @jeffpaul @dkotter, In response to this possibilities, I've created a PR: #408 off of this one that would add the excerpt in the Pre Publish Panel to support regeneration before publishing. I also added some minor improvements.

includes/Classifai/Providers/OpenAI/ChatGPT.php
jeffpaul: I'm not strongly opinionated in either direction here, so if there's an approach that "feels right" to you then go for it.

includes/Classifai/Providers/OpenAI/ChatGPT.php
jeffpaul: Simpler by removal would be even better, maaaybe leave as a filter or just fully remove and we can look to add a filter later if there's a strong community desire?

jeffpaul:
@iamdharmesh tagging you for code review here as this week you've got some OSS time, hoping to get this ready for release as expeditiously as we can (hoping to get 1-2 features released before Summit as feasible)

includes/Classifai/Providers/OpenAI/ChatGPT.php
dkotter: I've gone ahead and removed that. And there is already a filter around the request body we send, so this and any other API settings can easily be added/changed using that

includes/Classifai/Providers/OpenAI/ChatGPT.php
dkotter: This feels better to me, we can definitely tweak/change this in the future if we want/get feedback

includes/Classifai/Helpers.php
```diff
@@ -583,3 +594,26 @@ function attachment_is_pdf( $post ) {
...
+ *
+ * @param string $slug Asset slug as defined in build/webpack configuration.
+ * @param string $attribute Optional attribute to get. Can be version or dependencies.
+ * @return string|array
+ */
+function get_asset_info( $slug, $attribute = null ) {
+	if ( file_exists( CLASSIFAI_PLUGIN_DIR . '/dist/' . $slug . '.asset.php' ) ) {
+		$asset = require CLASSIFAI_PLUGIN_DIR . '/dist/' . $slug . '.asset.php';
+	} elseif ( file_exists( CLASSIFAI_PLUGIN_DIR . '/dist/' . $slug . '.asset.php' ) ) {
+		$asset = require CLASSIFAI_PLUGIN_DIR . '/dist/' . $slug . '.asset.php';
```
iamdharmesh: May we can remove the 2nd condition here? as it seems duplicate to the 1st condition?

includes/Classifai/Providers/OpenAI/APIRequest.php
```diff
@@ -0,0 +1,106 @@
...
+	public function post( $url = '', $options = [] ) {
+		$this->add_headers( $options );
+		return $this->get_result( wp_remote_post( $url, $options ) ); // phpcs:ignore WordPressVIPMinimum.Functions.RestrictedFunctions.wp_remote_get_wp_remote_get
+	}
+
+	/**
+	 * Get results from the response.
+	 *
+	 * @param object $response The API response.
+	 * @return array|WP_Error
```
iamdharmesh, line 59:
```suggestion
	 * @return array|\WP_Error
```

includes/Classifai/Providers/OpenAI/ChatGPT.php
```diff
@@ -0,0 +1,505 @@
...
+				'temperature' => 0,
+			],
+			$post_id
+		);
+
+		// Make our API request.
+		$response = $request->post(
+			$this->chatgpt_url,
+			[
+				'body' => wp_json_encode( $body ),
```
iamdharmesh, line 442: Should we add the `timeout` option here? I came across the `Error: cURL error 28: Operation timed out after 5000 milliseconds with 0 bytes received` error 2-3 times during Generate excerpt.

iamdharmesh (dismissed):
Thanks for adding this @dkotter. This looks amazing. 🎉 

I just added 2-3 minor notes to discuss but otherwise, all looks great.

includes/Classifai/Helpers.php
dkotter: Ah, good catch. I copied this function over from our [scaffold](https://github.com/10up/wp-scaffold/blob/trunk/themes/10up-theme/includes/utility.php#L23) and the second condition there looks in your `css` directory, which we don't have. So I had removed the `css` directory but as you point out, that makes that line the exact same as the first. I've removed this now

includes/Classifai/Providers/OpenAI/ChatGPT.php
dkotter: Ah, good idea. I have not run into any timeouts myself but I've added the timeout option now to lessen the chance of that happening

PR #407, opened 2023-03-10
Fix failing eslint tests

dkotter:
### Description of the Change

Updates our linting action to run `npm run lint:js`, which in turn runs `wp-scripts lint-js`. This ensures we have more accurate linting done in this action.

Also cleaned up a few things in the eslint action like:
- updating a few actions
- adding caching
- only run eslint and phpcs linting when we have files that need it

Finally this PR cleans up all existing eslint errors. 

### How to test the Change

Verify the eslint action is passing on this PR.

### Changelog Entry
_Not sure this even needs to go in our changelog_

> Fixed - resolved some eslint issues

### Credits

Props @dkotter

### Checklist:

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -17,6 +17,7 @@ jobs:
   eslint:
     name: eslint
     runs-on: ubuntu-latest
+
     steps:
     - uses: actions/checkout@v3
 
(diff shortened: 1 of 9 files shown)


jeffpaul (approved):
Looks good, thanks for fixing all this up!

PR #408, opened 2023-03-12
Excerpt Pre-publish Check

zamanq:
### Description of the Change
This is an enhancement to the existing PR for Automatic excerpt generation: #405 

I've added the excerpt generation option in the Pre Publish Panel so that if needed the excerpt can be re-generated in that screen before publishing.

I've also made some minor changes to the button label i.e if the excerpt is not yet generated or empty the button will say "Generate excerpt" if the excerpt is already generated, then it would say "Re-generate excerpt"

### How to test the Change
The testing process is same as this PR: #405 
In addition, when clicked on the Publish button, the pre-publish panel will open and at the bottom you'll see the excerpt textarea with the button to generate/re-generate the excerpt.

Screenshot 2023-03-12 at 10 04 11 PM

### Changelog Entry
Improvement: Added excerpt generation option to the Pre Publish Panel

### Credits
@dkotter @jeffpaul @zamanq 

### Checklist:

### DIFF
--- a/src/js/post-excerpt/index.js
+++ b/src/js/post-excerpt/index.js
@@ -25,7 +25,9 @@ const PostExcerpt = () => {
 			<PluginDocumentSettingPanel title={ __( 'Excerpt' ) }>
 				<PostExcerptForm />
 			</PluginDocumentSettingPanel>
-			<MaybeExcerptPanel />
+			<MaybeExcerptPanel>
+				<PostExcerptForm />
+			</MaybeExcerptPanel>
 		</PostExcerptCheck>
 	);
 };
--- a/src/js/post-excerpt/index.js
+++ b/src/js/post-excerpt/index.js
@@ -3,7 +3,7 @@
  */
 const { __ } = wp.i18n;
 const { dispatch } = wp.data;
-const { PluginDocumentSettingPanel } = wp.editPost;
+const { PluginDocumentSettingPanel, PluginPrePublishPanel } = wp.editPost;
 const { PostExcerptCheck } = wp.editor;
 const { registerPlugin } = wp.plugins;
 
@@ -24,6 +24,13 @@ const PostExcerpt = () => {
 			<PluginDocumentSettingPanel title={ __( 'Excerpt' ) }>
 				<PostExcerptForm />
 			</PluginDocumentSettingPanel>
+			<PluginPrePublishPanel
+				title={ __( 'Excerpt' ) }
+				icon='aside'
+				initialOpen={ true }
+			>
+				<PostExcerptForm />
+			</PluginPrePublishPanel>
 		</PostExcerptCheck>
 	);
 };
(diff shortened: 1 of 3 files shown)


dkotter:
@zamanq Thanks for this PR, this is much simpler than I thought the end code would be. I've tested and everything works great.

I do have one discussion point that may or may not be the right approach. Looking at some of the other panels that get added to this section (like `Suggestion: Add tags` or `Suggestion: Assign a category`) I wonder if we should model our panel after those?

I think end result would be checking if an excerpt exists or not. If it already exists, don't show this panel at all. If it doesn't exist, show the panel with the text: `Suggestion: Generate excerpt`. As a reference, we can follow how the category panel handles this: https://github.com/WordPress/gutenberg/blob/3b04ce46f18d61aa25f0483ba49116e701835fdd/packages/editor/src/components/post-publish-panel/maybe-category-panel.js

zamanq:
@dkotter Yeah the pre publish panel can be customized to match with other models. I just did that for excerpt. So if we have excerpts already, the excerpt panel will not show in the pre publish panel, if excerpt is empty, it'll show.

One caveat is that for example, if the excerpt is empty and we display the excerpt panel in the pre publish panel and generate the excerpt, immediately after excerpt is generated in pre publish panel, the excerpt panel will disappear. This is due to the logic of not displaying the excerpt panel if excerpt already exists.

Other than that this is a nice touch.

dkotter:
> One caveat is that for example, if the excerpt is empty and we display the excerpt panel in the pre publish panel and generate the excerpt, immediately after excerpt is generated in pre publish panel, the excerpt panel will disappear

My main concern here is that the workflow here is likely someone either manually adding an excerpt or generating an excerpt in that pre-publish panel. In either case, the excerpt panel disappears before they can review the generated excerpt or manually add one.

I took a closer look at how Gutenberg handles this in both the category and tag panel and I think [following how they do the tag panel](https://github.com/WordPress/gutenberg/blob/3b04ce46f18d61aa25f0483ba49116e701835fdd/packages/editor/src/components/post-publish-panel/maybe-tags-panel.js) will help solve this for us. Not to take over this PR but I just pushed up some changes that mostly follow that code and should allow us to only show the panel if no excerpt is added but will continue to show the panel while in the pre-publish section. If you could look at that code and ensure it all looks good (making any changes you see that are needed) that would be great.

zamanq:
Hi @dkotter, I tested the changes that you did and it works fine. I think code wise it matches with what Gutenberg has right now. Thank you!

jeffpaul:
@dkotter I'll leave to you for final approval and merge back to your branch for the larger work in #405.

dkotter (approved):

dkotter:
> @dkotter I'll leave to you for final approval and merge back to your branch for the larger work in #405.

I think I'm going to leave this until #405 has been approved and then both can get merged in. I just know #405 is fairly large already so I'd prefer not to make it even larger, especially since the code here has already been reviewed.

PR #409, opened 2023-03-13
Do not cache node_modules/

szepeviktor:
It is already handled by `actions/setup-node`.
Never cache `node_modules/`. It results in an unpredictable state.
It takes 1-2 seconds to copy files from npm cache to `node_modules/`.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -21,23 +21,13 @@ jobs:
     steps:
     - uses: actions/checkout@v3
 
-    - name: Cache node_modules
-      id: cache-node-modules
-      uses: actions/cache@v3
-      env:
-        cache-name: cache-node-modules
-      with:
-        path: node_modules
-        key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
-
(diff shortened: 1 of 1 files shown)


.github/workflows/lint.yml
```diff
@@ -21,23 +21,13 @@ jobs:
...
-      with:
-        path: node_modules
-        key: ${{ runner.os }}-build-${{ env.cache-name }}-${{ hashFiles('**/package-lock.json') }}
-
     - name: Setup node v16 and npm cache
       uses: actions/setup-node@v3
       with:
         node-version: 16
-        cache: 'npm'
+        cache: npm
```
dkotter, line 28: Looking at the documentation for `actions/setup-node` they say to have quotes around this value: https://github.com/actions/setup-node#caching-global-packages-data. Not sure if it matters or not but feels like you should follow their guidance

szepeviktor: Actually all YAML strings should be quoted: https://github.com/szepeviktor/debian-server-tools/blob/master/webserver/Continuous-integration-Continuous-delivery.yml
I've removed those quotes to be in harmony with other parts of the workflow.

PR #441, opened 2023-04-26
OpenAI image generation tweaks

### LINKED ISSUE #440: OpenAI Image Generation Tweaks

The following tweaks to our OpenAI image generation functionality would help surface the feature a bit better in the editor as well as make the user flow a bit more akin to the existing Media Library experience within the editor.

1.) Like the `Media Library` tab aligns the `Filter media` and `All dates` items with the `Upload files` horizontal border, it would be great to do the same in the `Generate images` tab with the `Enter a prompt...`, `Once images...`, and `Enter prompt` items.
Screenshot 2023-04-24 at 9 10 32 PM

Screenshot 2023-04-24 at 9 10 47 PM

2.) Change the header text of `Select or Upload Media` to `Select, Upload, or Generate Media`:
Screenshot 2023-04-24 at 9 10 32 PM

3.) Like the `Select or Upload Media` modal has options for `Upload files`, `Media Library`, and `Generate images`, let's update the core image block to have `Generate image` alongside the current `Upload`, `Media Library`, `Insert from URL` options that would automatically deep-link into the `Generate images` modal tab.
Screenshot 2023-04-24 at 9 10 32 PM

Screenshot 2023-04-24 at 9 10 58 PM

4.) Update the `Enter prompt` text field to be a larger text area so that lengthy prompt inputs can display more (most?) of the prompt before a user clicks the `Generate images` button.
Screenshot 2023-04-24 at 9 10 47 PM

Screenshot 2023-04-24 at 9 11 43 PM

5.) After an image(s) are generated, maintain the prompt input in the `Enter prompt` text area in case a user wants to generate different images by tweaking the prompt (e.g. they don't like the options generated and want more to select from).  [additionally, if there's an ability in the API to create variants from a specific image result or upscale a certain image, then we should explore that via a separate feature enhancement issue]
Screenshot 2023-04-24 at 9 12 17 PM

6.) Regardless of what size image the user has set ClassifAI to generate in the plugin settings, let's only render smaller thumbnail sizes for the results so that however many are returned from OpenAI will easily display alongside eachother (and respective action buttons/links) versus now showing very large images and having to scroll significantly to view all those options.
Screenshot 2023-04-24 at 9 12 17 PM

### Designs

Screenshots of existing areas for iteration are above, if any suggestions are unclear then let me know and I can hack together samples from those screenshots to try and visually express updates.

n/a

### Code of Conduct

dkotter:
### Description of the Change

This PR refines and fixes some issues with the OpenAI image generation feature that was added in #419.

1. Ensure the CSS we need is always enqueued. Previously our CSS was in the `language-processing.scss` file but that is only enqueued if the IBM Watson integration is on. So if that feature was off, the CSS we need was never loaded, resulting in some UI issues. In addition to ensuring CSS is always enqueued, moved the media modal CSS into it's own file
2. Changed the prompt input from a normal text input into a textarea, allowing the prompt to be longer without cutting off
3. After images are generated, keep the prompt text in the input so it can easily be tweaked if another request needs to be made
4. Adjust some styling, most based on comments left on #419 after it was merged but also to reduce the size of the generated images so we get 4 across instead of 3 across
5. Add the prompt text as the alt text to any images that we import
6. Remove the secondary import button after an image is imported, as it's no longer needed
7. Moves the `Generate images` tab as the second item, before the `Media Library` tab

Partially closes #440 

Screen Shot 2023-04-26 at 11 25 09 AM

### How to test the Change

1. Setup the OpenAI DALLE image generation feature
2. Within a post, add a new Image block and then enter the media modal flow
3. Switch to the Generate image tab and ensure styling looks correct
4. Enter in a prompt and submit that, ensuring loading states show and images populate
5. Import one or more images, ensuring alt text is set

### Changelog Entry

> Added - Add the image generation prompt as `alt` text to imported generated images
> Changed - Tweaks to the image generation UI
> Fixed - Ensure proper CSS is always loaded based on the enabled features

### Credits

Props @dkotter, @jeffpaul, @mehidi258 

### Checklist:


jeffpaul:
Awesome, ship it!

PR #443, opened 2023-04-27
#433: Define $attachment_id, replace $metadata with useful information

### LINKED ISSUE #433: Undefined variables in `Read.php`
### Describe the bug

In testing something else, noticed we have a couple undefined variables within the `Read.php` file. See

https://github.com/10up/classifai/blob/b05ed69a388e5020d99ac7748da7860bafa089c6/includes/Classifai/Providers/Azure/Read.php#L128

The `$metadata` and `$attachment_id` variables are undefined there. As far as I can tell, there's no need to pass those variables in the first place so I'd suggest we just remove those.

### Steps to Reproduce

See above for details

_No response_

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

benlk:
### Description of the Change

In the section of the Azure document read function, this code change replaces some undefined variables with useful information.

- Removed undefined `$attachment_id` which was used as the second parameter on `$this->log_error()`, a method which only accepts one parameter
- Removes undefined `$metadata` as the third parameter of the `WP_Error` call — The third parameter of the `WP_Error` constructor gets saved as the error's `data` property, which code within `\Classifai\Providers\Azure\Read` does not use. Consumers of this object would have been given an undefined or empty `data` property because `$metadata` was undefined, so removing the undefined value won't break anything.
- changes the error message to include:
    - the current file size
    - the current file size limits in addition to the default.
- changes one internationalizable string to include placeholders so that we can add that additional information about file sizes.

Alternative implementations considered:
- replacing `$metadata` with `$filesize` — since `$metadata` was undefined before, we'd be changing what upstream consumers of the `WP_Error` object would see in its `data` property.
- not changing the translation string — since the data was already available within this function, I thought it best to add that data to the log message since I was in here anyways. This changeset doesn't change the error code, so it ought not affect upstream consumers of the `WP_Error` object unless they were parsing the error _message_ instead of the _code_. **Potential drawback**: Confusing consumers of the error message who rely on strict text matching.

Testing completed:
- [x] PHPCS
- [ ] ran tests
- [ ] tested on a site connected to Azure

Closes https://github.com/10up/classifai/issues/433

### How to test the Change

- [ ] Verify that the variables used are defined within the function
- [ ] Run the code on a server configured to talk to Azure, and have Azure try to scan an oversized document.

### Changelog Entry

> Added - New feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

Changed:

- When attempting to use Azure to parse a too-large document, the error log message will include the document size and the maximum document size.

### Credits

@benlk

### Checklist:

### DIFF
--- a/includes/Classifai/Providers/Azure/Read.php
+++ b/includes/Classifai/Providers/Azure/Read.php
@@ -125,7 +125,20 @@ class Read {
 
 		$filesize = filesize( get_attached_file( $this->attachment_id ) );
 		if ( ! $filesize || $filesize > computer_vision_max_filesize() ) {
-			return $this->log_error( new WP_Error( 'size_error', esc_html__( 'Document does not meet size requirements. Please ensure it is smaller than the maximum threshold (defaults to 4MB).', 'classifai' ), $metadata ), $attachment_id );
+			return $this->log_error(
+				new WP_Error(
+					'size_error',
+					esc_html(
+						sprintf(
+							// translators: %1$s is a file size
+							__( 'Document does not meet size requirements. Please ensure it is smaller than the maximum threshold (currently %1$s, defaults to 4MB).', 'classifai' ),
+							computer_vision_max_filesize()
+						)
+					),
+					$filesize
+				),
+				$this->attachment_id
+			);
 		}
 
 		/**
--- a/includes/Classifai/Providers/Azure/Read.php
+++ b/includes/Classifai/Providers/Azure/Read.php
@@ -131,7 +131,7 @@ class Read {
 					esc_html(
 						sprintf(
 							// translators: %1$s is the document file size, %2$s is the current default max filesize
-							__( 'Document (%1$s) does not meet size requirements. Please ensure it is smaller than the maximum threshold (currently %1$s, defaults to 4MB).', 'classifai' ),
+							__( 'Document (%1$s) does not meet size requirements. Please ensure it is smaller than the maximum threshold (currently %2$s, defaults to 4MB).', 'classifai' ),
 							! $filesize ? __( 'size not found', 'classifai' ) : $filesize,
 							computer_vision_max_filesize()
 						)
--- a/includes/Classifai/Providers/Azure/Read.php
+++ b/includes/Classifai/Providers/Azure/Read.php
@@ -130,8 +130,9 @@ class Read {
 					'size_error',
 					esc_html(
 						sprintf(
-							// translators: %1$s is a file size
-							__( 'Document does not meet size requirements. Please ensure it is smaller than the maximum threshold (currently %1$s, defaults to 4MB).', 'classifai' ),
+							// translators: %1$s is the document file size, %2$s is the current default max filesize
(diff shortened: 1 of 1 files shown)


includes/Classifai/Providers/Azure/Read.php
```diff
@@ -125,7 +125,20 @@ public function read_document() {
...
 		if ( ! $filesize || $filesize > computer_vision_max_filesize() ) {
-			return $this->log_error( new WP_Error( 'size_error', esc_html__( 'Document does not meet size requirements. Please ensure it is smaller than the maximum threshold (defaults to 4MB).', 'classifai' ), $metadata ), $attachment_id );
+			return $this->log_error(
+				new WP_Error(
+					'size_error',
+					esc_html(
+						sprintf(
+							// translators: %1$s is the document file size, %2$s is the current default max filesize
+							__( 'Document (%1$s) does not meet size requirements. Please ensure it is smaller than the maximum threshold (currently %1$s, defaults to 4MB).', 'classifai' ),
+							$filesize,
```
dkotter: While it's extremely unlikely (maybe even impossible due to other checks that happen earlier), in theory `$filesize` could be `false` here, since the `filesize` function can return `false`. I don't believe that will cause an error here, I think it will just cast `false` to a string, which should result in the value being presented as `0`. If we want the message to make more sense, we could do something like:
```suggestion
							! $filesize ? __( 'size not found', 'classifai' ) : $filesize,
```

includes/Classifai/Providers/Azure/Read.php
```diff
@@ -125,7 +125,20 @@ public function read_document() {
...
 		$filesize = filesize( get_attached_file( $this->attachment_id ) );
 		if ( ! $filesize || $filesize > computer_vision_max_filesize() ) {
-			return $this->log_error( new WP_Error( 'size_error', esc_html__( 'Document does not meet size requirements. Please ensure it is smaller than the maximum threshold (defaults to 4MB).', 'classifai' ), $metadata ), $attachment_id );
+			return $this->log_error(
+				new WP_Error(
+					'size_error',
+					esc_html(
+						sprintf(
+							// translators: %1$s is the document file size, %2$s is the current default max filesize
+							__( 'Document (%1$s) does not meet size requirements. Please ensure it is smaller than the maximum threshold (currently %1$s, defaults to 4MB).', 'classifai' ),
```
dkotter: `%1$s` is duplicated here. I think the second one should be `%2$s`

```suggestion
							__( 'Document (%1$s) does not meet size requirements. Please ensure it is smaller than the maximum threshold (currently %2$s, defaults to 4MB).', 'classifai' ),
```

dkotter: Might be nice to make it clear that the `$filesize` and max file size values are in bytes, since we show the default here in MB (or convert those sizes into MB so they all match)

dkotter (changes requested):

includes/Classifai/Providers/Azure/Read.php
benlk: Updated in https://github.com/10up/classifai/pull/443/commits/f9e5aa69f16eacbd58d117704fb4730a460b4083 to use bytes for all sizes.

dkotter (approved):

PR #446, opened 2023-04-28
415: Remove the authentication check when registering the Watson NLU

### LINKED ISSUE #415: Ensure we don't send unnecessary requests to the Watson API
### Describe the bug

While testing #405 I noticed I was getting HTTP requests to the Watson API even when on the OpenAI settings page. In investigating further, I was seeing requests to this API on every single page load, including the front-end. This can result in potentially thousands of extra API requests on sites with lots of traffic.

In investigating this, the issue is in the `can_register` method of the `NLU` class. The `can_register` method for each `Provider` is called on `init`, so this code runs on all requests. The problem in the `NLU` class is we make an authentication request within `can_register`, so we hit the API on every single request.

This may have just been an oversight when this was built or it may have been done purposely to ensure we always have a valid connection but I would classify this as a bug that we need to fix. I would suggest we change how this works to match what we do for [ComputerVision](https://github.com/10up/classifai/blob/develop/includes/Classifai/Providers/Azure/ComputerVision.php#L1019) and [ChatGPT](https://github.com/10up/classifai/blob/develop/includes/Classifai/Providers/OpenAI/ChatGPT.php#L225), where we check these credentials when settings are saved and store the result of that, which we then check in the `can_register` methods.

The only downside I can think of is there is a scenario where credentials are valid when they are saved but later an API key may be revoked. Checking credentials more often allows us to alert the user sooner, though I don't think we are actually outputting any error message at the moment if that happens. So I don't think there's any issue with making this change.

### Steps to Reproduce

Install a plugin like Query Monitor and inspect the HTTP requests that are made on any page load

_No response_

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

benlk:
### Description of the Change

When the Watson `NLU` Provider is registered with the plugin, instead of calling the Watson API to check whether the API key is valid, instead this now just checks whether we've saved whether the API key was previously authenticated. That means we're no longer pinging the Watson API for every request made to the server.

Closes #415

### Changelog Entry

Fixed: Removes an unnecessary Watson API authentication check on every page load.

### Credits

Props @dkotter, @benlk 

### Checklist:

### DIFF
--- a/includes/Classifai/Providers/Watson/NLU.php
+++ b/includes/Classifai/Providers/Watson/NLU.php
@@ -122,8 +122,10 @@ class NLU extends Provider {
 	 * @return bool
 	 */
 	public function can_register() {
-		if ( $this->nlu_authentication_check_failed( $this->get_settings() ) ) {
-			return false;
+		$settings = $this->get_settings();
+
+		if ( empty( $settings ) || ( isset( $settings['authenticated'] ) && false === $settings['authenticated'] ) ) {
+			false;
 		}
 
 		return true;


benlk:
⚠️ This is marked as Draft because I think there's a better way to solve this; see this comment: https://github.com/10up/classifai/issues/415#issuecomment-1528182997

dkotter:
Closing in favor of #455

PR #451, opened 2023-05-04
Transcribe audio files using the OpenAI Whisper (speech-to-text) API

dkotter:
### Description of the Change

This PR introduces a new integration with the [OpenAI Whisper](https://platform.openai.com/docs/guides/speech-to-text) (speech-to-text) API. This integration will automatically create a transcript for any supported audio files and store that transcript as the `post_content` for the item (shows in the `Description` field).

### Workflow

A new settings section is added under `Tools > ClassifAI > Language Processing > OpenAI Whisper`. Here there are three options to choose from:

1. Enter your API key
2. Turn the transcription feature on
3. Choose which roles have access

OpenAI Whisper settings

Once configured, whenever a valid audio file is uploaded (must be under 25 MB and the file type has to be one of: mp3, mp4, mpeg, mpga, m4a, wav, or webm) we send that file to the Whisper API and if we get a successful response back, we'll store the transcript as the `post_content` of the attachment item.

For existing audio items, there's a few options to generate transcripts. You can go to the Media Library grid view and click on an audio file. Within the modal that pops up, there will be an option there to (Re-) Transcribe:

Transcribe audio file in media modal

You can also go to the single media view and there will be a custom metabox with an option to (Re-) Transcribe the item. Check the box and then save the item:

Transcribe audio file in single view

If you prefer using the Media Model list view, there's a `Transcribe audio` bulk edit option as well as an inline `Transcribe` option:

| Bulk transcribe | Inline transcribe |
| --- | --- |
| Bulk transcribe | Inline transcribe  |

### How to test the Change

1. Go to `Tools > ClassifAI > Language Processing > OpenAI Whisper` and configure the feature
2. Upload a new audio file that meets the requirements
3. Go to the audio file and ensure the `Description` field has content
4. Test out the other ways to transcribe as explained above, using the media modal, using the single media view and using bulk edit options

### Changelog Entry

> Added - Automatically create transcripts of audio files using the OpenAI Whisper API

### Credits

Props @dkotter

### Checklist:


includes/Classifai/Providers/OpenAI/Whisper.php
```diff
@@ -0,0 +1,345 @@
...
+		if ( $attachment_id && ! current_user_can( 'edit_post', $attachment_id ) ) {
+			return new WP_Error( 'no_permission', esc_html__( 'User does not have permission to edit this attachment.', 'classifai' ) );
+		}
+
+		// Ensure feature is turned on.
+		if ( ! isset( $settings['enable_transcripts'] ) || 1 !== (int) $settings['enable_transcripts'] ) {
+			return new WP_Error( 'not_enabled', esc_html__( 'Transcripts are not enabled.', 'classifai' ) );
+		}
+
+		return true;
```
iamdharmesh, line 84: I think we should also add a check for `authenticated`. without this when we don't have the API key saved and enable checkbox is checked it shows the "Transcribe audio" option in bulk actions.

```php
// Check if valid authentication is in place.
if ( empty( $settings ) || ( isset( $settings['authenticated'] ) && false === $settings['authenticated'] ) ) {
	return new WP_Error( 'auth', esc_html__( 'Please set up valid authentication with OpenAI.', 'classifai' ) );
}
```

src/js/media.js
```diff
@@ -69,119 +70,142 @@ import { __ } from '@wordpress/i18n';
...
+									'attachment-details-two-column-description'
+								) ??
+								document.getElementById(
+									'attachment-details-description'
+								);
+							if (textField) {
+								textField.value = resp;
+							}
+						}
+					},
```
iamdharmesh, line 151:
```suggestion
					},
					buttonText: __('Re-transcribe', 'classifai'),
```
To keep the button text "Re-transcribe" after the process is complete. Currently, it becomes "Rescan".

iamdharmesh (dismissed):
@dkotter Thanks for the amazing work and very detailed information in the PR description. PR looks good to me and it's working fine.  

Just added 2 minor notes to check and we are ready to merge this.

Thanks.

PR #452, opened 2023-05-09
Generate post titles using OpenAI's ChatGPT

### LINKED ISSUE #434: Suggest SEO-friendly post titles

Following on from #58, there's probably also an opportunity to suggest an updated post title based on post content that would be "more SEO friendly" than what's currently set for the post.

### Designs

It seems there are two approaches here: (1) Google's "i'm feeling lucky" where ClassifAI would generate a new title and set it right away and (2) ClassifAI generating a couple options for the editor to select one from to apply as the new title or generating a new set to select from.

_No response_

### Code of Conduct

dkotter:
### Description of the Change

This PR introduces a new integration with the [OpenAI ChatGPT](https://platform.openai.com/docs/guides/chat) API. This integration gives you the ability to generate one or more (in theory, SEO-friendly) post titles that you can choose to use instead of writing your own.

Closes #434

### Workflow

A new settings section is added to `Tools > ClassifAI > Language Processing > OpenAI`. There's an existing section for generating excerpts and we add a new section here to generate titles. We use the same API key for both of those features. The unique options to choose from are:

Turn the feature on
Choose which roles have access
Choose how many titles are generated in one request

Title generation settings

Once configured, anytime someone with the proper role is editing an item that supports titles, a new `Generate title` button is added to the Core post status/summary panel:

Generate titles button

When this button is clicked, we open a modal (using the core modal component) and make a request to our custom endpoint. This endpoint will then make a request to ChatGPT, sending along the content of the item and requesting a `short, SEO-friendly title` be generated. ChatGPT will return the number of titles selected in the settings (between 1-10) and those are then rendered in the modal:

Title modal

These are rendered in individual `textarea`'s, allowing you to easily edit these titles if needed. You can then choose one of the titles to use by clicking on the `Select` button. This will then set that content as the title of your item.

### How to test the Change

1. Go to `Tools > ClassifAI > Language Processing > OpenAI` and configure the feature
2. Edit or create a new item that supports titles (like a post)
3. Add and save some content for this item
4. Click on the `Generate titles` button in the `Summary` panel and ensure a modal appears
5. Select one of the titles in this modal and ensure that the title gets updated
6. Disable the feature and ensure the button no longer shows
7. Enable the feature but disable your user's role and ensure the button still doesn't show

### Changelog Entry

> Added - Generate SEO-friendly titles automatically using OpenAI's ChatGPT

### Credits

Props @dkotter, @jeffpaul 

### Checklist:


includes/Classifai/Providers/Azure/ComputerVision.php
```diff
@@ -1317,10 +1317,10 @@ public function filter_attachment_query_keywords( $clauses ) {
 	 *
 	 * @param int    $post_id       The Post Id we're processing.
 	 * @param string $route_to_call The name of the route we're going to be processing.
-	 *
+	 * @param array  $args          Optional arguments to pass to the route.
 	 * @return array|string|WP_Error
 	 */
-	public function rest_endpoint_callback( $post_id, $route_to_call ) {
+	public function rest_endpoint_callback( $post_id, $route_to_call, $args ) {
```
iamdharmesh:
```suggestion
	public function rest_endpoint_callback( $post_id, $route_to_call, $args = [] ) {
```

iamdharmesh (commented):
Thanks for the PR and great work here @dkotter. Code LGTM and it working fine.

Just wanted to add a minor note here, otherwise all good. During testing to generate titles for multiple posts, I noticed that sometimes the title comes with the suffix `(60 characters)`. It is coming from the response from ChatGPT. We may remove it from the title before it displays on Textarea OR maybe could update the chatGPT prompt to don't include this? what do you think? 

Screenshot 2023-05-18 at 11 18 40 AM

dkotter:
@iamdharmesh

> It is coming from the response from ChatGPT. We may remove it from the title before it displays on Textarea OR maybe could update the chatGPT prompt to don't include this

I've seen the same thing myself a few times. I'm hesitant to just strip this off as this text could change at any point and we'd need to adjust our code that removes this. I've modified the prompt a bit and in testing a dozen or so times, I never saw that again. Not confident it's completely solved but I think this is good enough for now and we can always adjust this prompt in the future.

PR #455, opened 2023-05-10
Ensure we don't send unnecessary API requests to NLU

### LINKED ISSUE #415: Ensure we don't send unnecessary requests to the Watson API

dkotter:
### Description of the Change

As detailed in #415, there's a request on each page load (admin or front-end) to the NLU API to check to see if authentication is valid. These requests are unnecessary, leading to potential performance issues but more importantly, extra API usage.

This PR fixes that by refactoring the `can_register` method that all Providers use. For all other Providers besides NLU, we've removed the `can_register` method and instead rely on the parent `can_register` method.

For NLU, the `can_register` method now calls the `is_configured` method, which calls the parent `is_configured` method (matching what other providers do) but also checks the existing `classifai_configured` option to retain backwards compatibility. These changes fix the extra API requests, remove duplicate code and helps make things a little more consistent between Providers. 

Closes #415 

### How to test the Change

Ensure all Providers still work if configured properly and don't work if not configured properly

### Changelog Entry

> Changed - Standardize on how we determine if a Provider is configured
> Fixed - Avoid extra API requests to the NLU endpoint

### Credits

Props @dkotter, @benlk 

### Checklist:

### DIFF
--- a/includes/Classifai/Providers/Watson/NLU.php
+++ b/includes/Classifai/Providers/Watson/NLU.php
@@ -118,15 +118,6 @@ class NLU extends Provider {
 	}
 
 	/**
-	 * Can the functionality be initialized?
-	 *
-	 * @return bool
-	 */
-	public function can_register() {
-		return $this->is_configured();
-	}
-
-	/**
 	 * Register what we need for the plugin.
 	 */
 	public function register() {
--- a/includes/Classifai/Providers/Watson/NLU.php
+++ b/includes/Classifai/Providers/Watson/NLU.php
@@ -9,6 +9,7 @@ use Classifai\Admin\SavePostHandler;
 use Classifai\Admin\PreviewClassifierData;
 use Classifai\Providers\Provider;
 use Classifai\Taxonomy\TaxonomyFactory;
+use WP_Error;
 
 use function Classifai\get_asset_info;
 use function Classifai\get_post_types_for_language_settings;
@@ -122,11 +123,7 @@ class NLU extends Provider {
 	 * @return bool
 	 */
 	public function can_register() {
-		if ( $this->nlu_authentication_check_failed( $this->get_settings() ) ) {
-			return false;
-		}
-
-		return true;
+		return $this->is_configured();
 	}
 
 	/**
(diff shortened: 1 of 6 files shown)


includes/Classifai/Providers/Watson/NLU.php
```diff
@@ -122,11 +123,7 @@ public function reset_settings() {
 	 * @return bool
 	 */
 	public function can_register() {
```
iamdharmesh: An optional nitpick: Maybe we can remove this method? As it is similar to the parent class `Provider`.

iamdharmesh (dismissed):
Thanks for working on this and saving lots of requests @dkotter. LGTM. 🚀 

Thanks

Issue #404, opened 2023-03-06
Refactor ClassifAI so that it's easier to add more Providers under a Service.


Sidsector9:
### Is your enhancement related to a problem? Please describe.

At the moment, the plugin offers services like Image Processing, Language Processing, etc. But the problem is when you add multiple Providers to a Service, accessing data for that Provider is not convenient.

See this line for example:

https://github.com/10up/classifai/blob/1c1a180913f1c2f03f8bae4dd510d56e45ec9522/includes/Classifai/Helpers.php#L47

The direct index access of `0` makes it impossible to access settings for Providers other than the one at `0`.

Helper functions such as:

https://github.com/10up/classifai/blob/1c1a180913f1c2f03f8bae4dd510d56e45ec9522/includes/Classifai/Helpers.php#L264

...have utility outside of Watson's Language Understanding feature. In future, it we add let say, Microsoft's Text to Speech feature which is a Provider for the Language Processing Service, then the current helper `get_supported_post_statuses()` will not work even though it is required by the new Provider. This will force developers to add helpers for each Provider which will lead to code duplication.

### Describe alternatives you've considered

1. Identify existing helper functions that can be useful for multiple Providers.
2. Refactor (if possible) or deprecate in favour of functions that are generic and available to every Provider that needs it.

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

dkotter:
I started some work on this in #405, see https://github.com/10up/classifai/pull/405/files#diff-502f71f01635ed43aac831e9a91f0097f7bd09bec4e36b7dc6b52af2b694a7a0R30. My plan is to clean this up even more once that PR is merged

dkotter:
Note that PR is now merged and I've opened up a simple first step follow up in #413. That said, I think there will still be more needed, as I've seen other areas of the code that assume we only have a single provider.

PR #456, opened 2023-05-12
Ensure you can properly remove Providers

dkotter:
### Description of the Change

There's an existing filter, `{$this->menu_slug}_providers`, that allows you to filter the available Providers for each Service. This filter has been around for 4 years but in testing it out, it doesn't work properly.

The problem is we have a number of places where we reference our Provider Service with code like `$this->provider_classes[0]`. There are two issues with this:

1. We assume the Provider Service we want is always going to be the first within the `array`, which may not be true now that we've started to add multiple Services in each Provider
2. It assumes this is always an `array` that has at least one value. This is the issue that is preventing full usage of the mentioned filter

This PR fixes the second point in its entirety by making sure we always have an `array` of at least one Provider Service before we use that variable. This allows you to remove all Services from a particular Provider and not get any PHP errors.

It mostly fixes the first point as well by adding a new `find_provider_class` helper function to get the Provider Service based on the Provider name, instead of position within the `array`. I think there is still one place that references the Provider Service by `array` position but it seemed fine to leave for now.

If no Providers are registered within a Service, the Service still shows up in our settings but no inputs show and instead we show an error message:

Service removed

Options also won't show in the new onboarding:

Only have a single Service registered

This helps with #404 but doesn't completely close it 

### How to test the Change

1. Either within your theme's `functions.php` or within a custom plugin, utilize the `{$this->menu_slug}_providers` filter to remove one or more Services. Here's the code I've been using:
```php
add_action(
	'init',
	function() {
		$services = [
			'image_processing',
			'language_processing',
			'personalizer',
		];

		foreach ( $services as $service ) {
			add_filter(
				"{$service}_providers",
				function( array $providers = [] ) use ( $service ) {
					if ( 'image_processing' === $service ) {
						/**
						 * Default providers:
						 * 'Classifai\Providers\Azure\ComputerVision'
						 * 'Classifai\Providers\OpenAI\DallE'
						 */
						return [ 'Classifai\Providers\OpenAI\DallE' ];
					}

					if ( 'language_processing' === $service ) {
						/**
						 * Default providers:
						 * 'Classifai\Providers\Watson\NLU'
						 * 'Classifai\Providers\OpenAI\ChatGPT'
						 */
						return [];
					}

					if ( 'personalizer' === $service ) {
						/**
						 * Default providers:
						 * Classifai\Providers\Azure\Personalizer
						 */
						return [];
					}
				}
			);
		}
	}
);
```
2. Go to the ClassifAI settings and ensure the Services you've removed don't show any options but the ones you've left do show proper options
3. Ensure whichever Services you've left still work properly
4. Remove the above filter so all Services are back and ensure settings for all of them show and that all features work as expected
5. Ensure no PHP warnings/errors are logged at any step

### Changelog Entry

> Fixed - Ensure the `{$this->menu_slug}_providers` filter works as expected when used to remove Providers

### Credits

Props @dkotter

### Checklist:

### DIFF
--- a/includes/Classifai/Services/Service.php
+++ b/includes/Classifai/Services/Service.php
@@ -127,6 +127,7 @@ abstract class Service {
 				<?php
 				if ( empty( $this->provider_classes ) ) {
 					echo '<p>' . esc_html__( 'No providers available for this service.', 'classifai' ) . '</p>';
+					echo '</div></div>';
 					return;
 				}
 				?>
--- a/includes/Classifai/Services/Service.php
+++ b/includes/Classifai/Services/Service.php
@@ -5,6 +5,7 @@
 
 namespace Classifai\Services;
 
+use function Classifai\find_provider_class;
 use WP_Error;
 
 abstract class Service {
@@ -137,7 +138,12 @@ abstract class Service {
 						submit_button();
 					?>
 					</form>
-					<?php if ( isset( $this->provider_classes[0] ) && ! empty( $this->provider_classes[0]->can_register() ) ) : ?>
+					<?php
+					// Find the right provider class.
+					$provider = find_provider_class( $this->provider_classes, 'Natural Language Understanding' );
+
+					if ( ! is_wp_error( $provider ) && ! empty( $provider->can_register() ) ) :
+						?>
 					<div id="classifai-post-preview-app">
 						<?php
 							$supported_post_statuses = \Classifai\get_supported_post_statuses();
--- a/includes/Classifai/Services/Service.php
+++ b/includes/Classifai/Services/Service.php
@@ -123,14 +123,22 @@ abstract class Service {
 			?>
 			<div class="classifai-wrap wrap wrap--nlu">
 				<h2><?php echo esc_html( $this->display_name ); ?></h2>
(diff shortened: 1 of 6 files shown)


includes/Classifai/Services/Service.php
```diff
@@ -121,14 +123,22 @@ public function render_settings_page() {
 			?>
 			<div class="classifai-wrap wrap wrap--nlu">
 				<h2><?php echo esc_html( $this->display_name ); ?></h2>
-				<?php if ( ! empty( $this->provider_classes ) ) : ?>
+
+				<?php
+				if ( empty( $this->provider_classes ) ) {
+					echo '<p>' . esc_html__( 'No providers available for this service.', 'classifai' ) . '</p>';
```
iamdharmesh, line 129:
```suggestion
					echo '<p>' . esc_html__( 'No providers available for this service.', 'classifai' ) . '</p>';
					echo '</div></div>';
```
An optional nitpick: Adding closing `div` tags to prevent any potential UI conflicts.

iamdharmesh (dismissed):
@dkotter Thanks for working on this and making the classifAI more extendible. LGTM 🚀

PR #459, opened 2023-05-20
Feat/432 add new filters to provide more granular control

### LINKED ISSUE #432: Add new filters to provide more granular control over new OpenAI features

In #405 and #419 we added new features that rely on the OpenAI service. For both of these features, we added settings that allow you to choose which user roles can access these features. While this provides a decent amount of control, it's probably fair to say some sites may want to allow just a single user access to these features, or maybe all admins and one editor, as an example.

Some of our other features have custom filters in place allowing you to override whatever settings we have, giving more granular control on when those features should or should not work. As an example, here's the filter we run within the smart cropping method that allows you to override which images should or should not be cropped:

https://github.com/10up/classifai/blob/b05ed69a388e5020d99ac7748da7860bafa089c6/includes/Classifai/Providers/Azure/ComputerVision.php#L483

I think ideally we would add similar filters in both the `ChatGPT.php` and `DallE.php` classes.

### Designs

_No response_

_No response_

### Code of Conduct

bjorn2404:
### Description of the Change

Added filters to provide more granular control over new OpenAI features as outlined in https://github.com/10up/classifai/issues/432

Closes #432

### Changelog Entry

> Fixed comment typo in includes/Classifai/Providers/Provider.php
> Added `classifai_chatgpt_allowed_image_roles` filter to allow ChatGPT image role settings to be overridden
> Added `classifai_openai_dalle_allowed_image_roles` filter to allow DALL·E image role settings to be overridden
> Added `classifai_openai_dalle_user_can_{$feature}` filter to allow granular control for DALL·E image generation
> Added `classifai_chatgpt_user_can_{$feature}` filter to allow granular control for ChatGPT title and excerpt generation

### Credits

Props @bjorn2404 

### Checklist:


bjorn2404:
@dkotter Wanted to double check on your intention from https://github.com/10up/classifai/issues/432. You had mentioned to update `ChatGPT.php` and `DallE.php`, which appears to just output the settings in the admin. If more of an override filter is wanted I was thinking of just adding one to the bottom of `generate_image_permissions_check()` within `includes/Classifai/Services/ImageProcessing.php`. Apologies if you had something else in mind - first time looking at this plugin and just interpreting everything. Thanks!

dkotter:
@bjorn2404 Thanks for the questions!

> If more of an override filter is wanted I was thinking of just adding one to the bottom of generate_image_permissions_check() within includes/Classifai/Services/ImageProcessing.php

So this would work for image generation but only if that is triggered via the REST API. This should be true for most use cases, someone in theory could directly use the methods within the `DallE` class. I personally would recommend adding a filter to the `generate_image_callback` within the `DallE` class instead, though I think some refactoring will be needed there to make this work.

In essence what I'm thinking is it's very likely that a client will want more fine-tune control over which users can use these features (since each request costs them money). As an example, they may have a single admin and two editors that should have access but none of the other admins or editors (or other account types). While we have settings in place to limit access based on role, that wouldn't help in this situation.

So having some sort of override filter that allows you to have that level of control could be very useful and is something we do in a few other places, like within the `OCR` class or `Read` class (though to be fair, these are more about allowing you to process items that you may not otherwise and less about user permissions).

I like the idea of introducing a new method to check if a user has access to the feature, that by default can have the checks we do right now (like checking the user role matches what is in the settings) but then has a filter around the final value, allowing someone total control on who can use the feature (this is just one potential approach though).

For `ChatGPT`, we have both a `generate_excerpt` and `generate_titles` methods that would be nice to support with this feature, as well as the `generate_image_callback` for `DallE`.

bjorn2404:
@dkotter Thanks for the info! Does this look closer now to what you were thinking? I've added a `user_can_access` method to both classes where the setting name is passed through to the filter dynamically. Naming is maybe a little odd but I thought it would be nice to pass the setting name along as part of the feature filter name - ex: `classifai_chatgpt_user_can_enable_excerpt`.

Also, not sure if it's worth keeping the previous allowed_image_roles filters I added. Let me know if you think there would be of any value. Otherwise, I will update to remove them.

includes/Classifai/Providers/OpenAI/ChatGPT.php
```diff
@@ -189,6 +226,19 @@ public function setup_fields_sections() {
 		$roles = get_editable_roles() ?? [];
 		$roles = array_combine( array_keys( $roles ), array_column( $roles, 'name' ) );
 
+		/**
+		 * Filter the allowed WordPress roles for ChatGTP
+		 *
+		 * @since x.x.x
+		 * @hook classifai_chatgpt_allowed_image_roles
```
dkotter: I don't think this filter should have `image` in it, I'm assuming this was probably just a copy/paste error

includes/Classifai/Providers/OpenAI/ChatGPT.php
```diff
@@ -61,6 +61,43 @@ public function __construct( $service ) {
...
+	 * @return bool
+	 */
+	public function user_can_access( $feature, $feature_roles = [] ) {
+		$access   = false;
+		$settings = $this->get_settings();
+
+		if (
+			! empty( $settings ) &&
+			isset( $settings['authenticated'] ) &&
+			false !== $settings['authenticated'] &&
```
dkotter: I would suggest we remove the authentication checks here.

1. Seems to go against what this function says it does (checking if a user can access)
2. And we recently introduced an `is_configured` method that does this

includes/Classifai/Providers/OpenAI/ChatGPT.php
```diff
@@ -61,6 +61,43 @@ public function __construct( $service ) {
...
+			isset( $settings[ $feature ] ) &&
+			'no' !== $settings[ $feature ]
```
dkotter: Similar to the above, not sure I'd put this check here, was this checks if a feature is enabled, not if the user has access to the feature. I think I'd recommend removing this and instead adding our role checking in this method (which currently that hasn't been moved). So if we aren't authenticated or if the feature isn't turned on, no one will have access. But if both of those are true, we can then run this method to determine if an individual user should have access (by default, just checking if their role is allowed but developers can hook in and change access on a per user basis if needed).

includes/Classifai/Providers/OpenAI/ChatGPT.php
```diff
@@ -87,7 +124,7 @@ public function enqueue_editor_assets() {
 
 		if (
 			( ! empty( $excerpt_roles ) && empty( array_diff( $user_roles, $excerpt_roles ) ) )
```
dkotter: As mentioned above, I'd move this check within the new `user_can_access` method

includes/Classifai/Providers/OpenAI/ChatGPT.php
```diff
@@ -87,7 +124,7 @@ public function enqueue_editor_assets() {
...
-			&& ( isset( $settings['enable_excerpt'] ) && 1 === (int) $settings['enable_excerpt'] )
+			&& $this->user_can_access( 'enable_excerpt', $excerpt_roles )
```
dkotter: As mentioned above, I'd leave this enabled check alone rather than moving it into the new method

includes/Classifai/Providers/OpenAI/ChatGPT.php
```diff
@@ -423,7 +473,7 @@ public function generate_excerpt( int $post_id = 0 ) {
 
 		// These checks (and the one above) happen in the REST permission_callback,
 		// but we run them again here in case this method is called directly.
-		if ( empty( $settings ) || ( isset( $settings['authenticated'] ) && false === $settings['authenticated'] ) || ( isset( $settings['enable_excerpt'] ) && 'no' === $settings['enable_excerpt'] ) ) {
+		if ( ! $this->user_can_access( 'enable_excerpt' ) ) {
```
dkotter: As mentioned above, if we remove auth checks and enabled checks from the `user_can_access` method, I think these lines can stay as they are

includes/Classifai/Providers/OpenAI/DallE.php
```diff
@@ -52,22 +52,48 @@ public function __construct( $service ) {
...
 		$roles      = $settings['roles'] ?? [];
 		$user_roles = wp_get_current_user()->roles ?? [];
 
-		if (
-			current_user_can( 'upload_files' )
+		if ( current_user_can( 'upload_files' )
 			&& ( ! empty( $roles ) && empty( array_diff( $user_roles, $roles ) ) )
-			&& ( isset( $settings['enable_image_gen'] ) && 1 === (int) $settings['enable_image_gen'] )
-		) {
+			&& ( isset( $settings[ $feature ] ) && 1 === (int) $settings[ $feature ] ) ) {
```
dkotter: Similar here, I'd move this enabled check out of this method

dkotter (changes requested):

dkotter:
In addition to some of the above comments, we have user role checking that happens in our REST endpoints which will need to be handled (check out the `LanguageProcessing.php` and `ImageProcessing.php` Services)

includes/Classifai/Providers/OpenAI/ChatGPT.php
phpbits: @dkotter The fix for this is now ready for review. Thanks!

cc @jeffpaul

includes/Classifai/Providers/OpenAI/ChatGPT.php
phpbits: @dkotter, The fix for this is now ready for review. Could you point me in the right direction for the `is_configured` 
 method? This needs to be added as well. Thanks!

cc @jeffpaul

includes/Classifai/Providers/OpenAI/DallE.php
phpbits: @dkotter Could you point me in the right direction for the `is_configured` 
 method?

cc @jeffpaul

includes/Classifai/Providers/OpenAI/ChatGPT.php
phpbits: @dkotter, The fix for this is now ready for review. 

cc @jeffpaul

includes/Classifai/Providers/OpenAI/ChatGPT.php
phpbits: @dkotter, The fix for this is now ready for review. 

cc @jeffpaul

includes/Classifai/Providers/OpenAI/ChatGPT.php
phpbits: @dkotter, The fix for this is now ready for review. 

cc @jeffpaul

includes/Classifai/Providers/OpenAI/ChatGPT.php
phpbits: @dkotter, The fix for this is now ready for review. 

cc @jeffpaul

dkotter (approved):

PR #468, opened 2023-05-26
Remove quotes from generated excerpts and modify prompt slightly

dkotter:
### Description of the Change

ChatGPT tends to add quotes around things, which is not needed when generating an excerpt. This was caught and fixed while working on the generate titles feature, so adding that same functionality to excerpts.

This PR also slightly tweaks the prompt we use to generate excerpts to try and get better summaries.

### How to test the Change

Test out the excerpt generation and ensure excerpts are generated and no quotes are added around the excerpts

### Changelog Entry

> Changed - Tweak the prompt that is used to generate excerpts
> Fixed - Ensure quotes aren't added around generated excerpts

### Credits

Props @dkotter

### Checklist:

### DIFF
--- a/includes/Classifai/Providers/OpenAI/ChatGPT.php
+++ b/includes/Classifai/Providers/OpenAI/ChatGPT.php
@@ -443,7 +443,7 @@ class ChatGPT extends Provider {
 		 *
 		 * @return {string} Prompt.
 		 */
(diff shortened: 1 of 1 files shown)


ravinderk (approved):
Remove quotes from the excerpt as expected.

image

PR #476, opened 2023-06-02
Bump various dependencies

jeffpaul:
### Description of the Change

This PR will be a slow process of iterative dependency updates to see what all can be bumped across the stack and still have actions/tests pass.

### How to test the Change

Load PR locally and test ClassifAI activation, setup flows, language processing, image processing, and recommended content tasks still function as expected.

### Changelog Entry

> Security - Bumped various dependencies.

Note we can be more specific on the above once we determine when the various iterative commits here are stable enough to merge and then see what dependencies ultimately were updated.

### Credits

Props @jeffpaul.

### Checklist:


jeffpaul:
Ok I think that's about as far as I'll likely be able to get things updated... all further updates I've explored have conflicting required dependencies that are not currently able to be resolved.

ravinderk (dismissed):
@jeffpaul  `npm install` and `npm run build` update command work as expected.

dkotter:
@ravinderk This is ready for another review.

I've gone through and gotten all dependencies updated besides one (at least as of this comment, I'm sure by the time you read this something will already be out of date).

The one remaining one is `svg-react-loader`. We currently use v0.4.0 and the latest is v0.4.6. There's no issue as far as I can see in updating but doing so does introduce some new security issues (though to be fair, the current version also has security issues). For now I've left this alone, though we may want to find an alternative as it seems it may be abandoned.

In updating `husky`, `cypress` and `wp-env` to their latest versions, I did have to fix a few things, so those are what needs to be reviewed the closest. All tests are currently passing but just want to ensure everything looks good to you.

ravinderk (approved):

PR #477, opened 2023-06-02
Add bulk actions to trigger text to speech generation

dkotter:
### Description of the Change

A text-to-speech feature was added in #403. We missed adding bulk action support for that, so this PR adds that, matching how we've done this for other features.

If the text-to-speech feature is enabled for a post type, you'll now see two new options to trigger this:

1. In the `Bulk actions` dropdown, there is a new `Text to speech` option. This will run the text-to-speech generation on whichever items you have selected
Bulk action support

2. Hovering over an individual item in the post lists table, there's a new `Text to speech` link that will trigger the text-to-speech generation for that individual item
Inline support

Partially closes #460. A separate PR is coming to add `WP-CLI` support

### How to test the Change

1. With this branch checked out, enable the Text to Speech feature for a specific post type
2. Go to that post types edit screen
3. Select one or more items and then choose the `Text to speech` option in the `Bulk actions` dropdown and then hit `Apply`
4. Ensure a success message shows and that the generation of audio files worked as expected
5. Hover over a single item and click on the `Text to speech` link
6. Ensure a success message shows and that the generation of the audio file worked
7. There was some refactoring on our bulk action handling to make it more generic so if desired, test bulk classification features with Watson and OpenAI Embeddings

### Changelog Entry

> Added - Ability to trigger text-to-speech generation in bulk
> Added - Ability to trigger text-to-speech generation on an individual item from the post lists screen

### Credits

Props @dkotter

### Checklist:

- [x] I agree to follow this project's [**Code of Conduct**](https://github.com/10up/.github/blob/trunk/CODE_OF_CONDUCT.md).
- [ ] I have updated the documentation accordingly.
- [ ] I have added tests to cover my change.
- [x] All new and existing tests pass.

PR #478, opened 2023-06-02
Add `WP-CLI` command that can be used to bulk process Text to Speech data

### LINKED ISSUE #460: Add `WP-CLI` command and bulk actions handling to generate audio from text

In #403 we added the ability to generate audio from your post content using Microsoft Azure's Text to Speech API. At the moment this is only triggered when a post is saved. For sites with existing content, it would be great to add both a `WP-CLI` command to trigger this generation, as well as integrating with the bulk actions dropdown.

### Designs

_No response_

_No response_

### Code of Conduct

dkotter:
### Description of the Change

Following on from #477, this PR adds a custom `WP-CLI` command that can be used to generate Text to Speech data in bulk.

The new command looks like:
```
wp classifai text_to_speech
```

- A `post_type` to process. This will override the post IDs if both happen to be passed. We check this value against the list of supported post types that are set in settings. If it's not found, it stops processing
- A `post_status` to process. This will only be used if the `post_type` is passed. Defaults to `publish`

```
wp classifai text_to_speech 1,2,3 --dry-run=false
```

```
wp classifai text_to_speech --post_type=post --dry-run=false
```

```
wp classifai text_to_speech --post_type=page --post_status=draft --dry-run=false
```

Here's some sample output from me running a command:
```
➜ wp classifai text_to_speech --post_type=page --per_page=1 --dry-run=false
Starting processing of "page" post types that have the "publish" status in batches of 1
Batch 1 is done, proceeding to next batch
Batch 2 is done, proceeding to next batch
Batch 3 is done, proceeding to next batch
Batch 4 is done, proceeding to next batch
Success: 4 items have been processed
0 items had errors
```

Partially closes #460 

### How to test the Change

### Changelog Entry

> Added - Custom `WP-CLI` command that can be used to generate Text to Speech data in bulk 

### Credits

Props @dkotter

### Checklist:

### DIFF
--- a/includes/Classifai/Command/ClassifaiCommand.php
+++ b/includes/Classifai/Command/ClassifaiCommand.php
@@ -250,7 +250,7 @@ class ClassifaiCommand extends \WP_CLI_Command {
 
 		// If we have a post type specified, process all items in that type.
 		if ( ! empty( $opts['post_type'] ) ) {
-			\WP_CLI::log( sprintf( 'Starting processing of %s items in batches of %d', $opts['post_type'], $opts['per_page'] ) );
+			\WP_CLI::log( sprintf( 'Starting processing of "%s" post types that have the "%s" status in batches of %d', $opts['post_type'], $opts['post_status'], $opts['per_page'] ) );
 
 			$paged = 1;
 
@@ -260,7 +260,7 @@ class ClassifaiCommand extends \WP_CLI_Command {
 						'post_type'        => $opts['post_type'],
 						'posts_per_page'   => $opts['per_page'],
 						'paged'            => $paged,
-						'post_status'      => 'publish',
+						'post_status'      => $opts['post_status'],
 						'suppress_filters' => 'false',
 						'fields'           => 'ids',
 					)
--- a/includes/Classifai/Command/ClassifaiCommand.php
+++ b/includes/Classifai/Command/ClassifaiCommand.php
@@ -299,8 +299,17 @@ class ClassifaiCommand extends \WP_CLI_Command {
 			$progress_bar = \WP_CLI\Utils\make_progress_bar( 'Processing ...', count( $post_ids ) );
 
 			foreach ( $post_ids as $post_id ) {
+				// Ensure we have a valid post ID.
 				if ( ! get_post( $post_id ) ) {
 					\WP_CLI::log( sprintf( 'Item ID %d does not exist', $post_id ) );
+					$errors ++;
+					continue;
+				}
+
+				// Ensure we have a valid post type.
+				$post_type = get_post_type( $post_id );
+				if ( ! $post_type || ! in_array( $post_type, $allowed_post_types, true ) ) {
+					\WP_CLI::log( sprintf( 'The "%s" post type is not enabled for Text to Speech processing', $post_type ) );
 					$errors ++;
 					continue;
 				}
--- a/includes/Classifai/Command/ClassifaiCommand.php
+++ b/includes/Classifai/Command/ClassifaiCommand.php
@@ -274,7 +274,7 @@ class ClassifaiCommand extends \WP_CLI_Command {
 				\WP_CLI::error( sprintf( 'The "%s" post status is not valid for the "%s" post type', $opts['post_status'], $opts['post_type'] ) );
 			}
 
-			\WP_CLI::log( sprintf( 'Starting processing of "%s" post types that have the "%s" status in batches of %d', $opts['post_type'], $opts['post_status'], $opts['per_page'] ) );
+			\WP_CLI::log( sprintf( 'Starting processing of "%s" post type items that have the "%s" status in batches of %d', $opts['post_type'], $opts['post_status'], $opts['per_page'] ) );
 
 			$paged = 1;
 
--- a/includes/Classifai/Command/ClassifaiCommand.php
+++ b/includes/Classifai/Command/ClassifaiCommand.php
@@ -235,13 +235,8 @@ class ClassifaiCommand extends \WP_CLI_Command {
 			'per_page'    => 100,
 		];
 
-		$opts = wp_parse_args( $opts, $defaults );
-
-		// Only allow processing post types that are enabled in settings.
+		$opts               = wp_parse_args( $opts, $defaults );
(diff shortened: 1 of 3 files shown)


ravinderk (changes requested):
@dkotter I noticed the following issues:
1. I got a success notice for an invalid post id:
    image

**Questions:** 
1. Is there any reason we allow for processing post ids whose post types are not allowed from the setting?  Are we validating post ids?
    For example, the page post type is disabled in the setting, but I can process the page id.
    image
2. Are we validating `--per-page` option?
    image
3. Are we validating `--post-status` option?
    image
4. Can we run the command by default in dry run mode?

dkotter:
@ravinderk Thanks for the review!

> 1. I got a success notice for an invalid post id:
>    image

I've added validation now so we show an error message if an invalid post ID is passed in

> **Questions:**
> 
> 1. Is there any reason we allow for processing post ids whose post types are not allowed from the setting?  Are we validating post ids?
>    For example, the page post type is disabled in the setting, but I can process the page id.
>    image

Post ID validation has been added as mentioned above. As to your point on if we should allow processing for post types that aren't turned on in settings, I debated back and forth on this one but did change it so we validate this now. On one hand I like the idea of allowing `WP-CLI` users to override things but this could also be seen as a bug. So I'll wait and see if we get real world feedback on this before allowing setting overrides.

> 2. Are we validating `--per-page` option?
>    image

Validation for this has been added

> 3. Are we validating `--post-status` option?
>    image

Validation for this has been added

> 4. Can we run the command by default in dry run mode?

I debated on whether I wanted to add this, as it makes these commands slightly harder to run (you have to remember to add the `--dry-run=false` argument) but I do think it's a good idea so this has been added

includes/Classifai/Command/ClassifaiCommand.php
```diff
@@ -207,6 +208,162 @@ public function text( $args = [], $opts = [] ) {
...
+	 *
+	 * [--dry-run=<bool>]
+	 * : Whether to run as a dry-run. Default true
+	 *
+	 * @param array $args Arguments.
+	 * @param array $opts Options.
+	 */
+	public function text_to_speech( $args = [], $opts = [] ) {
+		$defaults = [
+			'post_type'   => false,
```
ravinderk, line 236: @dkotter Can we set default `post_type` to allowed post types instead of `false`?

ravinderk, line 236: @dkotter Can we set default post_type to allowed post types instead of false?

includes/Classifai/Command/ClassifaiCommand.php
```diff
@@ -207,6 +208,162 @@ public function text( $args = [], $opts = [] ) {
...
+			// Only allow processing post statuses that are valid for a particular post type.
+			if ( ! in_array( $opts['post_status'], get_available_post_statuses( $opts['post_type'] ), true ) ) {
+				\WP_CLI::error( sprintf( 'The "%s" post status is not valid for the "%s" post type', $opts['post_status'], $opts['post_type'] ) );
+			}
+
+			\WP_CLI::log( sprintf( 'Starting processing of "%s" post types that have the "%s" status in batches of %d', $opts['post_type'], $opts['post_status'], $opts['per_page'] ) );
+
+			$paged = 1;
+
+			do {
```
ravinderk, line 281: @dkotter I notice that there can be two possible errors with `post_type` and `post_status`. Can we add conditions to exist process upon error and display them? We should query post ids with valid parameters.

includes/Classifai/Command/ClassifaiCommand.php
```diff
@@ -207,6 +208,162 @@ public function text( $args = [], $opts = [] ) {
...
+			if ( $opts['post_type'] && ! in_array( $opts['post_type'], $allowed_post_types, true ) ) {
+				\WP_CLI::error( sprintf( 'The "%s" post type is not enabled for Text to Speech processing', $opts['post_type'] ) );
+			}
+
+			// Only allow processing post statuses that are valid for a particular post type.
+			if ( ! in_array( $opts['post_status'], get_available_post_statuses( $opts['post_type'] ), true ) ) {
+				\WP_CLI::error( sprintf( 'The "%s" post status is not valid for the "%s" post type', $opts['post_status'], $opts['post_type'] ) );
+			}
+
+			\WP_CLI::log( sprintf( 'Starting processing of "%s" post types that have the "%s" status in batches of %d', $opts['post_type'], $opts['post_status'], $opts['per_page'] ) );
```
ravinderk: @dkotter I think we allow a single post type in a command. Can we update the message?

```Starting processing of "%s" post types``` ---> ```Starting processing of "%s" post type```

ravinderk (dismissed):
@dkotter I added few suggestion but these are not blocker to this pull request. Let me know if you have any question about them.

Work fine.

includes/Classifai/Command/ClassifaiCommand.php
dkotter: So I've set this to `false` as a default to force someone to pass this argument in if they want to query by a particular post type; otherwise this allows us to default to the comma-separated list of post IDs. If we set this to the allowed list of post types, we'd have to change the order of operations a bit (default to looking at the passed-in post IDs first)

includes/Classifai/Command/ClassifaiCommand.php
dkotter: I'm not sure I follow. On the lines above (269 and 274) we ensure we have a valid post type and post status and if not, we show an error message using `WP_CLI::error`, which will output the message and exit processing. So by the time we get to this `do / while` statement, we should already have a valid post type and post status to query on

includes/Classifai/Command/ClassifaiCommand.php
ravinderk: Make sense.

includes/Classifai/Command/ClassifaiCommand.php
ravinderk: Make sense.

I misunderstood how `\WP_CLI::error` works.

ravinderk (approved):
Work Fine.

PR #487, opened 2023-06-14
#486 - Optimize calls to Microsoft Azure Text to Speech API

### LINKED ISSUE #486: Optimize calls to Microsoft Azure Text to speech API

Currently, an audio version of a post is generated everytime the post is updated, regardless of if the post content or title has changed or not. We should optimize the calls for the generation of the audio version of the post to also be mindful of the cost of the TTS service.

The approach I can see we take here, is to store an md5 hash of the content that is sent for audio generation in the post meta and compare the hash to determine if a new audio generation is required or not. 

### Designs

n/a

I looked for hooks to be able to update this behaviour but found none. Instead of using hooks, I felt this is a great addition to the codebase itself as it optimizes the calls to the API.

### Code of Conduct

joshuaabenazer:
### Description of the Change

Closes #486 

### How to test the Change

1. Create a new post on the post type that supports the Microsoft Azure Text to Speech feature.
2. Publish the post. This should also generate the audio file for the post.
3. In a new tab head over to the media library and edit the audio file for the post that got recently created. Should be the first in the list.
4. Now update the post once again without any changes. Upon refresh of the edit screen of the audio file opened in point 3  above should **not show an error** of post does not exist. This confirms that the file was not re-generated again.
5. Now make changes to content or title of the post and update again. Upon refresh of the edit screen of the audio file opened in point 3  above should **now show an error** of the post not existing anymore. This confirms that the file was re-generated due to update in post content or title.
6. Checking the media library again an updated audio file should be visible.

### Changelog Entry

> Changed - Optimized calls to Microsoft Azure Text to Speech API

### Credits

Props @joshuaabenazer 

### Checklist:

### DIFF
--- a/includes/Classifai/Admin/SavePostHandler.php
+++ b/includes/Classifai/Admin/SavePostHandler.php
@@ -200,8 +200,14 @@ class SavePostHandler {
 		$saved_attachment_id = (int) get_post_meta( $post_id, TextToSpeech::AUDIO_ID_KEY, true );
 
 		// Don't regenerate the audio file it it already exists and the content hasn't changed.
-		if ( $saved_attachment_id && ! empty( $content_hash ) && ( md5( $post_content ) === $content_hash ) ) {
-			return $saved_attachment_id;
+		if ( $saved_attachment_id ) {
+
+			// Check if the audio file exists.
+			$audio_attachment_url = wp_get_attachment_url( $saved_attachment_id );
+
+			if ( $audio_attachment_url && ! empty( $content_hash ) && ( md5( $post_content ) === $content_hash ) ) {
+				return $saved_attachment_id;
+			}
 		}
 
 		$voice        = $settings['voice'] ?? '';
--- a/includes/Classifai/Admin/SavePostHandler.php
+++ b/includes/Classifai/Admin/SavePostHandler.php
@@ -192,10 +192,18 @@ class SavePostHandler {
 			);
 		}
 
-		$normalizer   = new Normalizer();
-		$settings     = \Classifai\get_plugin_settings( 'language_processing', TextToSpeech::FEATURE_NAME );
-		$post         = get_post( $post_id );
-		$post_content = $normalizer->normalize_content( $post->post_content, $post->post_title, $post_id );
+		$normalizer          = new Normalizer();
+		$settings            = \Classifai\get_plugin_settings( 'language_processing', TextToSpeech::FEATURE_NAME );
+		$post                = get_post( $post_id );
+		$post_content        = $normalizer->normalize_content( $post->post_content, $post->post_title, $post_id );
+		$content_hash        = get_post_meta( $post_id, TextToSpeech::AUDIO_HASH_KEY, true );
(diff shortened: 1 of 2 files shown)


includes/Classifai/Admin/SavePostHandler.php
```diff
@@ -192,10 +192,18 @@ public function synthesize_speech( $post_id ) {
...
-		$settings     = \Classifai\get_plugin_settings( 'language_processing', TextToSpeech::FEATURE_NAME );
-		$post         = get_post( $post_id );
-		$post_content = $normalizer->normalize_content( $post->post_content, $post->post_title, $post_id );
+		$normalizer          = new Normalizer();
+		$settings            = \Classifai\get_plugin_settings( 'language_processing', TextToSpeech::FEATURE_NAME );
+		$post                = get_post( $post_id );
+		$post_content        = $normalizer->normalize_content( $post->post_content, $post->post_title, $post_id );
+		$content_hash        = get_post_meta( $post_id, TextToSpeech::AUDIO_HASH_KEY, true );
+		$saved_attachment_id = (int) get_post_meta( $post_id, TextToSpeech::AUDIO_ID_KEY, true );
+		if ( ! empty( $content_hash ) ) {
```
dkotter: Seems like we may want a check here to ensure the `$saved_attachment_id` actually exists, in case this was deleted somehow. We could also combine this `if` statement into a single line:
```suggestion
		if ( ! empty( $content_hash ) && md5( $post_content ) === $content_hash && $saved_attachment_id  ) {
```

dkotter: I would also maybe add a comment above this `if` statement describing what we're doing. Something like:
```php
// Don't regenerate the audio file it it already exists and the content hasn't changed.
```

dkotter (changes requested):

joshuaabenazer:
@dkotter just noticed that if the audio attachment is deleted accidentally from the media library, this update would make it such that no new audio would be generated unless we deleted the audio id meta from the post. Thinking of taking care of that as part of this PR. Thoughts?

dkotter (approved):

dkotter:
> just noticed that if the audio attachment is deleted accidentally from the media library, this update would make it such that no new audio would be generated unless we deleted the audio id meta from the post. Thinking of taking care of that as part of this PR. Thoughts?

Looks like you took care of that here, is that correct? If so, I think this is good to go

joshuaabenazer:
> Looks like you took care of that here, is that correct? If so, I think this is good to go

@dkotter yes that is correct. Let me know if there is anything further that I need to do on this. Thanks.

ravinderk (commented):
@joshuaabenazer Do we skip an audio generation when "Generate audio for this post" is set to `no` during bulk action or WP CLI command? If not, can we optimize this too?

dkotter:
> @joshuaabenazer Do we skip an audio generation when "Generate audio for this post" is set to `no` during bulk action or WP CLI command? If not, can we optimize this too?

I don't believe we do but I also think that is by design. If someone manually is triggering this via the bulk actions or via our WP-CLI command, I think that should bypass that option. That is just my opinion though, happy to discuss that but I'd suggest that should be handled in a separate issue.

PR #494, opened 2023-06-20
#493 - Re-purpose audio generation toggle to show / hide audio on post.

### LINKED ISSUE #493: Allow enable/disable of MS Azure Text to Speech on post already containing audio.

Currently, once audio is generated for a post, there is no way for a user to disable it from showing up on the post.
I recommend updating the text on the editor from "Generate audio for this post." to "Enable audio for this post" and then repurpose this meta info to determine if the post displays the audio or not ( regardless of if the audio exists or not ).

### Designs

_No response_

The only way to disable the audio from showing up on the post is to hook onto the the filter [`classifai_disable_post_to_audio_block`](https://github.com/10up/classifai/blob/develop/includes/Classifai/Providers/Azure/TextToSpeech.php#L515) and adding some custom code to enable disable audio.

### Code of Conduct

joshuaabenazer:
### Description of the Change

Closes #493 

### How to test the Change

1. Create a new post on the post type that supports the Microsoft Azure Text to Speech feature.
2. Publish the post. This should also generate the audio file for the post.
3. Check the posst on the front end and it should display the audio file for the post at the top of the post content.
4. Now disable the toggle for "Enable audio for this post" in the ClassifAI section of the post editor and update the post.
5. Now the audio should not display on the front end of the post although the audio exists and is generated.

### Changelog Entry

> Changed - Allow to show / hide audio for the post.

### Credits

Props @joshuaabenazer 

### Checklist:

### DIFF
--- a/src/js/gutenberg-plugin.js
+++ b/src/js/gutenberg-plugin.js
@@ -356,7 +356,7 @@ const ClassifAITSpeechSynthesisToggle = ( props ) => {
 	return (
 		<>
 			<ToggleControl
-				label={ __( 'Enable audio for this post.', 'classifai' ) }
+				label={ __( 'Enable audio generation', 'classifai' ) }
 				help={
 					isFeatureSupported
 						? __(
--- a/src/js/gutenberg-plugin.js
+++ b/src/js/gutenberg-plugin.js
@@ -356,7 +356,7 @@ const ClassifAITSpeechSynthesisToggle = ( props ) => {
 	return (
 		<>
 			<ToggleControl
-				label={ __( 'Generate audio for this post.', 'classifai' ) }
+				label={ __( 'Enable audio for this post.', 'classifai' ) }
 				help={
 					isFeatureSupported
 						? __(
--- a/includes/Classifai/Providers/Azure/TextToSpeech.php
+++ b/includes/Classifai/Providers/Azure/TextToSpeech.php
@@ -501,6 +501,11 @@ class TextToSpeech extends Provider {
 			return $content;
 		}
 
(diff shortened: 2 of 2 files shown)


ravinderk (commented):
@joshuaabenazer Pull request is working fine.

I have a few questions:

1. Do you think it makes sense to hide the `Text To speech` row action (list table page) when `Enable audio for this post` is set to `no`?
image

2. Does this setting impact bulk action `Text To speech`? Does it generate audio even if  `Enable audio for this post` is set to `no`?

dkotter:
> @joshuaabenazer Pull request is working fine.
> 
> I have a few questions:
> 
> 1. Do you think it makes sense to hide the `Text To speech` row action (list table page) when `Enable audio for this post` is set to `no`?
> 
> 2. Does this setting impact bulk action `Text To speech`? Does it generate audio even if  `Enable audio for this post` is set to `no`?

I think similar to my other comment, I see this as a feature, not a bug, where someone can force generation of the audio using these options (inline row action, bulk action, WP-CLI command) even if it's disabled at the post level. That said, I can see your point where someone else may see this as a bug.

This is how it works for other features as well though, like auto-classification. I'd suggest if we're going to change this, we should update all of those areas at once in a single PR instead of addressing that here.

src/js/gutenberg-plugin.js
```diff
@@ -356,7 +356,7 @@ const ClassifAITSpeechSynthesisToggle = ( props ) => {
 	return (
 		<>
 			<ToggleControl
-				label={ __( 'Generate audio for this post.', 'classifai' ) }
+				label={ __( 'Enable audio for this post.', 'classifai' ) }
```
```suggestion
				label={ __( 'Enable audio generation', 'classifai' ) }
```

dkotter (approved):

PR #514, opened 2023-06-28
Add `WP-CLI` command to bulk process audio transcriptions

### LINKED ISSUE #498: Add `WP-CLI` command to bulk transcribe audio files

In #451 we added the ability to transcribe audio files using OpenAI's Whisper API. This feature fires whenever an audio file is first uploaded or can be triggered on existing individual audio files or by selecting multiple audio files. We should look to add a `WP-CLI` command that can be used to do this processing in bulk.

### Designs

_No response_

_No response_

### Code of Conduct

dkotter:
### Description of the Change

In #451 we added the ability to generate a transcription from an audio file. This PR is a follow-up to that and adds a `WP-CLI` command that can be used to do the same thing. This is perfect for processing a large amount of audio files or having more control over which files are processed.

The new command looks like:
```
wp classifai transcribe_audio
```

- A comma-delimited list of attachment IDs to process
- A `force` argument that defaults to `false`. Will only process items that don't have data saved in `post_content` if set to `false` (basically won't re-process items that already have a transcription saved)

```
wp classifai transcribe_audio 1,2,3 --dry-run=false
```

```
wp classifai transcribe_audio --dry-run=false
```

```
wp classifai transcribe_audio --per_page=5 --force=true --dry-run=false
```

Closes #498 

### How to test the Change

### Changelog Entry

> Added - Custom `WP-CLI` command that can be used to generate audio transcriptions in bulk 

### Credits

Props @dkotter

### Checklist:

### DIFF
--- a/hookdocs/wp-cli.md
+++ b/hookdocs/wp-cli.md
@@ -109,6 +109,40 @@ The following WP-CLI commands are supported by ClassifAI:
     * `true` to run in dry-run mode
     * `false` to run in normal mode
 
+* `wp classifai transcribe_audio <attachment_ids> [--per_page=<per_page>] [--force=<bool>] [--dry-run=<bool>]`
+
+  Batch generation of audio transcriptions using OpenAI's Whisper API.
+
+  * `<attachment_ids>`: A comma-delimited list of attachment IDs to generate transcriptions for. If not set, will instead run a query to get all audio attachments to process.
+
+    default: `null`
+
+  * `[--per_page=<int>]`: How many items should be processed at a time. Will still process all items but will do it in batches matching this number. Defaults to 100.
+
+    default: `100`
+
+    options:
+
+    * N, max number of items to process at a time
+
+  * `[--force=<bool>]`: Whether to process audio files that already have a transcription set. Defaults to `false`.
+
+    default: `false`
+
+    options:
+
+    * `true` to process all items
+    * `false` to only process items that don't have transcriptions set
+
+  * `[--dry-run=<bool>]`: Whether to run as a dry-run. Defaults to `true`, so will run in dry-run mode unless this is set to `false`.
+
+    default: `true`
+
+    options:
+
+    * `true` to run in dry-run mode
+    * `false` to run in normal mode
+
 ### Image Processing Commands
 
 * `wp classifai image <attachment_ids> [--limit=<int>] [--skip=<skip>] [--force]`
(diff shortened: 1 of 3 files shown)


Sidsector9:
@dkotter there are some conflicts in this PR.

dkotter:
@Sidsector9 Thanks for pointing that out. Those should be cleaned up now

Sidsector9 (changes requested):
The code looks good 👍 

These are the points I noted:

#### 1. Logs both `error` and `success` messages

Screenshot 2023-07-11 at 12 58 23 PM

1. Disable `Generate transcripts from audio files`.
2. Run `wp classifai transcribe_audio <audio_id>`.

#### 2. Logging types can be changed

- `\WP_CLI::log( sprintf( '%d items had errors', $errors ) );` can use `\WP_CLI::error()` instead.
- `\WP_CLI::success( sprintf( '%d items would have had transcriptions added', $count ) );` can use `\WP_CLI::log()` instead.


#### 3. Minor inconsistency between CLI and GUI

When `Generate transcripts from audio files` is disabled, the `Re-Transcribe` button is not rendered, however, we can still use CLI to Re-Transcribe. The setting description says
**Automatically generate transcripts for supported audio files**, but Re-Transcribing is a manual process. So IMO we should render that button at all times, whether this setting is enabled or disabled.

Adding to that, we should add a separate setting to enable/disable the feature, and use this setting value as a conditional to decide whether to render the button in the GUI.

Let me know your thoughts on this.

dkotter:
@Sidsector9 I've updated the log types you've mentioned.

> When Generate transcripts from audio files is disabled, the Re-Transcribe button is not rendered, however, we can still use CLI to Re-Transcribe. The setting description says
> Automatically generate transcripts for supported audio files, but Re-Transcribing is a manual process. So IMO we should render that button at all times, whether this setting is enabled or disabled.
> 
> Adding to that, we should add a separate setting to enable/disable the feature, and use this setting value as a conditional to decide whether to render the button in the GUI.

I think these are both good points though I wouldn't tackle them as part of this PR. I'd suggest opening a new issue to track those things and we can look to address those in a separate PR.

dkotter:
@Sidsector9 Let me know if you have any thoughts on the above. Hoping to get this finished off so I can proceed with a 2.2.3 release. Thanks!

Sidsector9 (approved):

PR #515, opened 2023-06-29
Add OpenAI Embeddings in the Classic Editor Support

### LINKED ISSUE #496: Support OpenAI Embeddings in the Classic Editor

In #437 we added the ability to auto-classify content within your existing content structure using the OpenAI Embeddings API. While the hooks we use should work fine in both the block editor and classic editor, we only tested this feature in the block editor.

So first step here is ensuring this feature works as expected in the classic editor.

In addition, I believe we'll need to add a custom metabox that allows you to turn this feature off on a item-by-item basis, which we support in the block editor (and I believe we support in the classic editor with the similar NLU classification feature).

### Designs

_No response_

_No response_

### Code of Conduct

phpbits:
### Description of the Change
This PR adds the ClassifAI metabox for the classic editor for OpenAI Embeddings support.

Screen Shot 2023-06-29 at 8 54 52 PM

Closes [#496](https://github.com/10up/classifai/issues/496)

### How to test the Change
- Set up an integration with OpenAI in Language Processing > OpenAI Embeddings
- Ensure the Classify content option is turned on and at least one Post type that has the block editor disabled.
- In whatever taxonomy you've selected, ensure you have one or more terms saved
- Create a new piece of content and publish that content (or match whatever post status option you selected)
- Ensure one or more terms gets auto-assigned

### Credits

Props @phpbits 

### Checklist:


phpbits:
@jeffpaul @dkotter  This is now ready for code review and QA. Thanks!

includes/Classifai/Admin/Metabox.php
```diff
@@ -0,0 +1,149 @@
...
+namespace Classifai\Admin;
+
+use Classifai\Providers\OpenAI\Embeddings;
+
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * This class generates the metabox for the classic editor post types.
+ */
+class Metabox {
```
dkotter: So just my opinion here but because all the code here right now is specific to the Embeddings feature, I'd suggest we move this into that file rather than create a new class

dkotter: That said, if we want a more generic class, I think we'll want to migrate any existing meta box handling to use this. So up to you on how you want to approach that, just want to make sure we're consistent

includes/Classifai/Admin/Metabox.php
```diff
@@ -0,0 +1,149 @@
...
+	public function add_metabox( $post_type ) {
+
+		$this->post_type = $post_type;
+		$settings        = $this->embeddings->get_settings();
+
+		// Set up the embeddings metabox if the feature is enabled.
+		if ( isset( $settings['enable_classification'] ) && 1 === (int) $settings['enable_classification'] ) {
+			$embeddings_post_types = $this->embeddings->supported_post_types();
+			$classic_post_types    = get_post_types( [ 'show_in_rest' => false ], 'names', 'and' );
+			$post_types            = apply_filters( 'classifai_embeddings_classic_types', array_intersect( $embeddings_post_types, $classic_post_types ) );
```
dkotter: Can we add a filter docblock here, so this gets added to our documentation automatically?

includes/Classifai/Admin/Metabox.php
```diff
@@ -0,0 +1,149 @@
...
+
+			add_meta_box(
+				'classifai_embeddings_metabox',
+				__( 'ClassifAI', 'classifai' ),
+				array( $this, 'render_metabox' ),
+				$post_types,
+				'side',
+				'default'
```
dkotter: So I think we want to add the `'__back_compat_meta_box' => true` argument here so this metabox will never show in the Block Editor

dkotter (changes requested):

dkotter:
Would also be great to add a Cypress test here to ensure this metabox shows properly in the Classic Editor. See https://github.com/10up/classifai/pull/518/files#diff-c4172826b0f360a287099211a31d6a1b8bee386d8dc8b6b5059730bfa77d4b30R84 for an example of how to do this

includes/Classifai/Admin/Metabox.php
phpbits: Added! Thanks, @dkotter

includes/Classifai/Admin/Metabox.php
phpbits: It makes sense. I moved the code within the Embeddings feature. Thanks, @dkotter!

phpbits:
> Would also be great to add a Cypress test here to ensure this metabox shows properly in the Classic Editor. See https://github.com/10up/classifai/pull/518/files#diff-c4172826b0f360a287099211a31d6a1b8bee386d8dc8b6b5059730bfa77d4b30R84 for an example of how to do this

@dkotter I've added one but it seems to be failing. Would you mind checking it out? Thanks!

dkotter (approved):

PR #516, opened 2023-06-29
Add `WP-CLI` command to bulk process excerpts

### LINKED ISSUE #416: Add `WP-CLI` command to bulk generate excerpts

In #405 we added excerpt generation with ChatGPT. These excerpts are generated on a post by post basis as a user decides. For sites that have lots of existing content, would be nice to integrate this feature into our existing WP-CLI commands.

### Designs

_No response_

_No response_

### Code of Conduct

dkotter:
### Description of the Change

In #405 we added the ability to manually generate excerpts on individual items. This is a follow up that adds a `WP-CLI` command that can be used to process items in bulk, adding generated excerpts to those items.

One thing to note is that OpenAI has [rate limits](https://platform.openai.com/docs/guides/rate-limits/what-are-the-rate-limits-for-our-api) on their APIs, so you may run into that if processing too many items at a time. For instance, free accounts only get 3 requests per minute on the ChatGPT API, so very easy to hit that. Paid accounts go up to 3500 requests per minute, so unlikely to run into issues. We may want to explore adding rate limiting into ClassifAI at some point to ensure these limits aren't hit.

The new command looks like:
```
wp classifai generate_excerpt
```

- A `post_type` argument that will be used if no post IDs are passed in
- A `post_status` argument that will be used along with the `post_type` argument. Defaults to `publish`
- A `force` argument that defaults to `false`. Will only process items that don't have an excerpt set if set to `false`

```
wp classifai generate_excerpt 1,2,3 --dry-run=false
```

```
wp classifai generate_excerpt --post_type=post --dry-run=false
```

```
wp classifai generate_excerpt --post_type=page --per_page=5 --force=true --dry-run=false
```

Closes #416 

### How to test the Change

### Changelog Entry

> Added - Custom WP-CLI command that can be used to generate excerpts in bulk

### Credits

Props @dkotter

### Checklist:

### DIFF
--- a/includes/Classifai/Command/ClassifaiCommand.php
+++ b/includes/Classifai/Command/ClassifaiCommand.php
@@ -383,6 +383,9 @@ class ClassifaiCommand extends \WP_CLI_Command {
 	 * [--per_page=<int>]
 	 * : How many items should be processed at a time. Default 100
 	 *
+	 * [--force=<bool>]
+	 * : Whether to process items that already have an excerpt set. Default false
+	 *
 	 * [--dry-run=<bool>]
 	 * : Whether to run as a dry-run. Default true
 	 *
@@ -394,6 +397,7 @@ class ClassifaiCommand extends \WP_CLI_Command {
 			'post_type'   => false,
 			'post_status' => 'publish',
 			'per_page'    => 100,
+			'force'       => false,
 		];
 
 		$opts             = wp_parse_args( $opts, $defaults );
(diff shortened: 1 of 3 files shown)


includes/Classifai/Command/ClassifaiCommand.php
```diff
@@ -365,6 +366,188 @@ public function text_to_speech( $args = [], $opts = [] ) {
...
+					if ( '' !== trim( $post->post_excerpt ) && ! $opts['force'] ) {
+						\WP_CLI::log( sprintf( 'Item ID %d has an existing excerpt and the force option hasn\'t been set. Skipping...', $post->ID ) );
+						$errors ++;
+						continue;
+					}
+
+					$result = $chat_gpt->generate_excerpt( (int) $post->ID );
+
+					if ( is_wp_error( $result ) ) {
+						\WP_CLI::log( sprintf( 'Error while processing item ID %d: %s', $post->ID, $result->get_error_message() ) );
```
iamdharmesh: Nitpick:
```suggestion
						\WP_CLI::error( sprintf( 'Error while processing item ID %d: %s', $post->ID, $result->get_error_message() ), false );
```

includes/Classifai/Command/ClassifaiCommand.php
```diff
@@ -365,6 +366,188 @@ public function text_to_speech( $args = [], $opts = [] ) {
...
+						'suppress_filters' => 'false',
+					)
+				);
+				$total = count( $posts );
+
+				foreach ( $posts as $post ) {
+					// Don't process if an item has an existing excerpt and we aren't forcing it.
+					if ( '' !== trim( $post->post_excerpt ) && ! $opts['force'] ) {
+						\WP_CLI::log( sprintf( 'Item ID %d has an existing excerpt and the force option hasn\'t been set. Skipping...', $post->ID ) );
+						$errors ++;
```
iamdharmesh: Nitpick:
Maybe better to classify this as `skipped` and display something like `2 Items skipped` instead of grouped with errors `2 items had errors`?

iamdharmesh (dismissed):
Thanks for adding this @dkotter. This looks great and works well. I haven't tested with large dataset but looks great. 🚀 

Added some nice to have notes, but not necessary to fix to merge this

Thanks.

includes/Classifai/Command/ClassifaiCommand.php
dkotter: Ah, didn't know about the second argument. Thanks!

includes/Classifai/Command/ClassifaiCommand.php
dkotter: Makes sense, updated now

iamdharmesh (dismissed):
:shipit:

PR #518, opened 2023-06-29
Support the Text-to-Speech feature in the Classic Editor

### LINKED ISSUE #499: Ensure the Text-to-Speech feature works in the Classic Editor

In #403 we added an integration with Microsoft Azure's Text to Speech API. This feature was built specifically for the Block Editor, though the hooks used should also work in the Classic Editor. Some testing needs to be done to ensure this actually works in the Classic Editor and fixes things if not. In addition, I think we'll need to add a custom metabox where this feature can be toggled off on an item by item basis.

### Designs

_No response_

_No response_

### Code of Conduct

dkotter:
### Description of the Change

In #403 we introduced a feature that can convert your content into speech using the Microsoft Azure's Text to Speech API. This only worked in the Block Editor so this PR brings the same feature over to the Classic Editor.

There's a new meta box that shows in the Classic Editor that allows you to toggle audio generation on or off:

Classic Editor meta box to toggle speech generation

The text here mirrors what we show in the Block Editor. If this toggle is turned on, when the item is saved, we will generate an audio file for this item and connect the two.

We then show an audio preview in the same meta box:

Audio preview in meta box

Behind the scenes we use the same meta keys as are used in the Block Editor so things should all function the same. For example, if the toggle is turned off, we don't render the audio element on the front-end.

Closes #499

### How to test the Change

1. Install the Classic Editor plugin
2. Ensure you have Text to Speech configured in ClassifAI
3. Create a new item in whatever post type you have supported
4. With the toggle off, save the item. Make sure audio isn't generated
5. Turn the toggle on and save again. Ensure audio is generated and shows on the front-end
6. Turn the toggle back off and ensure the audio file stays but doesn't show on the front-end
7. Disable the Classic Editor plugin and repeat the steps to ensure Block Editor content still works as well

### Changelog Entry

> Added - Support our Text to Speech capability using Microsoft Azura Text to Speech in the Classic Editor

### Credits

Props @dkotter

### Checklist:

### DIFF
--- a/includes/Classifai/Providers/Azure/TextToSpeech.php
+++ b/includes/Classifai/Providers/Azure/TextToSpeech.php
@@ -550,7 +550,7 @@ class TextToSpeech extends Provider {
 		</p>
 
 		<?php
-		if ( $audio_id && wp_get_attachment_url( $audio_id ) ) {
+		if ( 'yes' === $process_content && $audio_id && wp_get_attachment_url( $audio_id ) ) {
 			$cache_busting_url = add_query_arg(
 				[
 					'ver' => time(),
--- a/includes/Classifai/Providers/Azure/TextToSpeech.php
+++ b/includes/Classifai/Providers/Azure/TextToSpeech.php
@@ -130,6 +130,8 @@ class TextToSpeech extends Provider {
 	public function register() {
 		add_action( 'enqueue_block_editor_assets', [ $this, 'enqueue_editor_assets' ] );
 		add_action( 'rest_api_init', [ $this, 'add_synthesize_speech_meta_to_rest_api' ] );
+		add_action( 'add_meta_boxes', [ $this, 'add_meta_box' ] );
+		add_action( 'save_post', [ $this, 'save_post_metadata' ], 5 );
 		add_filter( 'the_content', [ $this, 'render_post_audio_controls' ] );
 	}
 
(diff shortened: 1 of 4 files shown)


includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -492,6 +495,108 @@ function( $post_type ) {
...
+			<span class="description">
+				<?php
+				/* translators: %s Post type label */
+				printf( esc_html__( 'ClassifAI will generate audio for this %s when it is published or updated', 'classifai' ), esc_html( $post_type_label ) );
+				?>
+			</span>
+		</p>
+
+		<?php
+		if ( $audio_id && wp_get_attachment_url( $audio_id ) ) {
```
iamdharmesh: Maybe only show audio preview if audio generation is enabled?

image

iamdharmesh (dismissed):
Thanks for the PR @dkotter, This looks great. Just added one minor note to discuss otherwise all LGTM!

Thanks

includes/Classifai/Providers/Azure/TextToSpeech.php
dkotter: Yeah, I debated on this as I'm not sure for a UX perspective it's clear that unchecking this option should also remove the previously generated audio. But that is how we do it right now in the Block Editor so in order to try and keep parity between the two, I've updated this.

Probably worth a follow up at some point to make it more clear in both the Block and Classic Editor that unchecking this box not only won't generate audio anymore but also removes the display of any previously generated audio

iamdharmesh (approved):
🚀

PR #519, opened 2023-06-29
Replace the word `REGION` with `LOCATION`

### LINKED ISSUE #517: Azure TtS setup incorrect reference
### Describe the bug

We're displaying the following helper text:
> Text to Speech region endpoint, e.g., https://LOCATION.tts.speech.microsoft.com/. Replace REGION with the region you selected for the resource in Azure.

but are mixing up LOCATION and REGION there.  Let's pick one of those two terms and make it the same.

### Steps to Reproduce

Go through set up and get to Set up AI Providers and view MS Azure Text to Speech section, see helper text after Endpoint URL field.

_No response_

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

dkotter:
### Description of the Change

Update the helper text we show when setting up the Text to Speech feature to be more clear. On the Azure side, this setting is labeled as `Location/Region`, so I've used `LOCATION` as the placeholder (both in the admin and in our README) but I also reference `Location/Region` since that's what a user will see when getting this value from Azure.

Text to Speech settings

Closes #517

### How to test the Change

Go to the Text to Speech settings page (`wp-admin/tools.php?page=classifai&tab=language_processing&provider=azure_text_to_speech`) and ensure the text looks correct

### Changelog Entry

None

### Checklist:

### DIFF
--- a/includes/Classifai/Providers/Azure/TextToSpeech.php
+++ b/includes/Classifai/Providers/Azure/TextToSpeech.php
@@ -159,7 +159,7 @@ class TextToSpeech extends Provider {
 				'label_for'     => 'url',
 				'input_type'    => 'text',
 				'default_value' => $default_settings['credentials']['url'],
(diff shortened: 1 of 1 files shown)


jeffpaul (approved):
Makes sense to me on the rationale on text used, thanks!

PR #521, opened 2023-06-30
Add WP-CLI command to bulk process OpenAI Embeddings

### LINKED ISSUE #497: Add `WP-CLI` command to bulk classify items using the OpenAI Embeddings feature

In #437 we added a feature to classify content within your existing content structure using the OpenAI Embeddings API. We should look to add a `WP-CLI` command that can be used to process items in bulk, so sites that have existing content can take advantage of this feature. We already have a `WP-CLI` command to bulk classify items using the existing NLU classification feature, so we should add a similar one for the OpenAI Embeddings feature.

### Designs

_No response_

_No response_

### Code of Conduct

phpbits:
### Description of the Change
In [437](https://github.com/10up/classifai/pull/437) we added a feature to classify content within your existing content structure using the OpenAI Embeddings API. This PR adds the WP CLI script to do bulk embeddings using the command below:

`wp classifai embeddings`

- A `post_type` argument that defaults to `post`. Will only process items that belong to the defined value.
- - A `post_status` argument that defaults to `publish`. Will only process items that belong to the defined value.
- A `dry-run` argument that defaults to `true`. You must pass false to actually run the command

### Changelog Entry

> Added - Custom WP-CLI command that can be used to run Embeddings processing

### Credits

Props @phpbits 

### Checklist:

### DIFF
--- a/hookdocs/wp-cli.md
+++ b/hookdocs/wp-cli.md
@@ -109,6 +109,47 @@ The following WP-CLI commands are supported by ClassifAI:
     * `true` to run in dry-run mode
     * `false` to run in normal mode
 
+* `wp classifai embeddings <post_ids> [--post_type=<post_type>] [--post_status=<post_status>] [--per_page=<per_page>] [--dry-run=<bool>]`
+
+  Batch classification of items using the OpenAI Embeddings API.
+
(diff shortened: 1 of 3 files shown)


phpbits:
@jeffpaul @dkotter This functionality is now ready for code review. Thanks!

dkotter (approved):

PR #523, opened 2023-06-30
Add bulk action handling to our excerpt generation

dkotter:
### Description of the Change

In #405 we added the ability to manually generate excerpts on individual items. This is a follow up that adds bulk handling support so you can generate excerpts on multiple items in the admin UI.

There are two ways to do this:

Select multiple items in a post list view and then choose the `Generate excerpt` option from the `Bulk actions` dropdown:

Bulk action support

Or hover over a single item's row and click on the `Generate excerpt` link:

Generate excerpt inline option

Both of these options should only show up if the current user's role matches a role that has been chosen in the ChatGPT Excerpt settings (and if someone were to manually make a request to the endpoint, it should fail if their role isn't allowed).

### How to test the Change

See screenshots above for the location of the two changes introduced in this PR.

Test generating excerpts using both options and ensure they work

Ensure users without a proper role can't use these options

### Changelog Entry

> Added - Ability to generate excerpts in bulk using the `Bulk actions` dropdown
> Added - Ability to generate excerpts on an individual item from the post lists screen

### Credits

Props @dkotter

### Checklist:

### DIFF
--- a/includes/Classifai/Admin/BulkActions.php
+++ b/includes/Classifai/Admin/BulkActions.php
@@ -3,9 +3,11 @@ namespace Classifai\Admin;
 
 use Classifai\Providers\Azure\ComputerVision;
 use Classifai\Providers\Azure\TextToSpeech;
+use Classifai\Providers\OpenAI\ChatGPT;
 use Classifai\Providers\OpenAI\Embeddings;
 use Classifai\Providers\OpenAI\Whisper;
 use Classifai\Providers\OpenAI\Whisper\Transcribe;
+use function Classifai\get_post_types_for_language_settings;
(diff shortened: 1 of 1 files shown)


Sidsector9 (approved):
@dkotter this works well 👍 

I have a suggestion, maybe we can handle it iteratively. 
Sites that perform bulk operations like these will benefit from running in the background, what do you say?

dkotter:
> I have a suggestion, maybe we can handle it iteratively. Sites that perform bulk operations like these will benefit from running in the background, what do you say?

Yeah, some sort of background processor would be ideal but out of scope for this particular PR. That is something I would love to see added though

PR #525, opened 2023-07-06
Add POST endpoints for excerpt and title generation 

### LINKED ISSUE #501: Add `POST` endpoints for excerpt and title generation

For both our excerpt and title generation features, we added custom REST endpoints that are used when those features are requested (when the `Generate excerpt` or `Generate titles` buttons are clicked, for instance). These endpoints are set up as `GET` endpoints and we pass in the current post ID. We then pull the content from that matches that post ID and send that along to ChatGPT.

While this works, the problem here is if you try to use one of these features on an item using the Block Editor that hasn't been saved yet, the content that we send to ChatGPT will be whatever was previously saved (if any). This means if you're working on a post that hasn't been saved yet, no content will be sent. Or if you've made a bunch of edits but haven't saved them yet, the old content will be sent.

I think it's fine to leave these `GET` endpoints alone but we should investigate adding `POST` endpoints as well and changing our functionality to pass in the full content of a post instead of just a post ID. This will require a `POST` endpoint as we'll probably run into content length issues if we try and use a `GET` endpoint for this (as it sets all the content in the URL instead of the `body`).

So at an overview, here's what I'd see needs to be done:

1. Add new `POST` endpoints for both excerpt and title generation
2. This endpoint will accept a content param in the `body`, or maybe two params, one for post content and one for post title
3. Update our JS handler for both of these to pull content from the block editor and send it along to this new endpoint, instead of sending the post ID to the `GET` endpoint
4. Ideally the callback our new endpoint uses will be the same as the existing one, but will probably need some tweaks to support passing in the content instead of an ID

### Designs

_No response_

_No response_

### Code of Conduct

phpbits:
### Description of the Change
Add  `POST` endpoint to both Generate Titles and Generate Excerpt functionality. The endpoints accepts the following:

`id` : The post ID. `Required`.
`post_content`: The current post content.

Closes [#501](https://github.com/10up/classifai/issues/501)

### How to test the Change
Edit a post and use the Generate Titles and Generate Excerpt options.

### Credits

Props @phpbits 

### Checklist:

### DIFF
--- a/includes/Classifai/Providers/OpenAI/ChatGPT.php
+++ b/includes/Classifai/Providers/OpenAI/ChatGPT.php
@@ -462,7 +462,7 @@ class ChatGPT extends Provider {
 		// Handle all of our routes.
 		switch ( $route_to_call ) {
 			case 'excerpt':
-				$return = $this->generate_excerpt( $post_id );
+				$return = $this->generate_excerpt( $post_id, $args );
 				break;
 			case 'title':
 				$return = $this->generate_titles( $post_id, $args );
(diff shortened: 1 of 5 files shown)


phpbits:
@dkotter @jeffpaul This is now ready for code review. Thanks!

dkotter (approved):

PR #528, opened 2023-07-06
#526 - Provide ability to override Audio Player Controls via filter.

### LINKED ISSUE #526: Ability to override audio player on post

Was wondering if it would be beneficial if we are able to just takeover the markup / scripts / styles for the audio player so that it is easily customizable.

We could do this similar to how the pre option filter is done on core - https://github.com/WordPress/WordPress/blob/master/wp-includes/option.php#L150-L154

And this can be added somewhere close to where the scripts are being enqueued - https://github.com/10up/classifai/blob/develop/includes/Classifai/Providers/Azure/TextToSpeech.php#L648

### Designs

_No response_

Alternatively we can disable the audio on the post using the `classifai_disable_post_to_audio_block` filter and then adding our own filter to `the_content` and then add the player. The only thing with this is that we have to do all the similar checks that the plugin already does again and may need to keep up with it incase anything changes in the plugin.

### Code of Conduct

joshuaabenazer:
### Description of the Change

Closes #526 

### How to test the Change

1. Use the newly added `classifai_pre_render_post_audio_controls` filter to return audio player markup, and enqueue its related styles and scripts.
2. Check to see if the default ClassifAI plugin output for the TTS audio player is overridden as expected.

### Changelog Entry

> Added - Ability to override Audio Player Controls.

### Credits

Props @joshuaabenazer 

### Checklist:

### DIFF
--- a/includes/Classifai/Providers/Azure/TextToSpeech.php
+++ b/includes/Classifai/Providers/Azure/TextToSpeech.php
@@ -607,7 +607,7 @@ class TextToSpeech extends Provider {
 
 		$_post = get_post();
 
-		if ( ! $_post instanceof \WP_Post) {
+		if ( ! $_post instanceof \WP_Post ) {
 			return $content;
 		}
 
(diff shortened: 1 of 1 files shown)


includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -646,6 +647,39 @@ public function render_post_audio_controls( $content ) {
...
+		 * @param {WP_Post}  $_post                 The Post object.
+		 * @param {int}      $audio_attachment_id   The audio attachment ID.
+		 * @param {string}   $audio_attachment_url  The URL to the audio attachment file.
+		 *
+		 * @return {bool|string} Custom audio block markup. Will be appended to the post content.
+		 */
+		$markup = apply_filters( 'classifai_pre_render_post_audio_controls', false, $content, $_post, $audio_attachment_id, $audio_attachment_url );
+
+		if ( false !== $markup ) {
+			return (string) $markup . $content;
```
dkotter, line 681: Curious for your thoughts here @joshuaabenazer. The more I think about it, the more I lean towards just returning whatever custom markup someone adds using this new filter, instead of us prepending that to the post content.

For instance, if someone wanted to add a play button to the end of their content, this wouldn't allow them to do that. Since we pass the post content into the filter (as well as the full post object) thinking maybe it's up to whoever uses this filter to ensure they are appending/prepending the post content with their custom markup rather than us hardcoding that here

dkotter:
Made a few PHPCS changes and some wording tweaks but tested and this works good. Left one comment to discuss but then this can get merged in

includes/Classifai/Providers/Azure/TextToSpeech.php
joshuaabenazer: @dkotter Yes, I originally **did not** have the content appended, but I felt like it might make sense after you posted the sample code on Slack. 

It would also depend on how we word the filter too, since it is named`classifai_pre_render_post_audio_controls`, I felt it gives a sense that we are in charge of the audio only. 

Now come to think of it we should just give the control to the developer as that might make it a little less confusing of what is expected. If you agree, then I can work on making that change and possibly change the filter nomenclature too(?) 
I will be able to pick this up early next week though as I sign off early on Fridays.

Also originally I was just going to supply `$content` and `$audio_attachment_url` params to the filter since mostly we would need that only and the remaining can be derived if needed, but it is good to provide the other params since we have it already.

dkotter: > If you agree, then I can work on making that change and possibly change the filter nomenclature too(?)
> I will be able to pick this up early next week though as I sign off early on Fridays.

Yeah, this sounds good to me 👍🏻

dkotter: @joshuaabenazer Let me know if you want me to finish this off. Hoping to get a 2.2.3 release out this week and would love to include this if possible

joshuaabenazer: @dkotter got sidetracked by other priorities. I may be able to get to this on Friday or early next week. That being said I don't mind you taking over to finish this off too.

dkotter: So in looking at this again, I've changed my mind and going to merge this in as-is. Looking at the `render_post_audio_controls` method, it builds the audio markup and prepends that to the post content. So I think having this custom filter do the same thing (take whatever custom markup someone provides and then prepending it to the post content) is fine. If we get complaints or use cases that this doesn't support, happy to look at changing this in the future

PR #531, opened 2023-07-08
Add comment moderation feature

### LINKED ISSUE #331: Utilize the OpenAI Moderation API to filter content (comments?)
### Enhancement idea.

OpenAI currently offers a [Moderation](https://platform.openai.com/docs/guides/moderation/overview) endpoint that is free to use. This engine allows you to pass in text and it will classify that text within a few different categories.

Thinking it would be cool to build this into Classifai as a way to determine if user generated content is safe or not. Seems most applicable for comments (think Akismet) but could also be useful on a site that accepts user content submissions, as a first line of defense.

### Code of Conduct

kirtangajjar:
### Description of the Change

Adds comment moderation feature. 

Closes #331

### Changelog Entry

> Added - Comment Moderation feature

### Credits
Props @kirtangajjar


jeffpaul:
@kirtangajjar is this up for review or still in progress?

kirtangajjar:
@jeffpaul The PR is ready, I just have to update the description. Once it's done, then ready to review.

Sidsector9 (changes requested):
@kirtangajjar can you fix the PHP issues reported by the action?

includes/Classifai/Providers/OpenAI/Moderation.php
```diff
@@ -0,0 +1,337 @@
...
+		$comment_id = sanitize_text_field( $_GET['c'] ?? null );
+		$nonce      = sanitize_text_field( $_GET['nonce'] ?? null );
+
+		if (
+			'moderate' === $action &&
+			$comment_id &&
+			wp_verify_nonce( $nonce, 'moderate_comment' )
+		) {
+			$this->moderate_comment( $comment_id );
+			wp_redirect( '/wp-admin/edit-comments.php' );
```
Sidsector9: Since we're redirecting on the same domain, it would be good to use `wp_safe_redirect()`

includes/Classifai/Providers/OpenAI/Moderation.php
```diff
@@ -0,0 +1,337 @@
...
+	 * @return mixed
+	 */
+	public function comment_row_actions( $actions, $comment ) {
+		$nonce = wp_create_nonce( 'moderate_comment' );
+
+		$actions['moderate'] = sprintf(
+			'<a href="%s" aria-label="%s">%s</a>',
+			"edit-comments.php?a=moderate&amp;c={$comment->comment_ID}&amp;nonce={$nonce}",
+			esc_attr__( 'Moderate this comment' ),
+			__( 'Moderate' )
```
Sidsector9: Missing text domain.

includes/Classifai/Providers/OpenAI/Moderation.php
```diff
@@ -0,0 +1,337 @@
...
+	 * @param array       $actions Comment row action
+	 * @param \WP_Comment $comment Comment object
+	 * @return mixed
+	 */
+	public function comment_row_actions( $actions, $comment ) {
+		$nonce = wp_create_nonce( 'moderate_comment' );
+
+		$actions['moderate'] = sprintf(
+			'<a href="%s" aria-label="%s">%s</a>',
+			"edit-comments.php?a=moderate&amp;c={$comment->comment_ID}&amp;nonce={$nonce}",
```
Sidsector9: Is it possible to use `add_query_arg()` for better readability?

includes/Classifai/Providers/OpenAI/Moderation.php
```diff
@@ -0,0 +1,337 @@
...
+
+	/**
+	 * Prints custom column header
+	 *
+	 * @param array $columns Columns
+	 * @return array
+	 */
+	public function add_comment_list_columns( $columns ) {
+		$columns['moderation_flagged'] = __( 'Moderation Flagged' );
+		$columns['moderation_flags']   = __( 'Moderation Flags' );
```
Sidsector9: Missing text domains.

includes/Classifai/Providers/OpenAI/Moderation.php
```diff
@@ -0,0 +1,337 @@
...
+		// Ensure feature is turned on.
+		if ( ! isset( $settings['enable_moderation'] ) || 1 !== (int) $settings['enable_moderation'] ) {
+			return new WP_Error( 'not_enabled', esc_html__( 'Comment moderation is not enabled.', 'classifai' ) );
+		}
+
+		return true;
+	}
+
+	/**
+	 * Start the audio transcription process.
```
Sidsector9: This documentation may not be accurate.

includes/Classifai/Providers/OpenAI/Moderation.php
```diff
@@ -0,0 +1,337 @@
...
+					[
+						'input' => $comment->comment_content,
+					]
+				),
+			]
+		);
+
+		if ( ! is_wp_error( $api_response ) && 200 === wp_remote_retrieve_response_code( $api_response ) ) {
+			$body    = json_decode( wp_remote_retrieve_body( $api_response ), true );
+			$flagged = $body['results'][0]['flagged'];
```
Sidsector9: Are we certain the response will always contain array up to 3 levels of nesting, or should we perform a check using `isset()` just to be safe? Ignore this comment if we are sure the response format will always be consistent.

includes/Classifai/Providers/OpenAI/Moderation.php
```diff
@@ -0,0 +1,337 @@
...
+
+			if ( false === $flagged ) {
+				update_comment_meta( $comment_id, 'classifai_moderation_flagged', '0' );
+			} else {
+				$flagged_categories = $body['results'][0]['categories'];
```
Sidsector9: Same question as above

Sidsector9 (changes requested):

includes/Classifai/Providers/OpenAI/Moderation.php
kirtangajjar: The response will be consistent :)

includes/Classifai/Providers/OpenAI/Moderation.php
kirtangajjar: The response will be consistent :)

kirtangajjar:
@Sidsector9 Done. Fixed the PHPCS issues and resolved your feedback as well :)

jeffpaul:
@Sidsector9 this might be a good PR to pick up (or perhaps close and open a subsequent related one) to use as an example of adding a new feature with the v3 release; thoughts?

Sidsector9:
This would definitely be a good example. I'll raise this as a separate PR 👍

Sidsector9:
Since we're adding this to core, it would be a good example for internal developers only. Third-party devs will still need to use hooks that will ship out with v3.

jeffpaul:
@Sidsector9 sorry, yeah, I meant to effectively have v3 stable in `develop` and then utilize a PR like this to have as an example for how to extend ClassifAI for new features/service providers.

dkotter:
@Sidsector9 I've migrated all code changes here into the new Feature-first format. I still need to add E2E tests for this new feature but it is ready for another review when you have some time.

Edit: tests are there now so this should be good to go

Sidsector9 (approved):
Thanks for the PR and making the changes. Tested and works well 👍

jeffpaul:
Noting for anyone who comes here looking for more context that comments will be sent into moderation if they're flagged by OpenAI Moderation for the following categories: https://platform.openai.com/docs/guides/moderation/overview

PR #532, opened 2023-07-12
feat/488: Ability to resize post content using OpenAI's ChatGPT

### LINKED ISSUE #488: `Expand text` and `Reduce text` post content functions

We've added the ability to set Post Title and Excerpts via OpenAI ChatGPT, so continuing on that API integration we could look at adding a Block Editor block toolbar option to "Expand" or "Reduce" content.  The use case here is someone has existing post content (mainly considering one-to-many paragraph blocks) that they would click into a single block or select multiple blocks and then select an "Expand" or "Reduce" option that would either make the content longer or make the content shorter by sending it to ChatGPT (likely with the entire post content and maybe meta for context) and then the responding expanded or reduced content would render in place of the single or multiple blocks.

### Designs

We'll want to consider what the two options are actually called/labeled, what any associated icon with them should be, as well as where they should appear in the block controls.  But for now just two "Expand text" and "Reduce text" options as a primary toolbar option would be a perfect start as a prototype PR and allow us time to fine tune copy/icon/location.

I would imagine these two new options appearing in the expanded paragraph block list of options when selecting a single paragraph block here:

Screenshot 2023-07-07 at 11 41 49 AM

When selecting text across more than one block, I would imagine the options in the expanded list of options here:

Screenshot 2023-07-07 at 11 43 33 AM

One side effect to be aware of here is if someone selects multiple blocks and some are NOT paragraph blocks that that content (e.g., images) will likely not be returned by ChatGPT and thus be removed when the resulting ChatGPT response overwrites that content.

### Code of Conduct

Sidsector9:
### Description of the Change

The feature adds the ability to "grow" and "shrink" the paragraph block content. (Watch video below for demo)

- The feature is implemented so as to be consistent with title generation feature using the same service.
- Instead of repeating calls to the API to re-generate the content until it is desirable by the editor, the editor can set the number of suggestions in a single call, and then choose the most desirable suggestion from the modal window, thus providing granular control.
- The modal window shows some additional stats regarding the increase/decrease in word and character count to help the editor make a more informed decision while selecting from the list of suggestions.
- The name and icons for the 2 menu options are not final and is open for suggestions.

### Limitations

- Currently, this feature only works with the paragraph block.
- At a time, it resizes content for a single block at a time. We can iteratively improve this to handle the situation where a a non-paragraph block is between 2 paragraphs.

### What is next?

The next improvement can be resizing content when multiple, consecutive paragraph blocks are selected.

Closes #488

#### 1. Resizing content settings:

Screenshot 2023-07-12 at 9 53 21 PM

#### 2. Grow content

https://github.com/10up/classifai/assets/17757960/bec85aa0-80fa-49f1-ace4-e3e0b73a422d

#### 3. Shrink content

https://github.com/10up/classifai/assets/17757960/45bf0f2d-f7f9-4250-9496-e491df44c448

### How to test the Change

1. Enable the feature under `/wp-admin/tools.php?page=classifai&tab=language_processing&provider=openai_chatgpt`
2. Go to a post with content

### Changelog Entry

> Added - Ability to resize post content using OpenAI's ChatGPT
> Changed - Bump WordPress minimum from 5.7 to 5.8

### Credits

Props @jeffpaul, @Sidsector9, @dkotter 

### Checklist:


jeffpaul:
@Sidsector9 I chatted with Jake on this today and agreed the best option would be to add a new block toolbar option using either https://github.com/10up/classifai/blob/develop/assets/img/icon.svg or https://github.com/10up/classifai/blob/develop/assets/img/menu-icon.svg for the toolbar icon and then nesting the expand/reduce options in a dropdown there.

jeffpaul:
Also, let's remove Subscriber from one of the default enabled roles for this (and maybe Contributor as well?) so that we don't get normal site visitors using up a sites OpenAI budget without an admin explicitly enabling that.

includes/Classifai/Services/LanguageProcessing.php
```diff
@@ -512,4 +522,83 @@ public function generate_post_title_permissions_check( WP_REST_Request $request
...
+	}
+
+	/**
+	 * Check if a given request has access to resize content.
+	 *
+	 * @param WP_REST_Request $request Full data about the request.
+	 * @return WP_Error|bool
+	 */
+	public function resize_content_permissions_check( WP_REST_Request $request ) {
+		return true;
```
dkotter: Assuming this was there for test purposes

includes/Classifai/Services/LanguageProcessing.php
```diff
@@ -142,6 +142,16 @@ public function register_endpoints() {
...
 			]
 		);
+
+		register_rest_route(
+			'classifai/v1/openai',
+			'resize-content',
+			[
+				'methods'             => WP_REST_Server::EDITABLE,
+				'callback'            => [ $this, 'resize_content' ],
+				'permission_callback' => [ $this, 'resize_content_permissions_check' ],
```
dkotter, line 189: We should add an `args` field here and detail out all the various arguments that can be passed in, similar to what we do on other endpoints

includes/Classifai/Services/LanguageProcessing.php
```diff
@@ -512,4 +522,83 @@ public function generate_post_title_permissions_check( WP_REST_Request $request
...
+		}
+
+		$settings = \Classifai\get_plugin_settings( 'language_processing', 'ChatGPT' );
+
+		// Check if valid authentication is in place.
+		if ( empty( $settings ) || ( isset( $settings['authenticated'] ) && false === $settings['authenticated'] ) ) {
+			return new WP_Error( 'auth', esc_html__( 'Please set up valid authentication with OpenAI.', 'classifai' ) );
+		}
+
+		// Check if title generation is turned on.
```
dkotter: This comment needs updated

src/js/gutenberg-plugins/content-resizing-plugin.js
```diff
@@ -0,0 +1,445 @@
+/* eslint-disable @wordpress/no-unsafe-wp-apis */
+import { store as blockEditorStore, BlockControls } from '@wordpress/block-editor';
```
dkotter: Some eslint errors in this file that need addressed

includes/Classifai/Services/LanguageProcessing.php
```diff
@@ -142,6 +142,16 @@ public function register_endpoints() {
 				'permission_callback' => [ $this, 'generate_post_title_permissions_check' ],
 			]
 		);
+
+		register_rest_route(
+			'classifai/v1/openai',
+			'resize-content',
+			[
+				'methods'             => WP_REST_Server::EDITABLE,
```
dkotter: Does this need to be `EDITABLE` or can we use `CREATABLE`, which will only support `POST` requests?

dkotter (changes requested):

Sidsector9:
As per the 1:1 with @dkotter, we concluded that we can keep the feature what we have now as it is, and maybe separately work on detecting the start of a word if an editor accidentally makes a text selection abruptly. For starters, we can detect space characters before a word to identify the start of the word.

Rest of the review changes are done.

cc: @jeffpaul

ankitguptaindia:
Hello there, I noticed when utilizing the Grow/Shrink feature with AI-generated content, certain inline styles such as Bold, Italic, Links, and others are being overridden by the newly generated content. This behavior may lead to the loss of important formatting if the author didn't notice and override content.

**More Details-**



https://github.com/10up/classifai/assets/6384299/432eaca4-6cb4-4d5e-80f6-3293622429ab




**Steps to Reproduce:**

1. Access the paragraph that allows using AI to Grow/Shrink content.
2. Apply the Grow/Shrink functionality to the existing content containing inline styles (e.g., Bold, Italic, Links).
3. Observe that the inline styles are replaced or removed in the AI-generated content.
 

**Proposed Solution:**

We can simply add a message in the new content modal so the author will know before importing newly created content by AI. 

**Example:**


Screenshot 2023-08-02 at 7 52 27 PM

Sidsector9:
@ankitguptaindia that is expected in this implementation as we send plaintext data to ChatGPT and receive in the same format.

includes/Classifai/Providers/OpenAI/ChatGPT.php
```diff
@@ -708,6 +804,105 @@ public function generate_titles( int $post_id = 0, array $args = [] ) {
...
+		 *
+		 * @return {array} Request body.
+		 */
+		$body = apply_filters(
+			'classifai_chatgpt_resize_content_request_body',
+			[
+				'model'       => $this->chatgpt_model,
+				'messages'    => [
+					[
+						'role'    => 'user',
```
dkotter, line 918: Wondering if similar to what we just did for our excerpt and title generation features, we should look to use `system` messages here? I may be wrong but that seems like the approach OpenAI suggests and also seems to provide more flexibility (for instance if you want to provide more in-depth instructions)

dkotter:
Left one minor comment but this looks really good, nice work here @Sidsector9!

Not something we need to change right now but we may want to consider making our prompts a little more specific. Rather than asking to increase or decrease by a small margin, we could ask to increase or decrease by a specified amount (characters, words, lines). I notice right now that sometimes the choices it returns are rather large, so giving it some parameters (increase this content between 2-4 sentences, as an example) may give better results.

includes/Classifai/Providers/OpenAI/ChatGPT.php
```diff
@@ -717,6 +814,109 @@ public function generate_titles( int $post_id = 0, array $args = [] ) {
...
+			[
+				'model'       => $this->chatgpt_model,
+				'messages'    => [
+					[
+						'role'    => 'system',
+						'content' => $prompt,
+					],
+					[
+						'role'    => 'user',
+						'content' => esc_html( $args['content'] ) . '',
```
dkotter: Do we need the extra concatenation here?
```suggestion
						'content' => esc_html( $args['content'] ),
```

dkotter (dismissed):

jeffpaul:
Note that we're waiting on some UX review from @jakemgold before merging/releasing this PR

Sidsector9:
## Update

Then changes requested by Jake are added. @jeffpaul as discussed on Slack, this PR is waiting on you for the updated copy to use for the heading and initial sentences.

includes/Classifai/Providers/OpenAI/ChatGPT.php
dkotter: @Sidsector9 Just following up again on this one

src/js/gutenberg-plugins/content-resizing-plugin.js
```diff
@@ -0,0 +1,569 @@
...
+		<Modal
+			title={ __( 'Suggestions', 'classifai' ) }
+			isFullScreen={ false }
+			className="classifai-content-resize__suggestion-modal"
+			onRequestClose={ () => {
+				setIsModalOpen( false );
+				resetStates();
+			} }
+		>
+			<p>{ __( 'Click on a row to apply:' ) }</p>
```
dkotter: Not sure we need this paragraph. It's not super accurate (you're not clicking on the row but on the button) and I think it's clear enough what you need to do

src/js/gutenberg-plugins/content-resizing-plugin.js
```diff
@@ -0,0 +1,569 @@
...
+	// We don't want to use the reszing feature when multiple blocks are selected.
+	// Nor do we want to use it when a block's content resizing is under process.
+	if ( isMultiBlocksSelected || isResizing ) {
+		return null;
+	}
+
+	// Result Modal JSX.
+	const suggestionModal = ! isResizing && textArray.length && isModalOpen && (
+		<Modal
+			title={ __( 'Suggestions', 'classifai' ) }
```
dkotter: Personal opinion but right now the title of the modal is `Suggestions` and then right below that we show `Suggestion` in the table header. Seems redundant to have these one after the other. Wondering if we should change this title to something else, like `Review Suggestions`, `Choose a Suggestion`, `Review Text`, `Select a Suggestion`, something else?

src/js/gutenberg-plugins/content-resizing-plugin.js
```diff
@@ -0,0 +1,569 @@
...
+	} );
+
+	if ( document.getElementById( 'classifai-content-resize__mock-content' ) ) {
+		document.getElementById(
+			'classifai-content-resize__mock-content'
+		).innerHTML = formattedCharArray.join( ' ' );
+	}
+
+	setTimeout( () => {
+		requestAnimationFrame( () => processAnimation( content ) );
```
dkotter: Error here because this function isn't defined.

dkotter (changes requested):

dkotter:
Left a few suggestions but also saw a couple other things:

1. Getting a deprecation error in the console: `Using custom components as toolbar controls is deprecated since version 5.6. Please use ToolbarItem, ToolbarButton or ToolbarDropdownMenu components instead. See: https://developer.wordpress.org/block-editor/components/toolbar-button/#inside-blockcontrols`
2. ~~E2E test is failing~~ Fixed now [141ea55](https://github.com/10up/classifai/pull/532/commits/141ea558c9ea8c64d0853f58c2effb94f2204fe4)
3. The animation we show when loading results has issues when we select a short paragraph. It expands outside of it's container into the content below. Not a huge deal but would be nice to solve:

screenshot

Sidsector9:
@dkotter this feature is not working well on WP 6.3 due to major changes in the Block editor. It works fine in WP 6.2. A trick to make this feature work in WP 6.3 is to enable Custom Fields under Preferences > Panels.

The animation works on a single line content as well:

https://github.com/10up/classifai/assets/17757960/2b38116d-6001-4564-bc56-9f270e4a4b64




I couldn't find a way to fix this, I'm still investigating.

src/js/gutenberg-plugins/content-resizing-plugin.js
dkotter: So looked at the failing E2E test and the issue was we were clicking on the row, not the button. I've fixed that in [141ea55](https://github.com/10up/classifai/pull/532/commits/141ea558c9ea8c64d0853f58c2effb94f2204fe4) but I guess wondering now if the intention is to allow the user to click on the entire row to insert? Or is clicking on the button good enough?

jeffpaul:
@Sidsector9 since this isn't a functional issue with the loading animation, perhaps remove that for now and open that in a separate PR to continue iterating on and can land in a 2.3.x minor release.  Also, I confirmed and the existing "Expand this text" and "Condense this text" copy is good for us to ship.

src/js/gutenberg-plugins/content-resizing-plugin.js
dkotter: @Sidsector9 Just flagging this again. Let me know if this is a feature that we want and just isn't working. Otherwise I'd suggest we remove this text

dkotter:
@Sidsector9 @jeffpaul Reviewing this again today in hopes of getting this merged and ran into one thing. Right now our E2E test is failing on WordPress 5.7 (our minimum) because we're using the `ToolbarDropdownMenu` component which wasn't added to core until 5.8.

I can see two paths forward:

1. Find a different approach for that dropdown menu that will work in 5.7+
2. Bump our minimum supported WordPress version to at least 5.8

Not sure how hard the first will be. I'm totally on board to bump our minimums but open to thoughts there.

Sidsector9:
@dkotter I've bumped minimum WordPress version from 5.7 to 5.8

dkotter (approved):

PR #536, opened 2023-07-13
Initial integration into the new Command Palette

dkotter:
### Description of the Change

Gutenberg v15.6.0 shipped with a new Command Palette that can be triggered within the site editor (by using the keyboard shortcut command+k). This feature should be shipping with WP 6.3. This PR does some initial integrations into this Command Palette but there are certainly more things we could add here (or in follow-up PRs).

At the moment, there are 4 commands we add here:

1. A link to the general ClassifAI settings page
2. Ability to trigger excerpt generation
3. Ability to trigger title generation
4. A link that takes you to the new stand-alone image generation screen

https://github.com/10up/classifai/assets/916738/0302f108-7305-468c-af95-5dbf25d516df

### How to test the Change

1. Ensure you're running the Gutenberg plugin, v15.6.0 or greater
2. Check this PR out and run `npm install && npm run build`
3. Create a new post or edit an existing post
5. Click into the post content (I found I couldn't trigger the command center until I clicked on something) and then use the shortcut command+k
6. In the modal that pops up, search for `ClassifAI: Generate excerpt`, `ClassifAI: Generate title`, `ClassifAI: Generate image`, or `ClassifAI settings`
7. Click on all three separately and ensure they work

### Changelog Entry

> Added - Initial integration with the new Command Palette API

### Credits

Props @dkotter

### Checklist:

### DIFF
--- a/includes/Classifai/Plugin.php
+++ b/includes/Classifai/Plugin.php
@@ -189,13 +189,15 @@ class Plugin {
 			]
 		);
 
-		wp_enqueue_script(
-			'classifai-commands',
-			CLASSIFAI_PLUGIN_URL . 'dist/commands.js',
-			get_asset_info( 'commands', 'dependencies' ),
-			get_asset_info( 'commands', 'version' ),
-			true
-		);
+		if ( wp_script_is( 'wp-commands', 'registered' ) ) {
+			wp_enqueue_script(
+				'classifai-commands',
+				CLASSIFAI_PLUGIN_URL . 'dist/commands.js',
+				get_asset_info( 'commands', 'dependencies' ),
+				get_asset_info( 'commands', 'version' ),
+				true
+			);
+		}
 	}
 
 	/**
--- a/includes/Classifai/Plugin.php
+++ b/includes/Classifai/Plugin.php
@@ -188,6 +188,14 @@ class Plugin {
 				'ajax_nonce'   => wp_create_nonce( 'classifai' ),
 			]
 		);
+
+		wp_enqueue_script(
+			'classifai-commands',
+			CLASSIFAI_PLUGIN_URL . 'dist/commands.js',
+			get_asset_info( 'commands', 'dependencies' ),
+			get_asset_info( 'commands', 'version' ),
+			true
+		);
 	}
 
 	/**
--- a/src/js/gutenberg-plugins/commands.js
+++ b/src/js/gutenberg-plugins/commands.js
@@ -27,7 +27,7 @@ const Commands = () => {
 		if ( excerptButton ) {
 			commands.push( {
 				name: 'classifai/generate-excerpt',
-				label: __( 'Generate excerpt', 'classifai' ),
+				label: __( 'ClassifAI: Generate excerpt', 'classifai' ),
 				icon: edit,
 				callback: ( { close } ) => {
 					close();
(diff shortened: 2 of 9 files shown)


includes/Classifai/Plugin.php
```diff
@@ -188,6 +188,14 @@ public function enqueue_admin_assets() {
 				'ajax_nonce'   => wp_create_nonce( 'classifai' ),
 			]
 		);
+
+		wp_enqueue_script(
```
iamdharmesh: Getting Missing Dependencies error with the current version of WP. Maybe we should enqueue the script conditionally only when the `wp-commands` script is registered? which will be registered in WP 6.3 as mentioned in the PR description.

```php
if ( wp_script_is( 'wp-commands', 'registered' ) ) {
   // Enqueue script
}
```

image

src/js/gutenberg-plugins/commands.js
```diff
@@ -0,0 +1,71 @@
+import { useCommandLoader } from '@wordpress/commands';
+import { edit, settings } from '@wordpress/icons';
+import { __ } from '@wordpress/i18n';
+import { registerPlugin } from '@wordpress/plugins';
+
+const Commands = () => {
+	const getCommandLoader = () => {
+		const commands = [];
+		const excerptButton = document.querySelector(
+			'.editor-post-excerpt button'
```
iamdharmesh: Nitpick: Maybe add some class to the button and add it here to make it more solid? This is just a minor note and not a blocker for merging this PR.

iamdharmesh (changes requested):
Thanks for the PR @dkotter, Code looks good and it tests well. However, I noticed 1 warning about missing dependencies. otherwise all good to proceed with merging this.

Note: My OpenAI key is expired so, not tested title generation etc.. but the command properly triggers the operation for generating titles and excerpt.

ravinderk (changes requested):
@dkotter, do you think we can make our command titles unique?  It will be difficult for the admin to differentiate between two commands if they have the same title.

includes/Classifai/Plugin.php
dkotter: Ah, nice catch. Forgot to test with the Gutenberg plugin disabled. This should be fixed now

src/js/gutenberg-plugins/commands.js
dkotter: Yeah, this is a good call out. I've added that

dkotter:
> @dkotter, do you think we can make our command titles unique? It will be difficult for the admin to differentiate between two commands if they have the same title.

Are you thinking of just adding `ClassifAI` to the commands? So it would be something like `ClassifAI: Generate excerpt`?

ravinderk:
> > @dkotter, do you think we can make our command titles unique? It will be difficult for the admin to differentiate between two commands if they have the same title.
> 
> Are you thinking of just adding `ClassifAI` to the commands? So it would be something like `ClassifAI: Generate excerpt`?

@dkotter I was thinking the same.
Adding `ClassifAI` to the command title will make the command easily discoverable. Admin can quickly get the list of commands supported by this plugin.

ravinderk (dismissed):

iamdharmesh (approved):
:shipit:

PR #539, opened 2023-07-17
#538 - Add setting to toggle the default behaviour of audio generation on a post.

### LINKED ISSUE #538: Text To Speech: Provide option to set default setting of audio generation on new posts

Currently when a post that does not have audio generated for the content is opened for editing or when a new post is created  the default option in the sidebar for TTS is to enable audio generation upon Update or Publish.

It would be nice to allow plugin users to set this option to be enabled or disabled giving more control to cut costs and be mindful of automatic generation if need be. I propose to add a setting to the TTS module that allows for this.

### Designs

Screenshot 2023-07-17 at 6 36 59 PM
Screenshot 2023-07-17 at 6 38 59 PM

_No response_

### Code of Conduct

joshuaabenazer:
### Description of the Change

Provides an extra setting for TTS to change the default behaviour of the the toggle to generate audio on posts.

Closes #538

### How to test the Change

1. Go to Tools -> ClassifAI -> Language Processing -> Microsoft Azure
2. A new setting "Enabled by default" should be present and should be checked by default.
    1. Review that the "Enable audio generation" sidebar Gutenberg setting on new post or posts that do not have audio generated is enabled.
    2. Review to make sure this update to the code does not change the way new posts behave in terms of audio generation and adheres to how it was before the code change.
3. Now uncheck the setting from point#2 
    1. Review that  the "Enable audio generation" sidebar Gutenberg setting on new post or posts that do not have audio generated is disabled.
    2. Review to make sure this toggle is adhered to and upon update the toggle the audio is generated if it does not exists.
    3. Review to make sure if the audio is available for a post and this toggle is then disabled it does not show up on the frontend.

### Changelog Entry

> Added - Setting to toggle the default behaviour of audio generation on a post.

### Credits

Props @joshuaabenazer 

### Checklist:

### DIFF
--- a/includes/Classifai/Providers/Azure/TextToSpeech.php
+++ b/includes/Classifai/Providers/Azure/TextToSpeech.php
@@ -194,6 +194,20 @@ class TextToSpeech extends Provider {
 			]
 		);
 
+		add_settings_field(
+			'default',
+			esc_html__( 'Enabled by default', 'classifai' ),
+			[ $this, 'render_input' ],
+			$this->get_option_name(),
+			$this->get_option_name(),
+			[
+				'label_for'      => 'default',
+				'input_type'     => 'checkbox',
+				'default_value'  => $default_settings['default'],
+				'description'    => esc_html__( 'Enables the toggle to generate audio on posts by default.', 'classifai' ),
+			]
+		);
+
 		if ( ! empty( $voices_options ) ) {
 			add_settings_field(
 				'voice',
@@ -262,6 +276,19 @@ class TextToSpeech extends Provider {
 				$current_settings['post_types'][ $post_type->name ] = $settings['post_types'][ $post_type->name ];
 			} else {
 				$current_settings['post_types'][ $post_type->name ] = null;
+			}
+		}
+
+		$checkbox_settings = [
+			'default',
+		];
+
+		foreach ( $checkbox_settings as $checkbox_setting ) {
+
+			if ( empty( $settings[ $checkbox_setting ] ) || 1 !== (int) $settings[ $checkbox_setting ] ) {
+				$current_settings[ $checkbox_setting ] = 'no';
+			} else {
+				$current_settings[ $checkbox_setting ] = '1';
 			}
 		}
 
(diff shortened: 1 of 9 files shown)


includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -436,10 +463,30 @@ private function get_default_settings() {
...
 	}
 
+	/**
+	 * Helper to get the settings and allow for settings default values.
+	 *
+	 * @param string|bool|mixed $index Optional. Name of the settings option index.
+	 *
+	 * @return string|array|mixed
+	 */
+	public function get_settings( $index = false ) {
```
dkotter: So this method is already defined in the parent `Provider` class. From comparing the two, the only difference is this one sets defaults properly, is that correct? If so, I'm inclined to make this change to the default method (which I think would involve adding a default `get_default_settings` method. That way we aren't overwriting that method here

dkotter (changes requested):

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -194,6 +194,20 @@ public function setup_fields_sections() {
...
+			'default',
+			esc_html__( 'Enabled by default', 'classifai' ),
+			[ $this, 'render_input' ],
+			$this->get_option_name(),
+			$this->get_option_name(),
+			[
+				'label_for'     => 'default',
+				'input_type'    => 'checkbox',
+				'default_value' => $default_settings['default'],
+				'description'   => esc_html__( 'Enables the toggle to generate audio on posts by default.', 'classifai' ),
```
```suggestion
				'description'   => esc_html__( 'Determines if audio generation is on by default for individual items.', 'classifai' ),
```

includes/Classifai/Providers/Azure/TextToSpeech.php
joshuaabenazer: @dkotter I noticed that, but the default settings function has been declared as a `private` method on all the other classes, so it does not really call the method from the child class.
I will need to make changes to the other classes too. I don't think the default settings method needs to be `private` and can be declared as `public`, but wanted to check with you regarding that. If there are concerns then we can declare them as `protected` for this to work as expected.

dkotter: Yeah, I see no reason why this should be `private`. In looking at the history, this method was first added to the `ComputerVision` class here: https://github.com/10up/classifai/pull/264 and no reason I can see from that PR why this needs to be `private`. I believe this was then just copied/pasted to other classes. We should be fine to change this

dkotter:
@joshuaabenazer Now that #549 has been merged, do you still want to proceed on this one? I know we discussed we may or may not want this one once the larger refactoring was done. If this is still valid, there are some merge conflicts that need cleaned up now.

joshuaabenazer:
@dkotter @jeffpaul with the other [PR](https://github.com/10up/classifai/pull/549) this does get satisfied code-wise, and I feel since we have two different states of toggle for audio generation this would be incomplete without providing an option for the other state too. I would vote for skipping this for now.

dkotter:
@joshuaabenazer Sounds good. I'm going to close this one out but feel free to re-open in the future if desired

PR #543, opened 2023-07-26
Update to the most recent version of the IBM Watson NLU API

dkotter:
### Description of the Change

We've been running the `2018-03-19` version of the IBM Watson NLU API for a number of years. According to their [documentation](https://cloud.ibm.com/docs/natural-language-understanding?topic=natural-language-understanding-release-notes#active-version-dates), the latest version is `2022-08-10`.

Ideally we should be running the latest versions of the various APIs we integrate with (where possible) so this PR updates the API version we make requests to. As far as I can tell in my testing, things still work as expected, no breaking changes were found. The actual results returned are different but that is expected with running a newer API version.

### How to test the Change

1. Ensure you have IBM Watson setup and configured properly
2. Ensure classification works with all four features
3. If desired, downgrade the API version back to `2018-03-19` and compare results to see what differences there are

### Changelog Entry

> Changed - Update the IBM Watson NLU API to the `2022-08-10` version.

### Credits

Props @dkotter

### Checklist:

### DIFF
--- a/config.php
+++ b/config.php
@@ -16,8 +16,9 @@ classifai_define( 'CLASSIFAI_PLUGIN_VERSION', $plugin_version );
 classifai_define( 'CLASSIFAI_PLUGIN_DIR', __DIR__ );
 classifai_define( 'CLASSIFAI_PLUGIN_URL', plugin_dir_url( __FILE__ ) );
 
-// API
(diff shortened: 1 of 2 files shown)


ravinderk (approved):
I agree with @dkotter, I also did not notice any difference. Between results from different IBM Watson API

PR #544, opened 2023-07-27
Update our excerpt prompt

dkotter:
### Description of the Change

We recently received some feedback about our excerpt generation feature. While it seems to provide a decent summary of the content, excerpts are often meant to be used alongside the title of the content. For instance, a common design pattern is to display a featured image with the title below and the excerpt below that. In this case, ideally the excerpt pairs well with this title and doesn't just repeat the same thing.

In testing, often the generated excerpts include the title, sometime verbatim, sometimes with a word or two changed. The generated excerpt will contain more than that but often it will start with the same title. In the above mentioned scenario, this isn't ideal as you end up with the same (or close to the same) text shown back to back (first in the title, then in the excerpt).

This PR does a few things to try and help with this:

1. We move our prompt into a `system` message instead of a `user` message. From OpenAI's docs:

> The system message helps set the behavior of the assistant. ... However note that the system message is optional and the model’s behavior without a system message is likely to be similar to using a generic message such as "You are a helpful assistant."

Not sure how much this helps but feels like best practice to set our prompt in the `system` message and the actual content we want summarized into a separate `user` message.

2. We've modified the prompt some and have included a bit in there that asks it to provide a summary that pairs well with the title:
```
Summarize the following message using a maximum of %d words. Ensure this summary pairs well with the following text: %s.
```

Not sure if that last part actually helps or not. Seems like the next change is the one that helped the most.

3. We've removed the title from the content we are summarizing. Previously we took the title and the post content, combined those together, and sent that along to OpenAI to summarize. Now we just send the post content.
4. We allow the title to be passed into our endpoint so we have the most up-to-date title (instead of pulling this from the database)

### How to test the Change

1. Setup the OpenAI ChatGPT functionality, in particular the Generate excerpt functionality
2. Generate an excerpt for a piece of content and ensure text is returned
3. This gets more into opinion territory but use your best judgement to determine if the generated excerpt seems to pair well with the title. Can also run these same tests with the current released plugin to see how the generated excerpts differ

### Changelog Entry

> Changed - Update the prompt we send to OpenAI that is used to generate excerpts to try and ensure the excerpts generated pair well with the title of the content.

### Credits

Props @dkotter

### Checklist:

### DIFF
--- a/includes/Classifai/Providers/OpenAI/ChatGPT.php
+++ b/includes/Classifai/Providers/OpenAI/ChatGPT.php
@@ -558,7 +558,7 @@ class ChatGPT extends Provider {
 		 *
 		 * @return {string} Prompt.
 		 */
-		$prompt = apply_filters( 'classifai_chatgpt_excerpt_prompt', sprintf( 'Summarize the following message using a maximum of %d words. Ensure this summary pairs well with the following text: %s.', $excerpt_length, get_the_title( $post_id ) ), $post_id, $excerpt_length );
+		$prompt = apply_filters( 'classifai_chatgpt_excerpt_prompt', sprintf( 'Summarize the following message using a maximum of %d words. Ensure this summary pairs well with the following text: %s.', $excerpt_length, $args['title'] ), $post_id, $excerpt_length );
 
 		/**
 		 * Filter the request body before sending to ChatGPT.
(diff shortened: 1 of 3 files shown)


Sidsector9 (approved):
I tested this with a few different samples, the responses are very different compared to `develop` and did not include the title in a "duplicated" sort of way.

Although it works for the sample set, I think we will only truly know how well this works from customer feedback.

PR #549, opened 2023-08-01
#540 - Audio Generation Cleanup / Optimization / Display

### LINKED ISSUE #540: TTS: Provide option to disable auto re-generation of audio upon post content update.

Currently when the content of a post is updated even though it may be correcting punctuations ( or spelling corrections or anything other minor corrections for that matter ), if an audio version of the post already exists, it will re-generate the audio based on the updated content.

Trying to be cost effective with the service being used here, it would be great if there was a setting to disable this auto regeneration of the audio, and if so allow the editors of the post to explicitly click a button to regenerate audio when they feel it makes sense.

### Designs

Screenshot 2023-07-18 at 12 11 38 AM
Screenshot 2023-07-18 at 12 12 40 AM

_No response_

### Code of Conduct

joshuaabenazer:
### Description of the Change

Closes #540 

### How to test the Change

We need to test the entire audio generation / regeneration process along with it's ability to hide/display the audio on the FE using the new display controls.
There are two new filters that allows us to manipulate the initial state of audio generation and the subsequent state of audio generation once it has already been generated. We need to manipulate the filters and review the process to see if it works as expected.

### Changelog Entry

> Added - Option to hide/unhide audio controls on the FE if the audio exists.
> Changed - The audio generation process.

### Credits

Props @joshuaabenazer 

### Checklist:

### DIFF
--- a/includes/Classifai/Helpers.php
+++ b/includes/Classifai/Helpers.php
@@ -279,7 +279,7 @@ function get_supported_post_types() {
 /**
  * The list of post types that TTS supports.
  *
- * return array Supported Post Types.
+ * @return array Supported Post Types.
  */
 function get_tts_supported_post_types() {
 	$classifai_settings = get_plugin_settings( 'language_processing', Azure\TextToSpeech::FEATURE_NAME );
--- a/includes/Classifai/Helpers.php
+++ b/includes/Classifai/Helpers.php
@@ -279,7 +279,7 @@ function get_supported_post_types() {
 /**
  * The list of post types that TTS supports.
  *
- * return {array} Supported Post Types.
+ * return array Supported Post Types.
  */
 function get_tts_supported_post_types() {
 	$classifai_settings = get_plugin_settings( 'language_processing', Azure\TextToSpeech::FEATURE_NAME );
--- a/includes/Classifai/Helpers.php
+++ b/includes/Classifai/Helpers.php
@@ -3,6 +3,7 @@
 namespace Classifai;
 
 use Classifai\Providers\Provider;
+use Classifai\Providers\Azure;
 use Classifai\Services\Service;
 use Classifai\Services\ServicesManager;
 use WP_Error;
@@ -271,6 +272,28 @@ function get_supported_post_types() {
 	 * @return {array} Array of post types.
 	 */
 	$post_types = apply_filters( 'classifai_post_types', $post_types );
+
+	return $post_types;
+}
+
+/**
+ * The list of post types that TTS supports.
+ *
+ * return {array} Supported Post Types.
+ */
+function get_tts_supported_post_types() {
+	$classifai_settings = get_plugin_settings( 'language_processing', Azure\TextToSpeech::FEATURE_NAME );
+
+	if ( empty( $classifai_settings ) ) {
+		$post_types = [];
+	} else {
+		$post_types = [];
+		foreach ( $classifai_settings['post_types'] as $post_type => $enabled ) {
+			if ( ! empty( $enabled ) ) {
+				$post_types[] = $post_type;
+			}
+		}
+	}
 
 	return $post_types;
 }
--- a/includes/Classifai/Providers/Azure/TextToSpeech.php
+++ b/includes/Classifai/Providers/Azure/TextToSpeech.php
@@ -239,7 +239,7 @@ class TextToSpeech extends Provider {
 				if ( ! empty( $current_settings['voices'] ) ) {
 					$current_settings['authenticated'] = true;
 				} else {
-					$current_settings['voices'] = [];
+					$current_settings['voices']        = [];
 					$current_settings['authenticated'] = false;
 				}
 			}
@@ -268,8 +268,6 @@ class TextToSpeech extends Provider {
 
 		if ( isset( $settings['voice'] ) && ! empty( $settings['voice'] ) ) {
 			$current_settings['voice'] = sanitize_text_field( $settings['voice'] );
-		} else {
-			$current_settings['voice'] = '';
 		}
 
 		return $current_settings;
@@ -452,6 +450,17 @@ class TextToSpeech extends Provider {
 	 * @return {bool}                     The initial state of audio generation. Default true.
 	 */
 	public function get_audio_generation_initial_state( $post = null ) {
+		/**
+		 * Initial state of the audio generation toggle when no audio already exists for the post.
+		 *
+		 * @since 2.3.0
+		 * @hook classifai_audio_generation_initial_state
+		 *
+		 * @param  {bool}    $state Initial state of audio generation toggle on a post. Default true.
+		 * @param  {WP_Post} $post  The current Post object.
+		 *
+		 * @return {bool}           Initial state the audio generation toggle should be set to when no audio exists.
+		 */
 		return apply_filters( 'classifai_audio_generation_initial_state', true, get_post( $post ) );
 	}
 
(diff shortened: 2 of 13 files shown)


includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -436,44 +437,86 @@ private function get_default_settings() {
...
+	 *
+	 * Fetch the subsequent state of audio generation once audio is generated for the post.
+	 *
+	 * @param {int|WP_Post|null} $post   Optional. Post ID or post object. `null`, `false`, `0` and other PHP falsey values
+	 *                                   return the current global post inside the loop. A numerically valid post ID that
+	 *                                   points to a non-existent post returns `null`. Defaults to global $post.
+	 * @return {bool}                    The subsequenet state of audio generation. Default false.
+	 */
+	public function get_audio_generation_subsequent_state( $post = null ) {
+		return apply_filters( 'classifai_audio_generation_subsequent_state', false, get_post( $post ) );
```
dkotter, line 489: For these filters, can we add filter docblocks so these make it to our automated documentation? See https://github.com/10up/classifai/blob/develop/includes/Classifai/Providers/Azure/TextToSpeech.php#L629 as an example

dkotter:
Left one minor comment but things look really good here!

One thing I noticed while testing, and maybe this is expected, but if I create a new piece of content, the `Enable audio generation` toggle is turned on (which is expected). If I toggle that off and then save the item (either save as draft or publish), the toggle immediately goes back to being on. It does seem to respect the off value though and no audio is generated, but I guess I'd expect the toggle to stay off if I manually toggled it off. It seems to only stay off after audio has been generated.

includes/Classifai/Providers/Azure/TextToSpeech.php
joshuaabenazer: Added the doc blocks. But let me know if the wording needs to be updated for it.

joshuaabenazer:
> Left one minor comment but things look really good here!
> 
> One thing I noticed while testing, and maybe this is expected, but if I create a new piece of content, the `Enable audio generation` toggle is turned on (which is expected). If I toggle that off and then save the item (either save as draft or publish), the toggle immediately goes back to being on. It does seem to respect the off value though and no audio is generated, but I guess I'd expect the toggle to stay off if I manually toggled it off. It seems to only stay off after audio has been generated.

@dkotter I see what you are saying there. Currently the state of the toggle switches back and forth depending on what the initial and subsequent states are set to relative to if an audio is generated ( subsequent ) or not ( initial ). Again can be changed via the filters too. Wondering if we should keep the initial state to default to false instead maybe?

I would hope that the audio generation takes place only on a publish since a draft post is still incomplete. But it could also make things complicated since we are not storing this data about audio generation, and it is just used as an indication for audio generation on the fly.

We should be able to control the generation for drafts in the [rest handler](https://github.com/10up/classifai/pull/549/files#diff-92e53023910a60a4596efdc9ac39392ee04c4920283229fd8050a78170a4b4c0R567) -  and similarly for the [classic editor](https://github.com/10up/classifai/pull/549/files#diff-92e53023910a60a4596efdc9ac39392ee04c4920283229fd8050a78170a4b4c0R695) but then am not sure if that would be intuitive or not. The wording for the toggle does say publish or update and probably we can then improve the description there to be more clear? We should probably get some more opinions on what users expect during these different usecases and then take a call.

includes/Classifai/Helpers.php
```diff
@@ -275,6 +276,28 @@ function get_supported_post_types() {
 	return $post_types;
 }
 
+/**
+ * The list of post types that TTS supports.
+ *
+ * return {array} Supported Post Types.
```
dkotter:
```suggestion
 * return array Supported Post Types.
```

includes/Classifai/Providers/Provider.php
```diff
@@ -164,6 +164,13 @@ public function get_settings( $index = false ) {
 		return $settings;
 	}
 
+	/**
+	 * Returns the default settings.
```
dkotter, line 168:
```suggestion
	 * Returns the default settings.
	 *
	 * @return array
```

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -436,44 +435,108 @@ private function get_default_settings() {
...
 		];
 	}
 
 	/**
-	 * Add `classifai_synthesize_speech` to rest API for view/edit.
+	 * Initial audio generation state.
+	 *
+	 * Fetch the initial state of audio generation prior to the audio existing for the post.
+	 *
+	 * @param  {int|WP_Post|null} $post   Optional. Post ID or post object. `null`, `false`, `0` and other PHP falsey values
```
```suggestion
	 * @param  int|WP_Post|null $post   Optional. Post ID or post object. `null`, `false`, `0` and other PHP falsey values
```

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -436,44 +435,108 @@ private function get_default_settings() {
...
+	 *                                    return the current global post inside the loop. A numerically valid post ID that
+	 *                                    points to a non-existent post returns `null`. Defaults to global $post.
+	 * @return {bool}                     The initial state of audio generation. Default true.
```
```suggestion
	 * @return bool                     The initial state of audio generation. Default true.
```

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -436,44 +435,108 @@ private function get_default_settings() {
...
+		 */
+		return apply_filters( 'classifai_audio_generation_initial_state', true, get_post( $post ) );
+	}
+
+	/**
+	 * Subsequent audio generation state.
+	 *
+	 * Fetch the subsequent state of audio generation once audio is generated for the post.
+	 *
+	 * @param {int|WP_Post|null} $post   Optional. Post ID or post object. `null`, `false`, `0` and other PHP falsey values
```
```suggestion
	 * @param int|WP_Post|null $post   Optional. Post ID or post object. `null`, `false`, `0` and other PHP falsey values
```

includes/Classifai/Providers/Azure/TextToSpeech.php
```diff
@@ -436,44 +435,108 @@ private function get_default_settings() {
...
+	 *                                   return the current global post inside the loop. A numerically valid post ID that
+	 *                                   points to a non-existent post returns `null`. Defaults to global $post.
+	 * @return {bool}                    The subsequenet state of audio generation. Default false.
```
```suggestion
	 * @return bool                    The subsequent state of audio generation. Default false.
```

includes/Classifai/Helpers.php
```diff
@@ -275,6 +276,28 @@ function get_supported_post_types() {
 	return $post_types;
 }
 
+/**
+ * The list of post types that TTS supports.
+ *
+ * return array Supported Post Types.
```
dkotter:
```suggestion
 * @return array Supported Post Types.
```

dkotter (approved):

dkotter:
@joshuaabenazer 

> Currently the state of the toggle switches back and forth depending on what the initial and subsequent states are set to relative to if an audio is generated ( subsequent ) or not ( initial ). Again can be changed via the filters too. Wondering if we should keep the initial state to default to false instead maybe?

I'm not so much worried about how we handle drafts but does feel like we may get complaints from users that want to toggle off audio generation on a specific piece of content but at the moment, they'll have to remember to turn that toggle off anytime the content is saved.

But maybe that's something we address in a followup PR? Unless you want to handle that here before I merge this in?

joshuaabenazer:
@dkotter I like the idea of a followup PR once we have some insights from users about this. I think if there is no other feedback for this particular PR then we can get this merged in. Thanks!

PR #558, opened 2023-08-10
#557 - Fix - Prevent TTS Audio leakage into places using excerpts (like archives)

### LINKED ISSUE #557: The TTS Audio text and assets leak into wherever excerpt is used
### Describe the bug

**Issue Details**
If the archive pages that contains the TTS audio version of a post is reviewed, it seems like the label part of the audio markup "Listen to this post" makes its way to the excerpt if **_it is not explicitly set_**.
Apart from this with the way we are enqueuing the assets ( scripts / styles ) for the audio on the `the_content` hook, so the assets also make it's way to the archives when there doesn't seem to be a need of it - causing uneccessary assets being loaded on the archives.

All of this happens since we have this line https://github.com/WordPress/WordPress/blame/master/wp-includes/formatting.php#L3986 attempting to use the same hook on the excerpt.

**Proposed Fix**
On the `the_content` hook being used, add a conditional to check for if we are on the singular version of the current post type, if not then return the content as is without any modifications - ensuring this feature only loads or affects the singular version.

### Steps to Reproduce

1. Generate audio for a post that does not have an excerpt set explicitly.
2. Review it on the post single. Works as expected.
3. Review it on an archive containing the same post. Here will now notice that the "Listen to this post" is now part of the archive excerpt of the post.

Screenshot 2023-08-10 at 12 53 29 PM

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

joshuaabenazer:
### Description of the Change

Closes #557 

### How to test the Change

1. Generate audio for a post that does not have an excerpt set explicitly.
2. Review it on the post single. Works as expected with the audio showing at the start of the beginning of the post content.
3. Review it on an archive containing the same post. Should not have any remnants of the TTS Audio feature like the "Listen to this post" appended to the excerpt. Also shouldn't be loading the TTS audio js and CSS on the archives.

### Changelog Entry

> Fixed - Prevent TTS Audio leakage into places using excerpts (like archives)

### Credits

Props @joshuaabenazer 

### Checklist:

### DIFF
--- a/includes/Classifai/Providers/Azure/TextToSpeech.php
+++ b/includes/Classifai/Providers/Azure/TextToSpeech.php
@@ -732,6 +732,10 @@ class TextToSpeech extends Provider {
 			return $content;
 		}
 
+		if ( ! is_singular( $_post->post_type ) ) {
+			return $content;
+		}
+
 		if ( ! in_array( $_post->post_type, get_tts_supported_post_types(), true ) ) {
 			return $content;
 		}


jeffpaul:
@Sidsector9 as you worked on the original feature, can you help check the fix here?

dkotter (approved):

PR #559, opened 2023-08-10
Update all Azure APIs to their latest public version

### LINKED ISSUE #553: Update the Azure APIs to their latest versions

Ideally we should be looking to update any APIs we use to their latest versions on a regular basis. This issue is focused on any Azure APIs we use. The following is a list of the APIs we are using and the version.

- Analyze Image v3.0
- OCR v3.2
- Read v3.2
- Generate Thumbnail v3.1
- Personalizer v1.0
- TTS `cognitiveservices/v1`

For the **Personalizer** API, v1.0 is the latest (though there is a v1.1 in preview) so nothing needed there. Same for our **Text to Speech** API, we are currently using the latest version.

The **Analyze Image**, **OCR**, **Read** and **Generate Thumbnail** APIs are all under the same service (previously known as Cognitive Services Computer Vision, since renamed to Azure AI Vision). The latest released version of this API is v3.2, while there is a v4.0 public preview API.

Azure is pushing for everyone to use the new v4.0 public preview API but in researching this, there are currently some limitations that may hold us back. For instance, generating image captions or smart cropping are [only available](https://learn.microsoft.com/en-us/azure/ai-services/computer-vision/overview-image-analysis?tabs=4-0) in a small set of regions in v4.0 (East US, France Central, Korea Central, North Europe, Southeast Asia, West Europe, and West US, East Asia).

There's also been quite a few changes to these APIs in v4.0, so will take some refactoring if we pursue these updates. For instance, all existing features we use, outside of reading content from PDFs, is now under a single Analyze API in v4.0. This will require some changes to how our code works to account for this.

That said, assuming we're okay with the region limitations, I'd like to pursue updating all of those to v4.0. If we're not okay with that, I think it would be ideal to get all of those on v3.2 (so just Analyze Image and Generate Thumbnail).

I tried updating to v3.2 of the Analyze Image API and while the results we get seem good, the confidence scores, at least for image captions, are lower, so that's something we would need to determine how best to handle (in using the Vision Studio tool, this seems to have been fixed in v4.0). Their docs even mention:

> In general, we advise a confidence threshold of 0.4 for the Image Analysis 3.2 API and of 0.0 for the Image Analysis 4.0 API (preview).

If we decide to update to v4.0, here's tasks as I see them:

- [ ] Update Analyze Image API to v4.0 and address any issues there. I believe we'll need to update how we send data and how we parse the received response
- [ ] Update how we handle [OCR](https://learn.microsoft.com/en-us/azure/ai-services/computer-vision/overview-ocr) to use this new API
- [ ] Update how we handle [generating thumbnails](https://learn.microsoft.com/en-us/azure/ai-services/computer-vision/concept-generate-thumbnails-40) to use this new API
- [ ] Investigate the Read API. It seems like this functionality moved to a new API ([Document Intelligence](https://learn.microsoft.com/en-us/azure/ai-services/document-intelligence/concept-read?view=doc-intel-3.1.0)). We should investigate what it would mean to use that API instead. We may find it's not worth the effort and we leave this on the current v3.2 API

If we stick with v3.2, here's what we'll want to do:

- [ ] Update the Analyze Image API to v3.2 and modify how we handle error responses (this changed in v3.2).
- [ ] Update how we deal with confidence scores to account for lower scores in v3.2
- [ ] Update the Generate Thumbnail API to v3.2 and address any issues there

### Designs

_No response_

_No response_

### Code of Conduct

kmgalanakis:
### Description of the Change

In this MR, I'm updating the Azure APIs version. More specifically I'm updating the Azure AI Vision Image Analyze endpoint version and Thumbnail Generate version to v3.2. I'm also handling the way the API error responses are handed since those changed in v3.2

Closes https://github.com/10up/classifai/issues/553

### Changelog Entry

> Changed - Updated the Azure AI Vision Image Analyze endpoint version and Thumbnail Generate version to v3.2. Note we recommend lowering the threshold values for the Descriptive Text Generator feature to 55% to ensure the best results. 

### Credits

Props @kmgalanakis 

### Checklist:


includes/Classifai/Providers/Azure/ComputerVision.php
```diff
@@ -1323,6 +1323,8 @@ public function rest_endpoint_callback( $post_id, $route_to_call, $args = [] ) {
 
 		$image_scan_results = $this->scan_image( $image_url, [ $route_to_call ] );
 
+		print_r( $image_scan_results );
```
dkotter: I'm assuming this was here for testing purposes and should be removed

dkotter (changes requested):

dkotter:
I tested this a bit when I put the issue together and one thing I noticed that we'll need to address in this PR is that the error's returned from the API have changed from v3.0 to v3.2. Can compare the [v3.0](https://westcentralus.dev.cognitive.microsoft.com/docs/services/computer-vision-v3-ga/operations/56f91f2e778daf14a499f21b) error responses to the [v3.2](https://westus.dev.cognitive.microsoft.com/docs/services/computer-vision-v3-2/operations/56f91f2e778daf14a499f21b) error responses to see that.

If helpful, here's the code I had added locally to handle this:
```php
if ( 200 !== wp_remote_retrieve_response_code( $request ) ) {
    if ( isset( $body->error ) ) {
         $rtn = new WP_Error( $body->error->code ?? 'error', $body->error->message ?? esc_html__( 'An error occurred.', 'classifai' ), $body );
     } elseif ( isset( $body->message ) ) {
         $rtn = new WP_Error( $body->error ?? 'error', $body->message, $body );
     } else {
         $rtn = new WP_Error( 'error', esc_html__( 'An error occurred.', 'classifai' ), $body );
     }
} else {
    $rtn = $body;
}
```

includes/Classifai/Providers/Azure/ComputerVision.php
kmgalanakis: Sorry for that, I don't usually forget leftover debugging code. Fixed.

kmgalanakis:
> I tested this a bit when I put the issue together and one thing I noticed that we'll need to address in this PR is that the error's returned from the API have changed from v3.0 to v3.2. Can compare the [v3.0](https://westcentralus.dev.cognitive.microsoft.com/docs/services/computer-vision-v3-ga/operations/56f91f2e778daf14a499f21b) error responses to the [v3.2](https://westus.dev.cognitive.microsoft.com/docs/services/computer-vision-v3-2/operations/56f91f2e778daf14a499f21b) error responses to see that.
> 
> If helpful, here's the code I had added locally to handle this:
> 
> ```
> if ( 200 !== wp_remote_retrieve_response_code( $request ) ) {
>     if ( isset( $body->error ) ) {
>          $rtn = new WP_Error( $body->error->code ?? 'error', $body->error->message ?? esc_html__( 'An error occurred.', 'classifai' ), $body );
>      } elseif ( isset( $body->message ) ) {
>          $rtn = new WP_Error( $body->error ?? 'error', $body->message, $body );
>      } else {
>          $rtn = new WP_Error( 'error', esc_html__( 'An error occurred.', 'classifai' ), $body );
>      }
> } else {
>     $rtn = $body;
> }
> ```

Your code was mostly fine @dkotter. In the code for the second `if condition`, I just had to change `$body->error` to `$body->code`. For `Response 415` the error format remained 

```
{
    "code": "BadArgument",
    "message": "Invalid Media Type"
}
```

Thank you

dkotter:
Overall this looks good to me but I do think we need to figure out how to handle the lower confidence scores, as mentioned [here](https://github.com/10up/classifai/issues/553#issuecomment-1673262900). That will need to be resolved as part of this PR before we can merge this in.

jeffpaul:
@dkotter assuming the TLDR from @kmgalanakis is that with the v3.2 API in order to see similar results the threshold needs to drop from 70% to 50/55%, then I think the suggestion from @kmgalanakis to display a dismissable admin notice when we see a threshold still at the default 70% and IBM Watson NLU active that we recommend dropping the value(s) to 55%.  So something like:
> The latest version of ClassifAI comes with an update to the Language Processing option with IBM Watson that provides best results if you change the default threshold from 70% to 55%.  [Click here to adjust those settings](#anchor-link-to-IBM-Watson-NLU-Settings).

If the IBM Watson NLU feature is not activated and the thresholds are at 70%, then let's change those defaults to 55%.  Otherwise new installs will receive the new default of 55%.

jeffpaul:
@dkotter any thoughts on the potential handling as noted in my comment above?

dkotter:
> @dkotter assuming the TLDR from @kmgalanakis is that with the v3.2 API in order to see similar results the threshold needs to drop from 70% to 50/55%, then I think the suggestion from @kmgalanakis to display a dismissable admin notice when we see a threshold still at the default 70% and IBM Watson NLU active that we recommend dropping the value(s) to 55%. So something like:
> 
> > The latest version of ClassifAI comes with an update to the Language Processing option with IBM Watson that provides best results if you change the default threshold from 70% to 55%.  [Click here to adjust those settings](#anchor-link-to-IBM-Watson-NLU-Settings).
> 
> If the IBM Watson NLU feature is not activated and the thresholds are at 70%, then let's change those defaults to 55%. Otherwise new installs will receive the new default of 55%.

I think this all sounds fine. We can call this out in the changelog and in the release as well as add an admin notice to hopefully make this clear to all existing users.

sksaju:
> > @dkotter assuming the TLDR from @kmgalanakis is that with the v3.2 API in order to see similar results the threshold needs to drop from 70% to 50/55%, then I think the suggestion from @kmgalanakis to display a dismissable admin notice when we see a threshold still at the default 70% and IBM Watson NLU active that we recommend dropping the value(s) to 55%. So something like:
> > > The latest version of ClassifAI comes with an update to the Language Processing option with IBM Watson that provides best results if you change the default threshold from 70% to 55%.  [Click here to adjust those settings](#anchor-link-to-IBM-Watson-NLU-Settings).
> > 
> > 
> > If the IBM Watson NLU feature is not activated and the thresholds are at 70%, then let's change those defaults to 55%. Otherwise new installs will receive the new default of 55%.
> 
> I think this all sounds fine. We can call this out in the changelog and in the release as well as add an admin notice to hopefully make this clear to all existing users.

@jeffpaul @dkotter I've updated the default thresholds value and added admin notice which will only run when the IBM Watson is active/configured. please check it and let me know your feedback. Thank you

includes/Classifai/Providers/Watson/NLU.php
```diff
@@ -1048,6 +1049,42 @@ public function add_process_content_meta_to_rest_api() {
...
+		$thresholds = [
+			$config['features']['category_threshold'] ?? 55,
+			$config['features']['keyword_threshold'] ?? 55,
+			$config['features']['entity_threshold'] ?? 55,
+			$config['features']['concept_threshold'] ?? 55,
+		];
+
+		if ( ! in_array( 70, $thresholds, true ) ) {
+			return;
+		}
```
dkotter: This doesn't work because `is_configured` will return a `boolean`, indicating if the feature is enabled or not. But we take that value and try and get the threshold settings from it, which won't work. We need to check if the feature is configured and then get the settings for the feature to look at

includes/Classifai/Providers/Watson/NLU.php
```diff
@@ -1048,6 +1049,42 @@ public function add_process_content_meta_to_rest_api() {
...
+		$config = $this->is_configured();
+
+		$thresholds = [
+			$config['features']['category_threshold'] ?? 55,
+			$config['features']['keyword_threshold'] ?? 55,
+			$config['features']['entity_threshold'] ?? 55,
+			$config['features']['concept_threshold'] ?? 55,
+		];
+
+		if ( ! in_array( 70, $thresholds, true ) ) {
```
dkotter: In addition to my other note, there's a few issues with showing the message based on this condition:

- If someone has changed the threshold setting to something other than 70 (like 75 or 65), the message won't show
- If in the future they manually change the setting back to 70, the admin notice will always show for them

I think we need to consider an approach where we show a one-time admin notice the next time someone updates the plugin. We'll need to store something in an option to track that we showed that notice and not display it again

dkotter (changes requested):

includes/Classifai/Providers/Watson/NLU.php
```diff
@@ -1048,6 +1049,42 @@ public function add_process_content_meta_to_rest_api() {
 		);
 	}
 
+	/**
+	 * Display a dismissable admin notice when a threshold still at the default 70% and IBM Watson NLU active
+	 */
+	public function thresholds_update_notice() {
```
dkotter: Ah, in thinking about this more, I don't think this is the right thing to update. This is the threshold for the IBM Watson NLU feature but we are updating the API used by Azure image processing features, which are different things. I think any code related to NLU in this PR can be reverted

dkotter (changes requested):

dkotter (approved):

dkotter:
I've gotten this PR ready for review. I think it would be a great addition to the v3 release as the v3.2 version of the AI Vision API brings support for quite a few new languages which I know has come up as a request.

I know one of the things we were concerned with was lower confidence scores. In testing, I do see that for Descriptive Text. As an example, I tested on v3.0 with an image of a dog and got a response of `a close up of a dog` with 96% confidence. On v3.2, I got `a dog with its mouth open` with 57% confidence. So I think the results are better (or at least just as good) but the confidence scores have decreased. As such, I've changed our default on that from 75% to 55% and we output an admin notice that lets people know they should decrease that threshold.

But for Image Tagging, I'm not seeing a decrease in the confidence scores. Results seem better but the actual scores seem to be the same, so I've left that threshold alone.

iamdharmesh (approved):
Code LGTM and PR tests well. Thanks all for the great work here.

PR #560, opened 2023-08-10
Update references from Computer Vision to Azure AI Vision

### LINKED ISSUE #554: Update references from Computer Vision to Azure AI Vision

It appears that Azure has updated what used to be know as Cognitive Services Computer Vision to just [Azure AI Vision](https://learn.microsoft.com/en-us/azure/ai-services/computer-vision/overview).

We should look for any user-facing use of the term Computer Vision (or Cognitive Service) and update those to Azure AI Vision. I think there's only one or two of these in the settings page so this should be an easy task.

I don't think we need to update any code (like function names, code comments, etc) just anything that is user-facing.

### Designs

_No response_

_No response_

### Code of Conduct

kmgalanakis:
### Description of the Change

In this PR I'm updatin the references of the renamed Computer Vision API to Azure AI Vision.

Closes https://github.com/10up/classifai/issues/554

### Changelog Entry

> Changed - Update the references of the renamed Computer Vision API to Azure AI Vision.

### Credits

Props @kmgalanakis

### Checklist:

### DIFF
--- a/includes/Classifai/Providers/Azure/Personalizer.php
+++ b/includes/Classifai/Providers/Azure/Personalizer.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * Azure Personalizer
+ * Azure AI Personalizer
  */
 
 namespace Classifai\Providers\Azure;
@@ -42,7 +42,7 @@ class Personalizer extends Provider {
 
 		// Set the onboarding options.
 		$this->onboarding_options = array(
-			'title'    => __( 'Microsoft Azure Personalizer', 'classifai' ),
+			'title'    => __( 'Microsoft Azure AI Personalizer', 'classifai' ),
 			'fields'   => array( 'url', 'api-key' ),
 			'features' => array(
 				'authenticated' => __( 'Recommended content block', 'classifai' ),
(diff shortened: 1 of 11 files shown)


includes/Classifai/Providers/Azure/Personalizer.php
```diff
@@ -97,7 +97,7 @@ public function setup_fields_sections() {
 				'description'   => sprintf(
 					wp_kses(
 						// translators: 1 - link to create a Personalizer resource.
-						__( 'Azure Cognitive Service Personalizer Endpoint, <a href="%1$s" target="_blank">create a Personalizer resource</a> in the Azure portal to get your key and endpoint.', 'classifai' ),
+						__( 'Azure AI Vision Personalizer Endpoint, <a href="%1$s" target="_blank">create a Personalizer resource</a> in the Azure portal to get your key and endpoint.', 'classifai' ),
```
dkotter: Looks like Personalizer has been renamed to [Azure AI Personalizer](https://learn.microsoft.com/en-us/azure/ai-services/personalizer/what-is-personalizer), not AI Vision.

dkotter (changes requested):

includes/Classifai/Providers/Azure/Personalizer.php
kmgalanakis: @dkotter I've added another commit that properly handles the updating of all the Personalizer's references to Azure AI Personalizer.

Could you please have another look?

dkotter (approved):

PR #562, opened 2023-08-10
Fix failing E2E tests and add back testing on `trunk`

dkotter:
### Description of the Change

E2E tests have been failing on WP `trunk` for a few months (and thus we had removed that from our test matrix) and now tests are failing on the latest release. In digging in, it seems the issue is that in WordPress 6.3, the post editor is loaded in an iframe if all registered blocks support the Block API version 3 or greater. It seems that Cypress doesn't like this iframed approach and the editor fails to load. This means any tests that need to interact with the block editor (like creating a new post) won't work.

I don't think this is an issue on the WordPress side, I think this is a Cypress specific problem. I was unable to figure out a true fix though and so this PR addresses this by ensuring our custom Recommended Content block is loaded for all tests. This block supports the Block API v2, so the editor ends up being loaded as before (not in an iframe) and tests work. This isn't a true fix and probably won't work forever but I figure having passing tests is worth it for now.

### How to test the Change

Ensure E2E tests are passing

### Changelog Entry

> Fixed - Ensure our E2E tests work properly on WordPress 6.3

### Credits

Props @dkotter

### Checklist:

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -17,6 +17,7 @@ jobs:
         core:
           - {name: 'WP latest', version: 'latest'}
           - {name: 'WP minimum', version: 'WordPress/WordPress#5.7'}
+          - {name: 'WP trunk', version: 'WordPress/WordPress#master'}
     steps:
     - name: Checkout
       uses: actions/checkout@v3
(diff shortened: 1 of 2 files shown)


iamdharmesh (approved):
Thank you for fixing this @dkotter. LGTM 🚀

PR #569, opened 2023-08-18
fix/561: delay image generation when using the Media Inserter

### LINKED ISSUE #561: Look to delay image generation when using the Media Inserter

In #535, we added an integration into the Media Inserter that allows you to generate and insert images straight from that inserter. At the moment, the [code](https://github.com/WordPress/gutenberg/pull/51542) that added the ability to interact with this Media Inserter is only in the Gutenberg plugin (assuming this will end up in WordPress 6.4) and we only have the ability to define the fetch function.

What this means is that as someone types in their image prompt, Gutenberg itself will slightly debounce this request (currently at [250ms](https://github.com/WordPress/gutenberg/blob/trunk/packages/block-editor/src/components/inserter/hooks/use-debounced-input.js)) before calling our custom fetch function. In reality, the debounce here is so small that as someone types in their image prompt, almost certainly we'll end up hitting the OpenAI API multiple times (and as such, be charged for each of those requests).

I [had tried](https://github.com/10up/classifai/pull/535#discussion_r1282280289) adding our own debouncing around this fetch function but I couldn't get it to work, I think because this function is already being debounced by Gutenberg itself. I see a few possibilities we could pursue here:

1. Continue investigating if we can add our own debouncing here, setting that to something higher to increase the likelihood that we only send a request once a user is truly done typing
2. Open an issue on the Gutenberg side to see if we could get an option added where you could choose between having search triggered on type or search triggered on a button submit, which the latter would solve our needs
3. Open an issue on the Gutenberg side to see if we can make the current debounce value filterable so we can increase that. Won't definitively solve the problem but we can try increasing that value high enough that it should solve it for most

Not sure how receptive the Gutenberg team will be to changes with this component but I do think either option would be useful for anyone that wants to integrate a 3rd party service with this Media Inserter, so not something that just helps ClassifAI.

### Designs

_No response_

_No response_

### Code of Conduct

Sidsector9:
Closes #561

### Description of the Change

Adds a modified version of debounce to add delay to an already debounced callback.

### How to test the Change

1. Pull the fix branch
2. To really see the difference, increase the delay from `700` to `2000` in this line `fetch: debounce( imageFetcher, 700 ),`
3. In the media inserter, type very slowly and observe that there are no multiple AJAX calls

### Changelog Entry

> Fixed - Delay image generation when using the Media Inserter

### Credits

Props @dkotter @Sidsector9 

### Checklist:


src/js/gutenberg-plugins/inserter-media-category.js
```diff
@@ -30,38 +30,65 @@ waitFor( isInserterOpened ).then( () =>
...
-					id: undefined,
-					alt: search,
-					caption: classifaiDalleData.caption,
-				} ) )
-			)
-			.catch( () => [] );
-
-		return images;
-	},
+	fetch: debounce( imageFetcher, 700 ),
```
dkotter: I think we need to increase this delay. In my testing, I was still seeing multiple ajax requests at this point. If I increase this up to 2000 or 3000, I still occasionally saw two ajax requests but often only one and never more than two

Sidsector9:
@dkotter I performed a programmatic test by simulating human input every "x" seconds.

## Case 1: Debounce delay: 1000ms, typing interval: 900. Result - No ajax calls


https://github.com/10up/classifai/assets/17757960/b3e21228-bf92-40d2-967c-8aa17e4e4906


## Case 2: Debounce delay: 1000ms, typing interval: 1100. Result - Ajax calls for every input change

https://github.com/10up/classifai/assets/17757960/b1fd2ec9-a297-416c-b5f1-73c22099ff2b

src/js/gutenberg-plugins/inserter-media-category.js
Sidsector9: @dkotter I performed a programmatic test by simulating human input every "x" seconds.

## Case 1: Debounce delay: 1000ms, typing interval: 900. Result - No ajax calls


https://github.com/10up/classifai/assets/17757960/b3e21228-bf92-40d2-967c-8aa17e4e4906


## Case 2: Debounce delay: 1000ms, typing interval: 1100. Result - Ajax calls for every input change

https://github.com/10up/classifai/assets/17757960/b1fd2ec9-a297-416c-b5f1-73c22099ff2b

Just wanted to confirm if we should be increasing the debounce delay between 2000 - 3000ms?

dkotter: So here's what I get right now:

https://github.com/10up/classifai/assets/916738/8aa12b8d-42fc-4639-a427-75fbc889cbac

I enter a fairly simple image prompt with a fairly normal typing speed and I end up getting 7 requests before I actually finish (which at 2 cents per image and generating 3 images per request, that ends up costing 42 cents).

If I change the debounce all the way up to 3000, I get only a single request, though it takes a bit for that request to fire off:

https://github.com/10up/classifai/assets/916738/fb513f7d-c21d-4d34-a7b3-59fa0a4f701b

So unless I'm testing this wrong, I think we need to increase that debounce value somewhere between 2000 and 3000 to be more sure less requests will get sent.

Sidsector9: Got it, it's the pauses a user may take after typing few words that will trigger AJAX requests. I've increased the delay to 2500ms.

dkotter (approved):

PR #606, opened 2023-10-24
Add support for role/user based access

### LINKED ISSUE #588: Enabling features by Role/User

In order to roll out and granularly control specific features to specific roles and users, we might best recommend applying role- and user-based access to individual features, and rearchitecting the way access is controlled (at the code level) with custom defined capabilities for each feature (e.g. “classifai_can_generate_titles”), making even more granular control and role management available to developers and “pro” administrators.

For the interface, an example like the below would be added, wherever it's currently not available on each ClassifAI feature settings page, that would allow for enabling role-based access and subsequently which roles should have access to the specific feature:

new-role-access

An example like the below would also be added that would allow for enabling user-based access, a listing of specific users with access, and an option that would enable users to opt themselves out of a specific feature even if they are allowed access to it:

new-user-access

A simplified view of editing a specific user to enable/disable access to specific ClassifAI features could look something like this:

edit-user

A simplified view of a user editing their own profile to opt-out of specific ClassifAI features could look something like this:

new-user-profile

Specific to each Assistive AI feature, an in-context option can be presented to users to allow them to disable a feature without having to navigate to their profile (e.g., ClassifAI recommending tags for a post via a modal would have a action link opposite any primary/secondary action buttons to allow a user to disable the feature so they’re not presented with tag recommendations in the future).

Finally within this batch of work would be ensuring that a capability exists for each ClassifAI feature to support the ability to enable capability-based access for users.

From a high-level estimation perspective, this will likely be a slightly larger effort (“t-shirt sized” L), will require adding additional settings to the Settings page, adding settings to the User profile page, and updating each feature to conditionally run depending upon these new settings with a rough estimate of 2-3 weeks for completion.

### Designs

See in-context in the description above.

n/a

### Code of Conduct

iamdharmesh:
### Description of the Change
This PR adds support for role- and user-based access control to the ClassifiAI feature. It also includes functionality for user-based opt-out, allowing users to choose not to use the specific feature if they prefer not to or if they dislike it.

**Feature settings:**
image

**User Profile:**
image

**Edit User:**
image

**In-Context disable feature link to user profile**
|Title generation|Resize content|
|---|---|
|Screenshot 2023-10-30 at 5 17 15 PM|Screenshot 2023-10-30 at 5 18 06 PM|

_**Note:** E2E tests are not added for this new feature._

Closes #588

### How to test the Change
1. Enable role-based access to the feature.
1. Select roles in the allowed roles to provide access to that feature.
1. Verify that only the selected roles have access to that feature.
1. Enable user-based access to the feature.
1. Select the user in the allowed users to provide access to that feature.
1. Verify that the selected user has access to that feature.
1. Enable user-based opt-out for the feature.
1. Go to the user profile and verify that you see the checkbox for that feature to opt out of using it.
1. Check the opt-out checkbox and save.
1. Verify that access to that feature has been disabled.
1. Uncheck the opt-out checkbox to opt back in to the feature.
1. Verify that the user now has access to that feature.
1. Follow these steps for each ClassifAI feature.

### Credits
Props @jeffpaul @dkotter @iamdharmesh @Sidsector9 

### Checklist:


includes/Classifai/Helpers.php
```diff
@@ -772,3 +772,26 @@ function get_all_post_statuses() {
...
+	if ( ! function_exists( 'get_editable_roles' ) ) {
+		require_once ABSPATH . 'wp-admin/includes/user.php';
+	}
+	$editable_roles = get_editable_roles() ?? [];
+
+	return array(
+		$feature . '_role_based_access'  => 1,
+		$feature . '_roles'              => array_keys( $editable_roles ),
+		$feature . '_user_based_access'  => 'no',
+		$feature . '_user_based_opt_out' => 'no',
```
iamdharmesh, line 795: @jeffpaul @dkotter Should user-based opt-out be enabled by default?

includes/Classifai/Providers/Watson/NLU.php
```diff
@@ -88,63 +93,81 @@ public function __construct( $service ) {
...
-				'entity_taxonomy'    => WATSON_ENTITY_TAXONOMY,
-			],
-		];
-
+		$settings = $this->get_default_settings() ?? [];
 		update_option( $this->get_option_name(), $settings );
 	}
 
+	/**
+	 * Default settings for ChatGPT
```
dkotter: Think this needs updated from ChatGPT

includes/Classifai/Providers/Watson/NLU.php
```diff
@@ -935,4 +962,36 @@ public function is_configured() {
...
+			$this->has_access( $feature ) &&
+			\Classifai\language_processing_features_enabled()
+		) {
+			$access = true;
+		}
+
+		/**
+		 * Filter to override permission to a IBM Watson classify content feature.
+		 *
+		 * @since 2.3.0
```
dkotter:
```suggestion
		 * @since 2.4.0
```

includes/Classifai/Providers/Provider.php
```diff
@@ -397,6 +430,44 @@ public function render_auto_caption_fields( $args ) {
...
 	}
 
+	/**
+	 * Render allowed users table.
+	 *
+	 * @param array $args The args passed to add_settings_field
+	 */
+	public function render_allowed_users( array $args = array() ) {
+		$setting_index = $this->get_settings();
+		$value         = ( isset( $setting_index[ $args['label_for'] ] ) ) ? $setting_index[ $args['label_for'] ] : array();
```
dkotter: This can be simplified I think:

```suggestion
		$value         = $setting_index[ $args['label_for'] ] ?? array();
```

includes/Classifai/Helpers.php
dkotter: I don't think so, I'd leave it off by default

includes/Classifai/Admin/UserProfile.php
```diff
@@ -0,0 +1,183 @@
...
+		add_action( 'edit_user_profile', array( $this, 'user_settings' ) );
+
+		add_action( 'personal_options_update', array( $this, 'save_user_settings' ) );
+		add_action( 'edit_user_profile_update', array( $this, 'save_user_settings' ) );
+	}
+
+	/**
+	 * Add ClassifAI features opt-out checkboxes to user profile and edit user.
+	 *
+	 * @param WP_User $user User object.
```
dkotter:
```suggestion
	 * @param \WP_User $user User object.
```

includes/Classifai/Admin/UserProfile.php
```diff
@@ -0,0 +1,183 @@
...
+	 * @return void
+	 */
+	public function user_settings( $user ) {
```
dkotter:
```suggestion
	public function user_settings( \WP_User $user ) {
```

includes/Classifai/Providers/AccessControl.php
```diff
@@ -0,0 +1,403 @@
+<?php
+/**
+ *  Class for handling the access control for ClassifAI features.
```
dkotter:
```suggestion
 * Class for handling the access control for ClassifAI features.
```

includes/Classifai/Providers/AccessControl.php
```diff
@@ -0,0 +1,403 @@
...
+		$this->provider = $provider;
+		$this->feature  = $feature;
+
+		$this->role_based_access_key  = $feature . '_role_based_access';
+		$this->roles_key              = $feature . '_roles';
+		$this->user_based_access_key  = $feature . '_user_based_access';
+		$this->user_based_opt_out_key = $feature . '_user_based_opt_out';
+		$this->users_key              = $feature . '_users';
+	}
+
```
dkotter, line 85: Some extra spaces here that could be removed

includes/Classifai/Providers/AccessControl.php
```diff
@@ -0,0 +1,403 @@
...
+	 */
+	protected $users_key;
+
+	/**
+	 * Constructor.
+	 *
+	 * @param Provider $provider The provider class instance.
+	 * @param string   $feature  The feature name.
+	 */
+	public function __construct( $provider, $feature ) {
```
dkotter:
```suggestion
	public function __construct( Provider $provider, string $feature ) {
```

includes/Classifai/Providers/AccessControl.php
```diff
@@ -0,0 +1,403 @@
...
+		return array_map( 'absint', $users );
+	}
+
+	/**
+	 * Add settings fields for Role/User based access.
+	 *
+	 * @param string $section Settings section.
+	 * @return void
+	 */
+	public function add_settings( $section = '' ) {
```
dkotter:
```suggestion
	public function add_settings( string $section = '' ) {
```

includes/Classifai/Providers/AccessControl.php
```diff
@@ -0,0 +1,403 @@
...
+	}
+
+	/**
+	 * Sanitization for the roles/users access options being saved.
+	 *
+	 * @param array $settings Array of settings about to be saved.
+	 *
+	 * @return array The sanitized settings to be saved.
+	 */
+	public function sanitize_settings( $settings ) {
```
dkotter:
```suggestion
	public function sanitize_settings( array $settings ) {
```

includes/Classifai/Helpers.php
```diff
@@ -772,3 +772,26 @@ function get_all_post_statuses() {
...
+
+/**
+ * Get the default settings for a feature.
+ *
+ * @since 2.4.0
+ *
+ * @param string $feature Feature key.
+ * @return array
+ */
+function get_feature_default_settings( $feature ) {
```
dkotter:
```suggestion
function get_feature_default_settings( string $feature ) {
```

includes/Classifai/Providers/Provider.php
```diff
@@ -465,4 +536,125 @@ public function is_configured() {
...
 	}
+
+	/**
+	 * Add settings fields for Role/User based access.
+	 *
+	 * @param string $feature Feature.
+	 * @param string $section Settings section.
+	 * @return void
+	 */
+	protected function add_access_settings( $feature, $section = '' ) {
```
dkotter:
```suggestion
	protected function add_access_settings( string $feature, string $section = '' ) {
```

includes/Classifai/Providers/Provider.php
```diff
@@ -465,4 +536,125 @@ public function is_configured() {
...
+
+	/**
+	 * Sanitization for the roles/users access options being saved.
+	 *
+	 * @param array  $settings Array of settings about to be saved.
+	 * @param string $feature  Feature key.
+	 *
+	 * @return array The sanitized settings to be saved.
+	 */
+	protected function sanitize_access_settings( $settings, $feature ) {
```
dkotter:
```suggestion
	protected function sanitize_access_settings( array $settings, string $feature ) {
```

dkotter:
Overall this looks really good, nice work here @iamdharmesh!

I've left a handful of minor code review comments. In addition to that, I have some thoughts on the UI and in particular some text we're showing.

In the settings for each feature, we have the following new text strings:

- Enables ability to select which role can access FEATURE
- Choose which roles are allowed to FEATURE
- Enables ability to select which users can access FEATURE
- Enables ability for users to opt-out from their user profile page

And then on the user profile page, we show:

- Opt out of using the FEATURE feature

For some features and settings, this text makes sense (like `Choose which roles are allowed to Classify Content`). But in most, the feature name doesn't pair well with this text (like `Enables ability to select which users can access Generate images`).

Open to ideas on any of these strings (especially from @jeffpaul) but I'd recommend the following:

- Since each of these strings is already within the settings section for a particular feature, I don't think we need to use the feature name itself. I'd suggest we standardize on:
     - Enables ability to select which roles can access this feature
     - Choose which roles are allowed to access this feature
     - Enables ability to select which users can access this feature

If we want to keep the feature name in there, I think we should change it to something like:

- Enables ability to select which role can access the FEATURE feature
- Choose which roles are allowed to access the FEATURE feature
- Enables ability to select which users can access the FEATURE feature

Again, I'd recommend the former over the latter but open to thoughts.

I think what we show on the profile page is fine, since it already has the `the FEATURE feature` part which makes more sense here in the context of a profile and not a settings page for the feature.

I would recommend though that we standardize on capitalization. Right now most features have the first word capitalized (like Generate captions) but some have both (like Classify Content). I suggest when using the feature name as a description, we should lower case all words. In addition, some of these descriptions have a trailing period and some don't. I think standardizing on no trailing period makes sense.

Another thing I'm seeing is that some of these settings descriptions don't have proper spacing (they're too close to the checkbox) while others are fine. See Watson NLU as an example:

Screenshot 2023-10-27 at 12 25 50 PM

Looks like there's some CSS scoped to `.classifai-content .classifai-wrap .classifai-input-description` that doesn't get applied everywhere. We should standardize on that to ensure we have proper spacing.

One area we already talked about but just mentioning again is this does make some of these settings pages pretty long (like Azure Image Processing). I think this is fine for now and will be helped once we complete the refactor to have stand-alone features in our settings (and even further once we do a refactor on these settings screens themselves) but just mentioning that there may be complaints on how unwieldy these pages are getting.

And I think final thing is looks like we are standardizing on `Image captions` being the feature name for the alt text generation. While I know we added that as a feature (where the alt text can be saved as a caption, description or alt text) I'm not sure this makes sense as the main feature name (also something for @jeffpaul to chime in on).

I think something like `Image descriptive text` instead of `Image captions` may make more sense, as well as `Generate descriptive text` instead of `Generate captions`

dkotter (changes requested):

src/js/admin.js
```diff
@@ -78,3 +81,83 @@ document.addEventListener( 'DOMContentLoaded', function () {
...
+		userBasedAccessCheckBoxes.forEach( function ( e ) {
+			e.addEventListener( 'change', toogleAllowedUsersRow );
+			e.dispatchEvent( new Event( 'change' ) );
+		} );
+	}
+} );
+
+// Search for users.
+( () => {
+	jQuery( '.classifai-search-users' ).select2( {
```
dkotter: Wondering if you looked at any Gutenberg components instead of using `select2` here? Thinking there may be a more native way to handle this. I know Gutenberg has an `Autocomplete` component, though not sure that allows you to populate from an API request. There's also a [`PostPicker` component](https://github.com/10up/ui-kit/tree/trunk/src/post-picker-block/blocks/post-picker) that 10up is working on that may provide some inspiration here.

Not saying this has to change but may be worth looking at alternatives that use built-in components

dkotter (changes requested):

includes/Classifai/Helpers.php
jeffpaul: Agreed

src/js/admin.js
jeffpaul: I agree on seeing what else is available as the select2 felt a bit off in terms of expected UI for that user selection.  @fabiankaegy any tips on your end for what could be used?

jeffpaul:
@dkotter @iamdharmesh I pretty much agree with Darin's recommendations in his https://github.com/10up/classifai/pull/606#issuecomment-1783343005

includes/Classifai/Providers/Watson/NLU.php
iamdharmesh: Thanks for catching this.

iamdharmesh:
Thank you for the detailed review @dkotter.

> - Since each of these strings is already within the settings section for a particular feature, I don't think we need to use the feature name itself. I'd suggest we standardize on:
>   - Enables ability to select which roles can access this feature
>   - Choose which roles are allowed to access this feature
>   - Enables ability to select which users can access this feature

Updated description strings to follow this. Thanks for adding a note for this.

> I would recommend though that we standardize on capitalization. Right now most features have the first word capitalized (like Generate captions) but some have both (like Classify Content). I suggest when using the feature name as a description, we should lower case all words. In addition,

Done. 

> some of these descriptions have a trailing period and some don't. I think standardizing on no trailing period makes sense.

Standardized this on having a trailing period. existing field descriptions follow the period at the end. So, followed it. please let me know if you think we should update all to "no trailing period".

> Another thing I'm seeing is that some of these settings descriptions don't have proper spacing (they're too close to the checkbox) while others are fine. See Watson NLU as an example:
>
> Screenshot 2023-10-27 at 12 25 50 PM
> Looks like there's some CSS scoped to `.classifai-content .classifai-wrap .classifai-input-description` that doesn't get applied everywhere. We should standardize on that to ensure we have proper spacing.

This was due to Watson NLU using an extended `render_input` function. Fixed. Thanks for catching this.

> One area we already talked about but just mentioning again is this does make some of these settings pages pretty long (like Azure Image Processing). I think this is fine for now and will be helped once we complete the refactor to have stand-alone features in our settings (and even further once we do a refactor on these settings screens themselves) but just mentioning that there may be complaints on how unwieldy these pages are getting.

We may explore quick win solutions here until we get a full refactor on these settings screens. Maybe something like vertical tabs for each section and only active section will be shown. This will help make the settings page smaller and organized. Obviously, this is not a perfect solution but something we can explore until we work on refactoring.


> And I think final thing is looks like we are standardizing on `Image captions` being the feature name for the alt text generation. While I know we added that as a feature (where the alt text can be saved as a caption, description or alt text) I'm not sure this makes sense as the main feature name (also something for @jeffpaul to chime in on).
>
> I think something like `Image descriptive text` instead of `Image captions` may make more sense, as well as `Generate descriptive text` instead of `Generate captions`

Updated accordingly. Thank you.

src/js/components/disable-feature-button.js
```diff
@@ -0,0 +1,34 @@
...
+	if (
+		! feature ||
+		! ClassifAI?.opt_out_enabled_features?.includes( feature )
+	) {
+		return null;
+	}
+
+	return (
+		<Button
+			href={ ClassifAI?.profile_url }
```
dkotter, line 22: Would be great to link down to the ClassifAI section, which I think should be as simple as adding `#classifai-profile-features-section` to the end of this URL (either here or where we localize this data)

dkotter (changes requested):

jeffpaul:
@iamdharmesh besides resolving merge conflicts, is there any additional work here before this can re-route for Darin's code review?

src/js/components/disable-feature-button.js
iamdharmesh: done.

iamdharmesh:
@jeffpaul @dkotter This is ready for review now.

dkotter (dismissed):

includes/Classifai/Admin/BulkActions.php
```diff
@@ -177,17 +179,16 @@ public function register_bulk_actions( $bulk_actions ) {
...
 	public function register_media_bulk_actions( $bulk_actions ) {
-		$computer_vision_settings = $this->computer_vision->get_settings();
-		$whisper_enabled          = $this->whisper->is_feature_enabled();
+		$whisper_enabled = $this->whisper->is_feature_enabled( 'speech_to_text' );
 
 		if (
-			'no' !== $computer_vision_settings['enable_image_tagging'] ||
-			! empty( $this->computer_vision->get_alt_text_settings() )
+			$this->computer_vision->is_feature_enabled( 'image_tagging' ) ||
+			$this->computer_vision->is_feature_enabled( 'image_captions' )
```
berkod, line 186: is it clear in instructions that this tagging and captions go together?

berkod (dismissed):

iamdharmesh:
@dkotter I have added E2E tests, could you please help with re-reviewing this? Thanks

dkotter (dismissed):

jeffpaul:
@iamdharmesh one item that came up in UAT today was the user opt-out links from within the editor/etc. is missing on the Generate Images page linked off the Media Library admin menu.  Please ensure we add the opt-out link to the user profile there (if the opt-out setting is allowed) and double-check that we're adding that link across all the ClassifAI features... thanks!

iamdharmesh:
@jeffpaul I have added a disable feature link to generate images popup and Generate excerpt metabox. Also, please find the status of each feature for the disable link. I have skipped most of the automated features and added a potential UI option to add a link if needed. Could you please help to check and please let me know if you think we should add disable links for the given features? We may handle it in follow-up PR to unblock this PR to be merged.

### Language Processing:
- **IBM Watson (Natural Language Understanding):** 
Skipped as we have confirmation popup development in progress. (disable link should be added there)

- **OpenAI Chatgpt**
    - **Excerpt Generation:** ✅ 
    - **Title Generation:** ✅ 
    - **Resize content:** ✅ 
    
- **OpenAI Embeddings:**
Skipped due to automated behavior

- **OpenAI Whisper:** 
Skipped due to automated behavior.
But we can add a link to the media model if needed (though it will make the UI a bit messy), what do you think?
Screenshot 2023-11-17 at 12 15 51 PM

- **Microsoft Azure (Text to Speech):** 
Skipped due to automated behavior.
We can add disable link in metabox and ClassifAI slotfill. what do you think?
image


### Image Processing:
- **AI Vision( Microsoft Azure):**
It skipped due to automated behavior, but we can add a link to the media model if needed.
image


- **OpenAI DALL·E (Image generation)**:  ✅ 

cc: @dkotter

jeffpaul:
I think adding the disable link to all the places you mention are worthwhile, but defer if @dkotter thinks any aren't necessary.  Also, if you could please push a commit to #609 for the post classification disabling that would be helpful... thanks!

dkotter:
Yeah, if we have places we can easily add this messaging, I think it's worth doing

dkotter (approved):

PR #609, opened 2023-11-02
Enhancement/566 Confirm Classification before Save

### LINKED ISSUE #566: Adjust handling of Language Processing's flow for classifying posts

As an alternate to #530, we could add a new button in the sidebar for "Classify Post":
ClassifAI_PostClassification_NewFlow_1-ClassifyPostButton

This mockup shows it next to the potential ClassifAI "Generate titles" button, but could also be in the "ClassifAI" sidebar panel or each of the "Categories", "Tags", etc. panels.

This would send the post content to Azure and display the returned classifications in a modal:
ClassifAI_PostClassification_NewFlow_2-ConfirmPostClassifcationModal

Someone could then add/remove the classifications and then save when the data from Azure meets expectations.

As a final catch, if there are no classifcations the Pre-Publish Panel could add a "Suggestion: Classify post" section:
ClassifAI_PostClassification_NewFlow_3-PrePublishPanel

Someone could then add/remove the classifications and then save when the data from Azure meets expectations.

### Designs

Here's how the current post classification flow works:
https://github.com/10up/classifai/assets/2818133/f0d13446-aad4-42eb-8df2-e4b8dca4f254

Essentially you get the categories/tags/etc. applied after publishing a post (or whatever post statuses are enabled in settings) with the ability to toggle on/off updates via a sidebar setting in the ClassifAI post sidebar section.

See #530.

### Code of Conduct

faisal-alvi:
### Description of the Change

* This is an enhancement PR that adds a pop-up when the classify post button is clicked and displays the terms coming from the API along with the existing terms. So users can add or remove terms and when they're expectations meet they can simply save it.
* In the popup, the terms with the prefix "[AI]" stand for the terms that are suggested by the API and do not exist in the post at the moment.
* Only those taxonomies will be displayed that are enabled from the settings.
* If users do not open the classification modal/popup, we remind them on the pre-publish panel as a suggestion to classify the post.
* The popup is also opened once the response is received.
* Users will see a notice when no unique terms are received from AI or all the received terms are already saved.
* Also bumps minimum WP version from 5.8 to 6.1.

### Screenshots

image

image

image

Closes #566

### How to test the Change
* Click a "Classify Post" button from the right sidebar in a post with enough content and confirm the popup shows the terms from ClassfAI and also includes the existing terms.
* Try to add/remove the terms and save and see if the final set of terms is applied to the post.
* Reload the page and click the Publish button, you should see the same suggestion dropdowns in the Pre-Publish panel.

### Credits

### Checklist:


faisal-alvi:
@jeffpaul We have discussed on a call that we should display a message when there are no more unique terms returned from the ClassifAI but IMO, we should keep displaying the taxonomy dropdown(s) even if there are no terms or only the terms that are already exist In the post. So if there are no terms, users may want to add more terms, and because they are already updating the terms in the popup, why not give them the ability to add terms in the empty terms taxonomy? For now, I have kept the taxonomy dropdowns regardless of whether there are no AI terms or any existing terms. cc @dkotter

jeffpaul:
@faisal-alvi the intent with the error message when AI returns no new, unique terms was to let the user know that the AI process completed but that there was nothing new.  Otherwise its not clear that the AI actually processed any recommendations.  I agree we want to keep the taxonomy sections still available even if no responses are returned, but displaying a message will help inform the user that what they're seeing is the set of terms to consider.

faisal-alvi:
@jeffpaul the terms will be displayed only when the API response is completed. So if user is seeing the terms in the dropdowns, it means the API is already ran. If API is in progress, user will not see any dropdowns. We do not show existing terms until AI terms are received.

faisal-alvi:
@jeffpaul the logic is tweaked so now users will see a notice when no unique terms are received from AI or all the received terms are already saved. The popup is also opened once the response is received.

image

faisal-alvi:
@dkotter this is ready for your review

jeffpaul:
@faisal-alvi if ClassifAI has no response from the API, we should still show the field (eg Keywords and Entities in your example above) and not hide that with the error message.  Let's show the error message above the entry field for that section, but still have the entry field in the modal.

Also, if the modal only appears once there's a result from the API, then we'll want to ensure there's some sort of loading animation so an editor/author knows their "Classify post" button press was accepted and ClassifAI is in-progress.  In general, I'm fine with the modal opening immediately and any necessary loading animation to happen within that modal (something somewhat similar to the loading animation for the expand/condense text feature).

src/js/gutenberg-plugin.js
```diff
@@ -159,19 +248,81 @@ const ClassifAIGenerateTagsButton = () => {
...
+			/>
+			<div className="classifai-modal__footer">
+				<div className="classifai-modal__notes">
+					{ __(
+						'Note that the lists above include any pre-existing terms from this post.',
+						'classifai'
+					) }
+					<br />
+					{ __(
+						'Al recommendations saved to this post will not include the "[AI]" text.',
```
```suggestion
						'AI recommendations saved to this post will not include the "[AI]" text.',
```

src/js/gutenberg-plugin.js
```diff
@@ -159,19 +248,81 @@ const ClassifAIGenerateTagsButton = () => {
...
+					contentPostType: 'page',
+					taxQuery: updatedTaxQuery,
+					taxTermsAI: taxTermsAI || {},
+					isLoading,
+				} }
+			/>
+			<div className="classifai-modal__footer">
+				<div className="classifai-modal__notes">
+					{ __(
+						'Note that the lists above include any pre-existing terms from this post.',
```
dkotter: Since this feature can be used on other post types besides `post`, would be great to pull in the actual post type name to use here and other places

src/js/gutenberg-plugin.js
```diff
@@ -159,19 +248,81 @@ const ClassifAIGenerateTagsButton = () => {
...
+	}
+
+	const modalData = (
 		<>
+			<TaxonomyControls
+				onChange={ ( newTaxQuery ) => {
+					setTaxQuery( newTaxQuery );
+				} }
+				query={ {
+					contentPostType: 'page',
```
dkotter: This is hardcoded to `page`, so won't work unless you're on a `page` post type. This should be dynamic

src/js/gutenberg-plugin.js
```diff
@@ -159,19 +248,81 @@ const ClassifAIGenerateTagsButton = () => {
...
+					onClick={ () => saveTerms( updatedTaxQuery ) }
+				>
+					{ __( 'Save', 'classifai' ) }
+				</Button>
+			</div>
+		</>
+	);
+
+	return (
+		<div id="classify-post-componenet">
```
```suggestion
		<div id="classify-post-component">
```

dkotter (changes requested):

dkotter:
Gave this a quick review and test and found a few things:

1. I couldn't get this to work and was only seeing this when clicking the Classify button. I flagged this in my review but the issue is the taxonomy controls component was hardcoded to only work on Pages, so that needs to be dynamic based on the post type being viewed.

Classify Modal

2. I think we also need a minimum width on the modal, otherwise there's a chance the text and button can be too close, see:

Modal with no terms

3. For the pre-publish panel, it seems we run classification as soon as that panel is opened. Should we instead show the same `Classify` button and only run if someone clicks that? That's how we handle the excerpt generation

4. Something that may have gotten lost in this request is the ability to easily disable auto-classification. Right now you have to toggle that off within each individual post (and even then only seems to work after you save twice. The first save still triggers classification). I think we need to add a global setting to this feature to disable auto-classification across the board, which would then have the toggle be turned off on each post and the `Classify` button would then show. This would ensure classification has to be done by a user and there's no risk of pre-existing terms being overwritten (which was one of the big concerns we need to fix)

dkotter:
@jeffpaul
> then we'll want to ensure there's some sort of loading animation

There is an animation that shows next to the button prior to the modal loading in:
Screenshot 2023-11-03 at 12 54 11 PM

src/js/gutenberg-plugin.js
```diff
@@ -8,11 +8,14 @@ import {
 	Icon,
 	ToggleControl,
 	BaseControl,
+	Modal,
 } from '@wordpress/components';
 import { __, sprintf } from '@wordpress/i18n';
 import { registerPlugin } from '@wordpress/plugins';
 import { useState, useEffect, useRef } from '@wordpress/element';
 import { store as postAudioStore } from './store/register';
+import TaxonomyControls from '../../includes/Classifai/Blocks/recommended-content-block/inspector-controls/taxonomy-controls';
```
dkotter: I think ultimately this recommended content block will probably go away, due to Azure Personalizer being sunset. I'm wondering if maybe we should pull this component out from the `Blocks` directory and into our general `src/js` directory and it can then be pulled into the block where needed?

dkotter (changes requested):

jeffpaul:
> 3. For the pre-publish panel, it seems we run classification as soon as that panel is opened. Should we instead show the same Classify button and only run if someone clicks that? That's how we handle the excerpt generation

Yes.

> 4. Something that may have gotten lost in this request is the ability to easily disable auto-classification. Right now you have to toggle that off within each individual post (and even then only seems to work after you save twice. The first save still triggers classification). I think we need to add a global setting to this feature to disable auto-classification across the board, which would then have the toggle be turned off on each post and the Classify button would then show. This would ensure classification has to be done by a user and there's no risk of preexisting terms being overwritten (which was one of the big concerns we need to fix)

I would actually advocate for a Post Classification setting to classify automatically which is our historical approach, have this enabled for any site currently using the feature otherwise disabled for anyone not yet using the feature (essentially the new default).  We might then be able to remove the toggle in the editor sidebar.  I could also be convinced that we just make human-required review the default and someone would have to opt for automatic classification instead admitting this could feel like a break for anyone currently relying on auto-classification.

src/js/gutenberg-plugin.js
faisal-alvi: Yes, I agree

includes/Classifai/Providers/Watson/NLU.php
```diff
@@ -448,20 +465,66 @@ public function render_input( $args ) {
 			case 'checkbox':
 				$attrs = ' value="1"' . checked( '1', $value, false );
 				break;
+				break;
```
dkotter: Double `break` here

includes/Classifai/Providers/Watson/NLU.php
```diff
@@ -448,20 +465,66 @@ public function render_input( $args ) {
...
+			case 'radio':
+				if ( 'classification_mode' === $args['option_index'] ) {
+					// Detect the existing vs new users and serve default value accordingly.
+					$get_classification_mode = get_classification_mode();
+					if ( $value === $get_classification_mode ) {
+						$attrs = ' value="' . esc_attr( $value ) . '" checked="checked"';
+					}
+				}
+				$attrs = empty( $attrs ) ? ' value="' . esc_attr( $value ) . '"' . checked( $setting_index, $value, false ) : $attrs;
```
dkotter: I'm getting a PHP warning here (`Array to string conversion`) here because `$setting_index` is an `array` while `$value` is a `string`

dkotter (changes requested):
In addition to the couple things flagged here, there's a few other things that I think still need to be addressed:

1. This may be related to the PHP warning I flagged but I can't change from the `Automatic classification` mode to the `Manual review` mode. Anytime I select `Manual review` and then save, it doesn't stay checked
2. I know this wasn't mentioned on the issue but one thing we discussed on a call was that the toggle we currently have requires you to save twice before it takes effect. This means if I toggle it off and save a post, it will still automatically classify the post. The next time I save it won't. This also is true in the opposite, if I want to turn it back on, the first save won't work but the second will. This is a long-standing bug but I think we should fix that here if possible. I think this may be due to the hook we use to save this meta
3. For the error state we show if no new terms are found, I have a couple suggestions. I think the red color we use is too bright. I'd match the red color WordPress uses for the `Move to trash` button. I also think we should move that error message to show below the input and below the helper text but above the `hr` and change the error message a bit:
Error state
4. Seems we need to properly handle entities in term names:
entities in terms
5. Would be nice to add some E2E tests here to ensure the manual mode works

dkotter:
> I know this wasn't mentioned on the issue but one thing we discussed on a call was that the toggle we currently have requires you to save twice before it takes effect.

As mentioned on the call, I tested this again and does seem to be working now so I think we're good on this point. Thanks!

jeffpaul:
@faisal-alvi is there any additional work needed here before this moves along for additional code review?

faisal-alvi:
@dkotter this is ready for your review.

dkotter (dismissed):

berkod (dismissed):

jeffpaul:
@faisal-alvi one item that came up in UAT today was that when we changed the Taxonomy selected in the ClassifAI settings (e.g., changing Category from Watson Category to the WP native Category) that the modal displayed in the editor did not properly update to show the newly selected mapped taxonomy.  So, please look to update to ensure the modal shows whichever taxonomies are selected from the settings.

dkotter:
Adding a bit more to the above, did a little digging and seems one of the issues is these lines: https://github.com/10up/classifai/pull/609/files#diff-b387b736a54bc04206792307dd8c2dffb5b5c416d7189f0063c352b2e8ce2142R256-R259. Seems we only support the Watson custom taxonomies but we need to support all taxonomies, since we allow them to map the data returned from Watson to any registered taxonomy within the settings page.

I think the most common scenario is someone mapping those to the core category and tag taxonomies so we need to ensure those work. If I comment out those mentioned lines, I do then see inputs in the modal for those taxonomies but no terms show up, so I'm assuming something else here needs to be changed as well.

I guess one thing we may end up needing to figure out is how to address hierarchical taxonomies (like categories). Seems we may need to have UI in place for both hierarchical and non-hierarchical taxonomies, whereas now we really only support non-hierarchical.

faisal-alvi:
The popup was limited to the Watson taxonomies,

https://github.com/10up/classifai/blob/c39ddc60a23f8b3b7a8e32e816d071de63c5e4a3/src/js/taxonomy-controls.js#L230-L235

I have updated it to cover any taxonomy selected in the ClassifAI settings.

> I guess one thing we may end up needing to figure out is how to address hierarchical taxonomies (like categories). Seems we may need to have UI in place for both hierarchical and non-hierarchical taxonomies, whereas now we really only support non-hierarchical.

@dkotter This requires some discussion on UI. I think this should be handled later in the other issue/PR.

dkotter (dismissed):

PR #610, opened 2023-11-03
Show the default prompt as the first prompt in our settings

dkotter:
### Description of the Change

Following on from #602, this PR ensures that our default prompt within ClassifAI always shows as the first prompt in our settings. This prompt can't be edited and can't be removed. This makes it easier for someone to go back to using the ClassifAI default prompt.

Important to note that to keep things simple, this ClassifAI default prompt is added and saved just like any other prompt. This does mean the prompt itself is saved in the database. In order to support situations where we may change the default prompt at a code-level, if this prompt is selected, instead of using the prompt text that is stored in the database, we use the prompt that is in the code.

The new flow will be as follows:

When first setting up a feature that uses prompts (like Title Generation), the only prompt you'll see is the default that is included in ClassifAI. The prompt fields are there but hidden (this allows us to save this as the default prompt):

Default prompt

If a custom prompt is needed, instead of editing the default prompt (which is how things landed in #602), you'll now need to click the `Add new prompt` button and then set a custom prompt in the fields that are added. In addition, to use the new prompt, you'll need to set that as the default:

New prompt added

You can add as many new prompts as needed, set whichever of those you want to be the default and delete any of those prompts. But you can't ever delete the ClassifAI default prompt but you can at any point set that as the default again.

### How to test the Change

1. Checkout the changes in this PR
2. If you've previously tested #602, you may need to delete the existing prompt data (easiest is to delete the `openai_chatgpt` option)
3. Ensure you have proper OpenAI credentials in place and at least one feature turned on (like excerpt generation)
4. Without entering a new prompt, test that feature out and ensure it works
5. Add a new prompt but don't set it as the default. Ensure things save and test the feature out again, ensuring it works
6. Set this new prompt as the default and ensure it saves and the feature still works
7. Can continue testing by adding more prompts across all the various features, ensuring you can change which one is the default and can delete prompts
8. Ensure you can set the ClassifAI default as the default prompt again and that the feature works
9. If desired, you can add some logging or debug statements in any of the features to ensure the prompt that gets used is the prompt you're setting as the default

### Changelog Entry

> Changed - Ensure the default prompts in ClassifAI show as the first prompt in our settings and that they can't be removed or edited.

### Credits

Props @dkotter, @jeffpaul 

### Checklist:

### DIFF
--- a/includes/Classifai/Providers/OpenAI/ChatGPT.php
+++ b/includes/Classifai/Providers/OpenAI/ChatGPT.php
@@ -436,7 +436,7 @@ class ChatGPT extends Provider {
 				'label_for'     => 'generate_excerpt_prompt',
 				'placeholder'   => $this->generate_excerpt_prompt,
 				'default_value' => $default_settings['generate_excerpt_prompt'],
-				'description'   => __( "Note the following variables that can be used in the prompt and will be replaced with content: {{WORDS}} will be replaced with the desired excerpt length setting. {{TITLE}} will be replaced with the item's title.", 'classifai' ),
+				'description'   => __( 'Note the following variables that can be used in the prompt and will be replaced with content: {{WORDS}} will be replaced with the desired excerpt length setting. {{TITLE}} will be replaced with the item\'s title.', 'classifai' ),
 			]
 		);
 
--- a/includes/Classifai/Providers/OpenAI/ChatGPT.php
+++ b/includes/Classifai/Providers/OpenAI/ChatGPT.php
@@ -436,7 +436,7 @@ class ChatGPT extends Provider {
 				'label_for'     => 'generate_excerpt_prompt',
 				'placeholder'   => $this->generate_excerpt_prompt,
 				'default_value' => $default_settings['generate_excerpt_prompt'],
-				'description'   => __( "Enter your custom prompt. Note the following variables that can be used in the prompt and will be replaced with content: {{WORDS}} will be replaced with the desired excerpt length setting. {{TITLE}} will be replaced with the item's title.", 'classifai' ),
+				'description'   => __( "Note the following variables that can be used in the prompt and will be replaced with content: {{WORDS}} will be replaced with the desired excerpt length setting. {{TITLE}} will be replaced with the item's title.", 'classifai' ),
 			]
 		);
 
@@ -501,7 +501,6 @@ class ChatGPT extends Provider {
 				'label_for'     => 'generate_title_prompt',
 				'placeholder'   => $this->generate_title_prompt,
 				'default_value' => $default_settings['generate_title_prompt'],
-				'description'   => __( 'Enter a custom prompt, if desired.', 'classifai' ),
 			]
 		);
 
@@ -570,7 +569,6 @@ class ChatGPT extends Provider {
 				'label_for'     => 'shrink_content_prompt',
 				'placeholder'   => $this->shrink_content_prompt,
 				'default_value' => $default_settings['shrink_content_prompt'],
-				'description'   => __( 'Enter a custom prompt, if desired.', 'classifai' ),
 			]
 		);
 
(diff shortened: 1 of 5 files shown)


dkotter:
Note that there are some failing tests here due to these changes but wanted to get some feedback on if we think this is the right approach before spending time getting those fixed up (and adding some additional tests)

ravinderk:
@dkotter  I exceed the daily query limit, so I cannot test this feature.

I have following questions -
- Is it intentional to show `Classifai default prompt` as  `Default prompt` while selecting it? I see correct text when reloading the page.
image

dkotter:
> Is it intentional to show Classifai default prompt as Default prompt while selecting it? I see correct text when reloading the page.

Yes, though we can update this if we think it's worth it. I didn't change any of the JS logic for that, just on the PHP side, so it only shows `Classifai default prompt` after loading or reloading the page.

Edit: as part of the changes requested below, I've actually changed this to just always show `Default prompt`, as we now show `ClassifAI default prompt` as part of the main text

jeffpaul:
Structurally I think this is good.  As a future UI iteration, I'd like to see us collapse a good bit of the UI around the default (we don't need the full text field/area) so that it takes up less vertical space.  Something more like this:

classifai-default-prompt

classifai-default-prompt-2

However, I'm fine with the current implementation as it achieves the same purpose but a future tweak could make the UI a bit simpler/cleaner.

dkotter:
> Structurally I think this is good. As a future UI iteration, I'd like to see us collapse a good bit of the UI around the default (we don't need the full text field/area) so that it takes up less vertical space. Something more like this:

Yeah, I like this and considered something similar before sticking with what we had. But it was easy enough to make this change so I've gone ahead and done that:

Default prompt only

One custom prompt

includes/Classifai/Providers/OpenAI/ChatGPT.php
```diff
@@ -436,7 +436,7 @@ public function setup_fields_sections() {
 				'label_for'     => 'generate_excerpt_prompt',
 				'placeholder'   => $this->generate_excerpt_prompt,
 				'default_value' => $default_settings['generate_excerpt_prompt'],
-				'description'   => __( "Enter your custom prompt. Note the following variables that can be used in the prompt and will be replaced with content: {{WORDS}} will be replaced with the desired excerpt length setting. {{TITLE}} will be replaced with the item's title.", 'classifai' ),
+				'description'   => __( "Note the following variables that can be used in the prompt and will be replaced with content: {{WORDS}} will be replaced with the desired excerpt length setting. {{TITLE}} will be replaced with the item's title.", 'classifai' ),
```
ravinderk:
```suggestion
				'description'   => __( 'Note the following variables that can be used in the prompt and will be replaced with content: {{WORDS}} will be replaced with the desired excerpt length setting. {{TITLE}} will be replaced with the item\'s title.', 'classifai' ),
```

includes/Classifai/Providers/Provider.php
```diff
@@ -249,20 +249,40 @@ public function render_prompt_repeater_field( array $args ): void {
...
+					</p>
+				<?php endif; ?>
+
 				<input type="hidden"
 					name="<?php echo esc_attr( $field_name_prefix . "[$field_index][default]" ); ?>"
 					value="<?php echo esc_attr( $prompt['default'] ?? '' ); ?>"
 					class="js-setting-field__default">
+				<input type="hidden"
+					name="<?php echo esc_attr( $field_name_prefix . "[$field_index][original]" ); ?>"
+					value="<?php echo esc_attr( $prompt['original'] ?? '' ); ?>">
```
ravinderk, line 285: @dkotter I suggest use the trinary short syntax `?:` instead of the null coalescing operator`??`, because `original` value can not be `null`, right?

ravinderk (dismissed):
@dkotter I left a few minor comments which are not blockers. This pull request is working as expected.

includes/Classifai/Providers/Provider.php
dkotter: So short ternary's aren't allowed by our PHPCS config so we would need to change this to `! empty( $prompt['original'] ? $prompt['original'] : ''` which I'm fine with changing that if we feel it's better. The goal here was to be defensive for a situation where this array value wasn't set (which should never happen) and using the null coalescing operator would catch that.

PR #620, opened 2023-11-22
Classify content within existing Taxonomy terms

### LINKED ISSUE #592: Classify content within existing Taxonomies

ClassifAI makes use of both IBM Watson NLU and OpenAI Embeddings to power the Post Content Classification features, with each having a slightly different approach/featureset.  Watson allows for new terms to be added to a site, for verbosity/precision threshold configuration, and a preview of how its settings changes would impact sample site posts.  OpenAI allows for limiting terms to existing site taxonomies, but no thresholds (to be done in #577) or previewing (to be done in #591).

Thus both offerings could be enhanced to have feature parity on post content classification, though the focus in this issue will be allowing IBM Watson to only classify posts with existing terms on a site.  We'll likely want this controlled somehow in the settings of the feature and determining what the default handling should be (while not changing how an existing ClassifAI user might have this feature configured).

From a high-level estimation perspective, this will likely be a rather straightforward effort (“t-shirt sized” S) with a little advance research with IBM Watson to determine the best technical approach on parity, and from then feasibly completed within 1-3 days.

### Designs

n/a

n/a

### Code of Conduct

iamdharmesh:
### Description of the Change
The PR introduces a new setting option in content classification (IBM Watson NLU) to classify content within existing taxonomy terms. By default, it will adhere to the current behavior, recommending and adding new terms to the post. When the user chooses "Only classify based on existing terms" in the classification method, it will recommend terms only from the existing terms.

image

Closes #592 

### How to test the Change
1. Go to `Tools` > `ClassifAI` > `Language Processing` > `IBM Watson`
1. Choose "Only classify based on existing terms" in the classification method.
1. Create/Edit the post and click the "Suggest terms & tags" button from the classifAI sidebar panel.
1. Verify that suggested terms/tags are from the existing terms only.
1. Go to ClassifAI settings, choose "Add recommended terms " in the classification method and verify that it now suggests new terms/tags as well.
1. Go to ClassifAI settings, and verify that the results of terms/tags in the language processing previewer are also aligned with the current classification method setting.

### Credits
Props @jeffpaul @iamdharmesh 

### Checklist:

### DIFF
--- a/includes/Classifai/Providers/Provider.php
+++ b/includes/Classifai/Providers/Provider.php
@@ -399,9 +399,13 @@ abstract class Provider {
 	 */
 	public function render_checkbox_group( array $args = array() ) {
 		$setting_index = $this->get_settings();
+		$options       = $args['options'] ?? [];
+		if ( ! is_array( $options ) ) {
+			return;
+		}
 
 		// Iterate through all of our options.
-		foreach ( $args['options'] as $option_value => $option_label ) {
+		foreach ( $options as $option_value => $option_label ) {
 			$value       = '';
 			$default_key = array_search( $option_value, $args['default_values'], true );
 
@@ -454,9 +458,13 @@ abstract class Provider {
 	public function render_radio_group( array $args = array() ) {
 		$setting_index = $this->get_settings();
 		$value         = $setting_index[ $args['label_for'] ] ?? '';
+		$options       = $args['options'] ?? [];
+		if ( ! is_array( $options ) ) {
+			return;
+		}
 
 		// Iterate through all of our options.
-		foreach ( $args['options'] as $option_value => $option_label ) {
+		foreach ( $options as $option_value => $option_label ) {
 			// Render radio button.
 			printf(
 				'<p>
--- a/includes/Classifai/Providers/Provider.php
+++ b/includes/Classifai/Providers/Provider.php
@@ -447,6 +447,42 @@ abstract class Provider {
 	}
 
 	/**
+	 * Render a group of radio.
+	 *
+	 * @param array $args The args passed to add_settings_field
+	 */
+	public function render_radio_group( array $args = array() ) {
+		$setting_index = $this->get_settings();
+		$value         = $setting_index[ $args['label_for'] ] ?? '';
+
+		// Iterate through all of our options.
+		foreach ( $args['options'] as $option_value => $option_label ) {
+			// Render radio button.
+			printf(
+				'<p>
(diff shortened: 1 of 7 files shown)


faisal-alvi (commented):
### Suggestion

The popup would show the notice,

https://github.com/10up/classifai/blob/e9198e508d7bf961fe9b6d407828470160e88670/src/js/gutenberg-plugin.js#L266

We should also tweak this notice when the existing terms-based classification setting is enabled. We can tweak to something like:

```
'Note that the lists above include only pre-existing terms from this %s. You can change this classification method in settings.'
```
and add a link on the 'settings' text to the settings page.

iamdharmesh:
@faisal-alvi 
> We should also tweak this notice when the existing terms-based classification setting is enabled. 

I think `Note that the lists above include any pre-existing terms from this Post` is meant for existing terms of the specific post and not for all existing terms. So, this might not need an update.

includes/Classifai/Providers/Provider.php
```diff
@@ -446,6 +446,42 @@ public function render_checkbox_group( array $args = array() ) {
...
+	 * Render a group of radio.
+	 *
+	 * @param array $args The args passed to add_settings_field
+	 */
+	public function render_radio_group( array $args = array() ) {
+		$setting_index = $this->get_settings();
+		$value         = $setting_index[ $args['label_for'] ] ?? '';
+
+		// Iterate through all of our options.
+		foreach ( $args['options'] as $option_value => $option_label ) {
```
dkotter: Minor thing but wondering if we need to add a check that `$args['options']` actually exists and is an array before we try using it?

iamdharmesh: Thanks for note, handled in [b24e017](https://github.com/10up/classifai/pull/620/commits/b24e017041936f009557cbabc58e7cb4581f67b0)

dkotter (dismissed):

berkod (dismissed):

jeffpaul:
@iamdharmesh @dkotter one item I noticed during UAT today was perhaps a copy tweak, sharing here to see if this resonates with either of you for a quick commit to change before merging this in...
- Changing `Add all suggested terms` to `Recommend terms even if they do not exist on the site`
- Changing `Only add suggested terms that already exist` to `Only recommend terms that already exist on the site`

dkotter:
> one item I noticed during UAT today was perhaps a copy tweak, sharing here to see if this resonates with either of you for a quick commit to change before merging this in.

Makes sense to me. Pushed an update for that:

Classification method settings

PR #621, opened 2023-11-23
Enhancement/577 Threshold values for Embeddings

### LINKED ISSUE #577: Threshold value for Embeddings

I am using classifai to provide suggestions to our editorial team for adding categories and tags to post as well as reapplying cats and tags to all of our posts for a big migration. The main problem i've found with embeddings is false positives when applying categories automatically in our migration, it would be great if you could add a threshold value to the embeddings, so that not all cats or tags get applied, but only ones that are highly relevant. This is already done in the IBM Watson interface and seems possible based on my quick glance of the openai api. 

### Designs

_No response_

_No response_

### Code of Conduct

faisal-alvi:
### Description of the Change

* This PR adds the threshold setting fields for each taxonomy in the Embeddings settings, allowing users to set the threshold value.
* The default value is set to 75%.
* These calculations are based on https://github.com/10up/classifai/issues/577#issuecomment-1808636943.

Closes #577

### Credits

Props @timatron @dkotter @faisal-alvi 

### Checklist:

### DIFF
--- a/includes/Classifai/Providers/OpenAI/EmbeddingCalculations.php
+++ b/includes/Classifai/Providers/OpenAI/EmbeddingCalculations.php
@@ -58,6 +58,13 @@ class EmbeddingCalculations {
 		// Do the math.
 		$distance = 1.0 - ( $combined_average / sqrt( $source_average * $compare_average ) );
 
+		/**
+		 * Filter the threshold for the similarity calculation.
+		 *
+		 * @param float $threshold The threshold to use.
+		 */
+		$threshold = apply_filters( 'classifai_threshold', $threshold );
+
 		// Ensure we are within the range of 0 to 1.0 (i.e. $threshold).
 		return max( 0, min( abs( (float) $distance ), $threshold ) );
 	}
--- a/includes/Classifai/Providers/OpenAI/EmbeddingCalculations.php
+++ b/includes/Classifai/Providers/OpenAI/EmbeddingCalculations.php
@@ -61,7 +61,12 @@ class EmbeddingCalculations {
 		/**
 		 * Filter the threshold for the similarity calculation.
 		 *
-		 * @param float $threshold The threshold to use.
+		 * @since 2.6.0
+		 * @hook classifai_threshold
+		 *
+		 * @param {float} $threshold The threshold to use.
+		 *
+		 * @return {float} The threshold to use.
 		 */
 		$threshold = apply_filters( 'classifai_threshold', $threshold );
 
--- a/includes/Classifai/Providers/OpenAI/EmbeddingCalculations.php
+++ b/includes/Classifai/Providers/OpenAI/EmbeddingCalculations.php
@@ -61,7 +61,7 @@ class EmbeddingCalculations {
 		/**
 		 * Filter the threshold for the similarity calculation.
 		 *
-		 * @since 2.6.0
+		 * @since 2.5.0
 		 * @hook classifai_threshold
 		 *
 		 * @param {float} $threshold The threshold to use.
--- a/includes/Classifai/Providers/OpenAI/EmbeddingCalculations.php
+++ b/includes/Classifai/Providers/OpenAI/EmbeddingCalculations.php
@@ -15,9 +15,11 @@ class EmbeddingCalculations {
 	 *
 	 * @param array $source_embedding Embedding data of the source item.
 	 * @param array $compare_embedding Embedding data of the item to compare.
+	 * @param float $threshold The threshold to use for the similarity calculation.
+	 *
 	 * @return bool|float
 	 */
-	public function similarity( array $source_embedding = [], array $compare_embedding = [] ) {
+	public function similarity( array $source_embedding = [], array $compare_embedding = [], $threshold = 1 ) {
 		if ( empty( $source_embedding ) || empty( $compare_embedding ) ) {
 			return false;
 		}
@@ -56,8 +58,8 @@ class EmbeddingCalculations {
 		// Do the math.
 		$distance = 1.0 - ( $combined_average / sqrt( $source_average * $compare_average ) );
 
-		// Ensure we are within the range of 0 to 2.0.
-		return max( 0, min( abs( (float) $distance ), 2.0 ) );
+		// Ensure we are within the range of 0 to 1.0 (i.e. $threshold).
+		return max( 0, min( abs( (float) $distance ), $threshold ) );
 	}
 
 }
--- a/includes/Classifai/Providers/OpenAI/Embeddings.php
+++ b/includes/Classifai/Providers/OpenAI/Embeddings.php
@@ -272,7 +272,7 @@ class Embeddings extends Provider {
 			// Sanitize the threshold setting.
 			$taxonomy_key = $taxonomy_key . '_threshold';
 			if ( isset( $settings['taxonomies'][ $taxonomy_key ] ) && '0' !== $settings['taxonomies'][ $taxonomy_key ] ) {
-				$threshold_value = min( absint( $settings['taxonomies'][ $taxonomy_key ] ), 100 );
+				$threshold_value                             = min( absint( $settings['taxonomies'][ $taxonomy_key ] ), 100 );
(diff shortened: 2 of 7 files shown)


includes/Classifai/Providers/OpenAI/EmbeddingCalculations.php
```diff
@@ -56,8 +58,15 @@ function( $x ) {
 		// Do the math.
 		$distance = 1.0 - ( $combined_average / sqrt( $source_average * $compare_average ) );
 
-		// Ensure we are within the range of 0 to 2.0.
-		return max( 0, min( abs( (float) $distance ), 2.0 ) );
+		/**
+		 * Filter the threshold for the similarity calculation.
+		 *
+		 * @param float $threshold The threshold to use.
+		 */
```
dkotter: This hook docblock will need updated to match our other hooks so it ends up in our doc site

includes/Classifai/Providers/Provider.php
```diff
@@ -446,6 +453,34 @@ public function render_checkbox_group( array $args = array() ) {
...
+	 */
+	public function render_threshold_field( $args, $option_value, $value ) {
+		printf(
+			'<p class="threshold_wrapper">
+				<label for="%1$s_%2$s_%3$s">%4$s</label>
+				<br>
+				<input type="number" id="%1$s_%2$s_%3$s" class="small-text" name="classifai_%1$s[%2$s][%3$s]" value="%5$s" />				
+			</p>',
+			esc_attr( $this->option_name ),
+			esc_attr( $args['label_for'] ),
```
dkotter: Wondering if we need a simple check here to ensure this value actually exists? Something like

```
esc_attr( $args['label_for'] ?? '' )
```

includes/Classifai/Providers/OpenAI/Embeddings.php
```diff
@@ -354,6 +363,34 @@ public function supported_post_types() {
...
+
+		// Convert $threshold (%) to decimal.
+		$threshold = 1 - ( (float) $threshold / 100 );
+
+		/**
+		 * Filter the threshold for the similarity calculation.
+		 *
+		 * @param float $threshold The threshold to use.
+		 * @param string $taxonomy The taxonomy to get the threshold for.
+		 */
```
dkotter: Same comment on this hook docblock

dkotter (changes requested):

dkotter (changes requested):
Note that there's some PHPCS warnings that we should look to fix: https://github.com/10up/classifai/actions/runs/6968812438/job/18963465180#step:8:16. Not sure why it's not showing this check as having failed, maybe we only fail on errors and not warnings? But these are easy enough to fix up so ideally we address these

EDIT: looked into this and seems it's due to the 10up-Default ruleset we extend, based on this line: https://github.com/10up/phpcs-composer/blob/master/10up-Default/ruleset.xml#L33

Personally I think we should remove this so we get failures on warnings, otherwise those items never get fixed up.

faisal-alvi:
@dkotter thanks for the feedback. I've implemented them (and also fixed the minor phpcs warnings). Please check.

> Personally I think we should remove this so we get failures on warnings, otherwise those items never get fixed up.

Agree. Should I raise a PR there?

includes/Classifai/Providers/OpenAI/EmbeddingCalculations.php
```diff
@@ -56,8 +58,20 @@ function( $x ) {
 		// Do the math.
 		$distance = 1.0 - ( $combined_average / sqrt( $source_average * $compare_average ) );
 
-		// Ensure we are within the range of 0 to 2.0.
-		return max( 0, min( abs( (float) $distance ), 2.0 ) );
+		/**
+		 * Filter the threshold for the similarity calculation.
+		 *
+		 * @since 2.6.0
```
dkotter:
```suggestion
		 * @since 2.5.0
```

includes/Classifai/Providers/OpenAI/Embeddings.php
```diff
@@ -354,6 +363,39 @@ public function supported_post_types() {
 		return apply_filters( 'classifai_openai_embeddings_post_types', $this->get_supported_post_types() );
 	}
 
+	/**
+	 * Get the threshold for the similarity calculation.
+	 *
+	 * @since 2.6.0
```
dkotter:
```suggestion
	 * @since 2.5.0
```

includes/Classifai/Providers/OpenAI/Embeddings.php
```diff
@@ -354,6 +363,39 @@ public function supported_post_types() {
...
+			$threshold = isset( $settings['taxonomies'][ $taxonomy . '_threshold' ] ) ? $settings['taxonomies'][ $taxonomy . '_threshold' ] : 75;
+		}
+
+		// Convert $threshold (%) to decimal.
+		$threshold = 1 - ( (float) $threshold / 100 );
+
+		/**
+		 * Filter the threshold for the similarity calculation.
+		 *
+		 * @since 2.6.0
```
dkotter:
```suggestion
		 * @since 2.5.0
```

includes/Classifai/Providers/Provider.php
```diff
@@ -446,6 +453,34 @@ public function render_checkbox_group( array $args = array() ) {
 		}
 	}
 
+	/**
+	 * Render a threshold field.
+	 *
+	 * @since 2.6.0
```
dkotter:
```suggestion
	 * @since 2.5.0
```

dkotter (changes requested):
In testing, found a few things:

1. When first going to the settings page, I don't see any default values for the thresholds. After saving the first time, they get set to 75% but those should be set on initial load:
Default values blank
2. If the Watson feature is turned off, doesn't seem we load the language-processing.css file, so styles aren't getting loaded:
Missing styles
3. Don't think this is a side-effect of this PR (probably due to changes in #609) but no longer seeing the Classify button if the feature is disabled on the post level:

Missing button

dkotter:
> Agree. Should I raise a PR there?

Yeah, or just make that change in this PR, either way works for me

faisal-alvi:
Thanks for the feedback. Changes are implemented.

> Don't think this is a side-effect of this PR (probably due to changes in https://github.com/10up/classifai/pull/609) but no longer seeing the Classify button if the feature is disabled on the post level

I have found that this is because the post status is being `auto-draft` on the first load and the plugin can not find it in the allowed post types list.

image

So as a fix, we can allow `auto-draft` if someone has allowed the `draft` type; no need to add a new post type in the settings.

includes/Classifai/Providers/Watson/NLU.php
```diff
@@ -145,7 +145,7 @@ public function get_default_settings() {
 	 * Register what we need for the plugin.
 	 */
 	public function register() {
-		if ( $this->has_access( 'content_classification' ) ) {
+		if ( $this->has_access( 'content_classification' ) || isset( $_GET['page'] ) && 'classifai' === $_GET['page'] ) { // phpcs:ignore WordPress.Security.NonceVerification.Recommended
```
dkotter: I'm not sure we'll want to add this conditional here. What this means is that even if the NLU feature is off, we register the NLU taxonomies anytime we're on any ClassifAI settings page (as well as load other functionality that isn't needed).

Unless there's something I'm missing, I think we can just take the same `enqueue_admin_assets` method here and load the correct assets in the Embeddings class, to ensure the assets we need are always loaded on that page. If both features are turned on, that's fine, as WordPress can handle the situation where an asset is enqueued twice.

dkotter (changes requested):

dkotter:
> I have found that this is because the post status is being auto-draft on the first load and the plugin can not find it in the allowed post types list.

I was testing with an already published item so I don't think that's the issue I was facing. In looking closer, it seems we have a conditional in our `gutenberg-plugin.js` file that only loads that button if NLU is enabled. Seems it was set up this way from the beginning and so that button has never shown for the Embeddings feature (which I was mistaken there, I thought I remember it being there but I guess not). I think it would be nice to show that button but that doesn't need to be handled in this PR.

dkotter (approved):

berkod (approved):

PR #626, opened 2023-11-28
Add post autosave for title generation and content resizing.

### LINKED ISSUE #590: Preserve existing data & provide rollbacks

ClassifAI has several features that when triggered (either by a purposed human action or via other publishing/editing workflows) could overwrite existing post content/meta (e.g., post classification, image tags, bulk & CLI ClassifAI interactions).  This focus will look at potentially changing the default handling in these scenarios to not overwrite existing content or perhaps requiring a purposeful human interaction to denote a desire to overwrite existing content (and a filter to be able to change the ClassifAI default handling).  That override could be at a post level, or at the plugin/site level, or perhaps a blend of both.

Similarly, any change that ClassifAI makes to post content/meta will be stored as a new Revision to that post so that a user could utilize the WordPress built-in revisions editor to roll back any unwanted Assistive AI edits to content/meta.

The Post Meta Revisions is relatively achievable, but the Meta and Taxonomy Term Revisions is not quite as clear on feasibility and approach and will require some research before implementation to ensure a stable result.

### Designs

n/a

n/a

### Code of Conduct

iamdharmesh:
### Description of the Change
PR adds autosave functionality for post title generation and content resizing when the post is not dirty (meaning it doesn't contain any other changes). This addition aims to keep a record of every small change made by classifAI, and the editor can restore previous post content by utilizing WordPress core's revision functionality.

Note: Autosave has not been added to excerpt generation yet, as it directly replaces the post excerpt before manual human review.

Closes #590 

### How to test the Change
1. Create/Edit any post.
1. Click "Generate titles" without making any changes to the post.
1. Verify that the post title value is updated and the post is also saved.
1. Make changes in post content.
1. Click "Generate titles".
1. Verify that it only updates the title value without saving the post.
1. Verify this with the classic editor.
1. Test the same with Resizing content.

### Credits
Props @jeffpaul @iamdharmesh 

### Checklist:

### DIFF
--- a/src/js/gutenberg-plugins/content-resizing-plugin.js
+++ b/src/js/gutenberg-plugins/content-resizing-plugin.js
@@ -222,7 +222,11 @@ const ContentResizingPlugin = () => {
 	 *
 	 * @param {string} updateWith The content that will be used to replace the selection.
 	 */
-	function updateContent( updateWith ) {
+	async function updateContent( updateWith ) {
+		const isDirty = await select( 'core/editor' ).isEditedPostDirty();
+		const postId = select( 'core/editor' ).getCurrentPostId();
+		const postType = select( 'core/editor' ).getCurrentPostType();
+
 		dispatch( blockEditorStore ).updateBlockAttributes(
 			selectedBlock.clientId,
 			{
@@ -237,6 +241,15 @@ const ContentResizingPlugin = () => {
 			updateWith.length
 		);
 		resetStates();
+
+		// If no edited values in post trigger save.
+		if ( ! isDirty ) {
+			await dispatch( 'core' ).saveEditedEntityRecord(
+				'postType',
+				postType,
+				postId
+			);
+		}
 	}
 
 	// We don't want to use the reszing feature when multiple blocks are selected.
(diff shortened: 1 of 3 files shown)


src/js/gutenberg-plugins/content-resizing-plugin.js
```diff
@@ -237,6 +241,15 @@ const ContentResizingPlugin = () => {
 			updateWith.length
 		);
 		resetStates();
+
+		// If no edited values in post trigger save.
+		if ( ! isDirty ) {
+			await dispatch( 'core' ).saveEditedEntityRecord(
```
dkotter, line 247: Does this trigger an autosave or a normal save? Or maybe in Gutenberg it's the same thing? I'm seeing the `saveEntityRecord` function has an `isAutosave` option but don't see the same in this function, so just want to ensure we're using the right thing here

iamdharmesh: > Does this trigger an autosave or a normal save?

@dkotter this is a normal save. As Gutenberg provides information on whether the post is dirty or not, we can make sure that the post doesn't contain any other changes than our changes, we are good to trigger the normal save here. Please let me know if you think we should trigger autosave here instead of the normal save. 

> Or maybe in Gutenberg it's the same thing?

Normal save is different from autosave in Gutenberg as well, like the classic editor.

dkotter (approved):

berkod (approved):

PR #630, opened 2023-12-04
Fix | Infinite loop when re-scanning image

### LINKED ISSUE #623: Infinite loop when re-scanning image
### Describe the bug

`wp_update_post()` is called inside the `edit_attachment` hook which causes an infinite loop.

https://github.com/10up/classifai/blob/e9198e508d7bf961fe9b6d407828470160e88670/includes/Classifai/Providers/Azure/ComputerVision.php#L804-L811

### Steps to Reproduce

1. Enable **Generate descriptive text** under Computer Vision
2. Go to the Edit page of any image attachment
3. Check the **No descriptive text? Rescan image** checkbox and click update
4. Observe that the function `protected function generate_alt_tags( $captions, $attachment_id )` triggers multiple times.

The following would fix it:

```php
$excerpt = get_the_excerpt( $attachment_id );

if ( in_array( 'caption', $enabled_fields, true ) && $excerpt !== $captions[0]->text ) {}
```

_No response_

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

sksaju:
### Description of the Change

Closes https://github.com/10up/classifai/issues/623

### Changelog Entry

> Added - New feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

### Credits

### Checklist:

### DIFF
--- a/includes/Classifai/Providers/Azure/ComputerVision.php
+++ b/includes/Classifai/Providers/Azure/ComputerVision.php
@@ -801,7 +801,9 @@ class ComputerVision extends Provider {
 					update_post_meta( $attachment_id, '_wp_attachment_image_alt', $captions[0]->text );
 				}
 
-				if ( in_array( 'caption', $enabled_fields, true ) ) {
+				$excerpt = get_the_excerpt( $attachment_id );
+
+				if ( in_array( 'caption', $enabled_fields, true ) && $excerpt !== $captions[0]->text ) {
 					wp_update_post(
 						array(
 							'ID'           => $attachment_id,


includes/Classifai/Providers/Azure/ComputerVision.php
```diff
@@ -801,7 +801,9 @@ protected function generate_alt_tags( $captions, $attachment_id ) {
 					update_post_meta( $attachment_id, '_wp_attachment_image_alt', $captions[0]->text );
 				}
 
-				if ( in_array( 'caption', $enabled_fields, true ) ) {
+				$excerpt = get_the_excerpt( $attachment_id );
+
+				if ( in_array( 'caption', $enabled_fields, true ) && $excerpt !== $captions[0]->text ) {
```
dkotter, line 806: I don't think this truly fixes the issue, though I didn't test. But in the scenario where an image doesn't have an excerpt set and you trigger the generation of one, this check will return `false` (which is correct) and `wp_update_post` will be called, which then triggers this function to be called again. The second time it runs this code check will return `true` and so things won't execute again but ideally we should never execute this code again in the first place.

I'd suggest instead of doing this we remove the hook if we make it to the `wp_update_post` part, so it doesn't trigger again in that execution:

`remove_action( 'edit_attachment', [ $this, 'maybe_rescan_image' ] );`

dkotter: In addition to the above, I think there are a few other places where we have the same issue (again, just looking at the code, didn't test this).

I see a `wp_update_post` call just a few lines below this (line 814) and then our OCR processing I think will have the same issue as we call `wp_update_post` in our `generate_ocr_data` method, which can be triggered by the `edit_attachment` hook

dkotter (changes requested):

dkotter:
Thanks for all the work here @sksaju! Due to the major refactoring we just merged in, I'm going to close this PR out in favor of #677.

PR #641, opened 2023-12-14
Update `10up/phpcs-composer` to version 3 and fix all newly flagged issues

dkotter:
### Description of the Change

We utilize the `10up/phpcs-composer` package and it recently pushed out a major update to 3.0.0. This update brings in major updates from multiple other packages, including much better support for PHP 8+. As such we should be using this in our projects whenever possible.

In making that update, lots of new issues were getting flagged which I've fixed in this PR. Majority of these are very minor things, like needing an extra space or deleting an extra space. But I know these sorts of fixes have caused problems in the past so definitely need a thorough eye on things here.

### How to test the Change

1. Ensure all changes look correct
2. A thorough top-to-bottom testing of the plugin to ensure things still work as expected

### Changelog Entry

> Changed - Update `10up/phpcs-composer` to version 3.0.0 
> Fixed - Address all new PHPCS issues

### Credits

Props @dkotter

### Checklist:

### DIFF
--- a/autoload.php
+++ b/autoload.php
@@ -58,23 +58,23 @@ class Psr4AutoloaderClass {
 	/**
 	 * Loads the class file for a given class name.
 	 *
-	 * @param string $class The fully-qualified class name.
+	 * @param string $classname The fully-qualified class name.
 	 * @return mixed The mapped file name on success, or boolean false on
 	 * failure.
 	 */
-	public function load_class( $class ) {
+	public function load_class( $classname ) {
 		// the current namespace prefix
-		$prefix = $class;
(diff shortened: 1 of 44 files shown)


faisal-alvi (approved):
@dkotter thanks for the PR. LGTM.

Tested features:

- ✅ Tagging
- ✅ Excerpt Generation
- ✅ Audio Transcripts
- ✅ Title Generation
- ✅ Expand or Condense Text
- ✅ Text to Speech
- ✅ Image Alt Text
- ✅ Image Tagging
- ✅ Generate Images

PR #654, opened 2024-01-09
Switch from the deprecated Completions API to the Models API

dkotter:
### Description of the Change

We currently make a request to the [Completions API](https://platform.openai.com/docs/api-reference/completions) when OpenAI settings are saved to verify the credentials are valid. This API is deprecated and was shut down January 4th. This PR switches from using that API to using the [Models API](https://platform.openai.com/docs/api-reference/models) for the same purpose.

### How to test the Change

1. With the current `develop` branch checked out, enter in valid credentials on the OpenAI ChatGPT settings page
2. Save the page and notice a warning shows about a deprecated model
3. Checkout this branch and save the settings again
4. Notice a success message is shown

### Changelog Entry

> Changed - Switch from using the Completions API to the Models API to verify credentials.

### Credits

Props @dkotter, @jeffpaul 

### Checklist:

### DIFF
--- a/includes/Classifai/Providers/OpenAI/OpenAI.php
+++ b/includes/Classifai/Providers/OpenAI/OpenAI.php
@@ -13,11 +13,11 @@ use function Classifai\get_all_post_statuses;
 trait OpenAI {
 
 	/**
-	 * OpenAI completions URL
+	 * OpenAI model URL
 	 *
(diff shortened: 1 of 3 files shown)


iamdharmesh (approved):
Thanks for fixing this @dkotter. LGTM.

PR #677, opened 2024-01-31
Ensure we don't run updates more than once

### LINKED ISSUE #623: Infinite loop when re-scanning image

dkotter:
### Description of the Change

As mentioned in #623, there's a scenario where manually triggering some of the Azure Image Processing Features from the attachment edit screen can result in an infinite loop.

As part of the refactoring done in #611, this was fixed so that we wouldn't ever encounter an infinite loop but we still end up running those update functions twice, which is not ideal.

This PR fixes that by ensuring we remove the hooked action before we update the attachment.

Closes #623 

### How to test the Change

1. Set up the Descriptive Text Generator Feature
2. Set it up to save the result to either Image caption or Image description (Alt text never had an issue)
3. On an image that has already been uploaded, go to the attachment edit screen
4. Check the box in the ClassifAI Image Processing metabox and save the attachment
5. Ensure the descriptive text is saved

### Changelog Entry

> Fixed - Ensure we only run image updates once.

### Credits

Props @dkotter, @Sidsector9, @sksaju 

### Checklist:

### DIFF
--- a/includes/Classifai/Features/DescriptiveTextGenerator.php
+++ b/includes/Classifai/Features/DescriptiveTextGenerator.php
@@ -263,6 +263,8 @@ class DescriptiveTextGenerator extends Feature {
 			$result = $this->run( $attachment_id, 'descriptive_text' );
 
 			if ( $result && ! is_wp_error( $result ) ) {
+				// Ensure we don't re-run this when the attachment is updated.
+				remove_action( 'edit_attachment', [ $this, 'maybe_rescan_image' ] );
 				$this->save( $result, $attachment_id );
 			}
 		}
--- a/includes/Classifai/Features/ImageCropping.php
+++ b/includes/Classifai/Features/ImageCropping.php
@@ -295,10 +295,8 @@ class ImageCropping extends Feature {
 	 * @param int $attachment_id Attachment ID.
 	 */
 	public function maybe_crop_image( int $attachment_id ) {
-		$metadata = wp_get_attachment_metadata( $attachment_id );
-
 		if ( clean_input( 'rescan-smart-crop' ) ) {
-			$result = $this->run( $attachment_id, 'crop', $metadata );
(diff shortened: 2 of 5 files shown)


Sidsector9 (changes requested):
The code looks good but I'm seeing fatal errors for the following:

### Regenerate smart thumbnail

Screenshot 2024-02-01 at 1 13 07 PM


### Rescan for text

Screenshot 2024-02-01 at 1 12 47 PM

dkotter:
> The code looks good but I'm seeing fatal errors for the following:

Hmm.. do you have any specific steps you're taking here? I'm not able to reproduce the problem but in looking at the error message, it seems the image metadata is returning as a `string` instead of an `array` (which we use `wp_get_attachment_metadata` to get that and supposedly it always returns either an `array` or `false`).

I've added an additional check to ensure we don't proceed if the metadata isn't an array but would still be interested to know how you're able to reproduce this.

Sidsector9 (approved):
Thanks for fixing this issue. Tests well 👍

PR #698, opened 2024-02-06
Show deprecation message on the Recommended Content feature

### LINKED ISSUE #672: Deprecate the Recommended Content Feature

The current Recommended Content Feature is supported by Azure as the only Provider. But as of September of last year, [Azure no longer allows you](https://learn.microsoft.com/en-us/azure/ai-services/personalizer/) to create new Personalizer resources. This means anyone new to ClassifAI can't use this Feature at all and I imagine most existing users didn't have this setup prior to that September 2023 date and thus can't use it either.

Discussion was had around if we should just remove this Feature all together or remove it add put it in a separate add-on plugin but we do hope to find a new Provider to power this Feature so for now, we've left it in place, with the goal to add a new Provider and potentially remove Azure at that point.

But because this Feature isn't usable by most, we want to do a soft-deprecation here. What I'm thinking on this is adding a message to this Feature that let's users know they won't be able to use this if they hadn't created a Personalizer resource prior to September 2023. But also let them know that a new Provider will be coming in the future.

### Designs

_No response_

_No response_

### Code of Conduct

dkotter:
### Description of the Change

Currently the only Provider for the Recommended Content Feature is Azure AI Personalizer. Microsoft announced that no Personalizer resources can no longer be created and the full resource is being retired in 2026.

We hope to find a replacement Provider for this Feature so for now, instead of completely removing this Provider and Feature, we're leaving things in place but showing a message, letting users know this Feature won't work unless they had previously created a Personalizer resource.

Closes #672 

Recommended Content deprecation notice

### How to test the Change

Go to the Recommendation Service settings page and see the notice. Go to other settings pages and make sure the message doesn't show

### Changelog Entry

> Deprecated - Show deprecation notice for the Azure AI Personalizer provider, as this will be removed in a coming release

### Credits

Props @dkotter, @jeffpaul 

### Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ Tap into leading cloud-based services like [OpenAI](https://openai.com/), [Micro
 * Moderate incoming comments for sensitive content using [OpenAI's Moderation API](https://platform.openai.com/docs/guides/moderation)
 * Convert text content into audio and output a "read-to-me" feature on the front-end to play this audio using [Microsoft Azure's Text to Speech API](https://learn.microsoft.com/en-us/azure/cognitive-services/speech-service/text-to-speech)
 * Classify post content using [IBM Watson's Natural Language Understanding API](https://www.ibm.com/watson/services/natural-language-understanding/) and [OpenAI's Embedding API](https://platform.openai.com/docs/guides/embeddings)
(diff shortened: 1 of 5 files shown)


jeffpaul (commented):
Perhaps linking on "As of September 23" to https://learn.microsoft.com/en-us/azure/ai-services/personalizer/ and adding a (see issue#392) and linking to https://github.com/10up/classifai/issues/392 at the end of the message will help give folks a bit more context?

iamdharmesh (approved):
LGTM 🚀

PR #709, opened 2024-02-09
Fully merge the Watson and Embeddings classification features

### LINKED ISSUE #671: Bring more parity between the Watson NLU and OpenAI Embeddings classification providers

In the refactoring done in #611, we merged the Watson NLU and OpenAI Embeddings features into a single Feature, where you can now choose the Provider you want to use.

This all seems to be working but most of the functionality for this Feature exists within each individual Provider. Ideally that functionality should be moved into the Feature and as part of that process, ensure that things function mostly the same irregardless of which Provider is selected.

For instance, if you choose Watson as the Provider, there's a fairly new Manual Review mode. But this isn't something that is available if OpenAI is chosen.

Ultimately there are differences in how these two Providers work but we should work to bring these as close to parity as possible and in that process, move most of the code from the individual Providers into the Feature.

### Designs

_No response_

_No response_

### Code of Conduct

### LINKED ISSUE #699: Moderation > Getting a fatal error on the edit screen of users comments on posts
### Describe the bug

Moderation > Getting a fatal error on the edit screen of users comments on posts

### Steps to Reproduce

1. Enabled Moderation feature of ClassifAI plugin
2. Add/edit the post > Enable post comments
3. Add few comment on post > click on edit button from comment
4. On edit comment page getting fatal error. 

SS:
image

image

### Environment information

Local

### WordPress information

Wordpress 6.4.3

### Code of Conduct

### LINKED ISSUE #714: Cursor Pointer Not Functional for Taxonomy Label on IBM Watson NLU Screen
### Describe the bug

Users encounter an issue where they are unable to enable taxonomy after clicking the 'enable' checkbox label when using IBM Watson NLU as a provider. This functionality works as expected for the OpenAI Embeddings provider screen.

### Steps to Reproduce

1. Follow the onboarding instructions.
2. Enable 'Classification' from the 'Set up ClassifAI to meet your needs' screen.
3. Click the Start Setup button.
4. Register the ClassifAI account.
5. Select 'PROVIDER' from the dropdown.
6. Try to enable texonomy. 

Branch- `Develop` 

Please watch this screencast to reproduce it easily- 

https://github.com/10up/classifai/assets/6384299/a266e514-908e-4219-b757-b6911ac0604f

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

dkotter:
### Description of the Change

Historically (though I use that term loosely since the Embeddings functionality hasn't been around that long) we've had two classification Features:

1. Analyze your content and suggest new terms using the Watson API
2. Analyze your content and existing terms, suggesting which of the existing terms are the best match using the OpenAI Embeddings API

Coming out of the work in #611, these two separate features were merged down into a single Feature, with Watson and Embeddings being two different providers for that Feature.

But we left most of the functionality in each individual Provider class instead of bringing that into the Feature class. These features are very similar but have diverged a bit over time so there wasn't an easy, clean merge that could be done.

The ultimate goal was to merge those two together and bring parity between the two and that is what this PR does. All shared functionality has been moved from the individual Provider classes into the Classification Feature class. Any functionality that was missing between the two (like the modal popup we added to Watson but never to Embeddings) has been taken care of. In theory, the Classification Feature should now function basically the same no matter the Provider that is selected. 

Closes #671
Closes #699 
Closes #714 

### How to test the Change

I would suggest fully testing the Classification Feature with both Providers. This could involve the following:

- Ensure classifying content works in the Classic Editor, both with the checkbox checked and clicking the classify button
- Ensure classifying content works in the Block Editor, both with the checkbox checked and clicking the classify button
- Ensure classifying an individual item works from the post list screen by clicking the inline link
- Ensure classifying one or more items works from the post list screen by using the bulk actions dropdown

### Changelog Entry

> Added - Ability to preview which terms will be added before they get added when using the Embeddings provider.
> Changed - Ensure the Classification functionality works the same for all Providers.
> Fixed - Fix fatal error if both the Classification Feature and Moderation Features are active.

### Credits

Props @dkotter

### Checklist:


includes/Classifai/Admin/BulkActions.php
```diff
@@ -283,6 +305,14 @@ public function register_language_processing_row_action( array $actions, \WP_Pos
 				continue;
 			}
 
+			if ( ! in_array( get_post_type(), $feature->get_supported_post_types(), true ) ) {
+				continue;
+			}
+
+			if ( ! in_array( get_post_status(), $feature->get_supported_post_statuses(), true ) ) {
+				continue;
+			}
```
iamdharmesh: I think this should be limited to classification feature only as we don't have post status selection in other features like Excerpt generation, TTS etc...

```suggestion
			if ( Classification::ID === $feature::ID && ! in_array( get_post_status(), $feature->get_supported_post_statuses(), true ) ) {
				continue;
			}
```

includes/Classifai/Features/Classification.php
```diff
@@ -112,46 +852,158 @@ public function get_feature_default_settings(): array {
...
+		<?php
+	}
+
+	/**
+	 * Return the list of supported taxonomies
+	 *
+	 * @return array
+	 */
+	public function get_supported_taxonomies(): array {
+		$taxonomies = get_taxonomies( [], 'objects' );
```
iamdharmesh, line 1000: Currently, when we select embedding as a provider, it shows. Image Tag and Format taxonomy as well. I think we should filter it.

```suggestion
		$taxonomies = get_taxonomies( [ 'show_ui' => true ], 'objects' );
```

iamdharmesh (commented):
@dkotter, Thanks a lot for merging this. The code LGTM and it tests well. However, I have added two minor suggestions; otherwise, everything looks good.

Also, I have noticed that when we use embeddings as a provider, it also suggests the 'Uncategorized' category as a matching term. I'm not sure if this happens in my setup only. Could you please check once? If it suggests 'Uncategorized,' then we might want to filter it from the suggestions list.

includes/Classifai/Admin/BulkActions.php
dkotter: Fixed [b2cde57](https://github.com/10up/classifai/pull/709/commits/b2cde57e60f8fb1e47e8a9aed2fe132d650a6ac9)

includes/Classifai/Features/Classification.php
dkotter: This is fixed in [b0b3146](https://github.com/10up/classifai/pull/709/commits/b0b3146045bd5bfc6eca728f7f0d221aa1c64e6b). I not only added the `show_ui` argument but also get the value of the post types that are enabled and ensure the taxonomies we show are registered to those post types. This ensures the Image Tag taxonomy, for instance, doesn't show up, as that is only registered to attachments

dkotter: So I actually ended up changing this because of how we register our custom Watson taxonomies. These are only registered if they are chosen as a setting and so `show_ui` is false until that happens. This means if we set `show_ui` to true, those taxonomies will never show up.

Also found a bug that if we pass in the supported post types to our `get_taxonomies` call, it requires all post types be set for that taxonomy. So if someone were to select both Post and Page, categories and tags no longer show up due to not being supported by Pages. I've modified this a bit to where we check post type support after the query, while we loop through our taxonomies

dkotter:
> Also, I have noticed that when we use embeddings as a provider, it also suggests the 'Uncategorized' category as a matching term. I'm not sure if this happens in my setup only. Could you please check once? If it suggests 'Uncategorized,' then we might want to filter it from the suggestions list

Saw the same thing. This should now be excluded

includes/Classifai/Features/Classification.php
iamdharmesh: This looks correct now. However, Post Format taxonomy was still showing in the settings So, I have explicitly excluded it in https://github.com/10up/classifai/pull/709/commits/420d2f051554b0d3b57b8163fa41d3c1c58579eb

Thank you.

iamdharmesh (approved):
LGTM 🚀 

@dkotter I have added 2 minor commits to fix #714 and Exclude the "Post Format" taxonomy from the supported taxonomies list. This is ready to merge now. Thank you.

PR #711, opened 2024-02-12
Add utility to migrate settings

### LINKED ISSUE #670: Add upgrade migration routine for settings

Coming out of the large refactoring work done in #611, some changes were made to how settings are saved.

I believe all settings are now stored under new meta keys and any `Provider` specific settings (like API keys) are now stored under a new array key within those settings.

What this means is that any current ClassifAI users, when they upgrade to the v3 release, all of their settings will be lost. Ideally we should add an upgrade routine that migrates all existing settings into the new structure and probably delete the old settings once done.

### Designs

_No response_

_No response_

### Code of Conduct

Sidsector9:
### Description of the Change
PR adds migration script to migrate existing configured settings to new feature first settings introduced in classifAI v3.

Closes #670

### How to test the Change
1. Clone this repo
2. Checkout to `trunk` branch, and build the plugin files
3. Activate the plugin and configure all of the AI providers.
4. Checkout to this PR branch. `upkeep/670`
5. Verify that you see the notice about ClassifiAI settings migrated and changes in v3
6. Go to Tools > ClassifAI and make sure all settings migrated properly.
7. Validate each features are working properly.

### Changelog Entry

> Added - Tool to migrate feature settings from versions < 3.0.0

### Credits

Props @dkotter @Sidsector9 @iamdharmesh 

### Checklist:


iamdharmesh:
**Note:** E2E tests are failing due to some [issue](https://github.com/WordPress/gutenberg/issues/59232#issuecomment-1956551889) in `wp-env`.

qasumitbagthariya:
### QA Update ⚠️
-------------

@iamdharmesh Thanks for the PR.

I have checked migration with branch `upkeep/670` and I'm getting some issue with migration.

I tried the same steps mentioned in the PR description and noticed that sometimes the migration is not working as expected.

Please check the video below, and let me know if you need any help to reproduce the issue.

Issue video:
https://drive.google.com/file/d/1YTHY4U0qRHKRnqH6kcpLnGdLbQo_3WxV/view?usp=drive_link

includes/Classifai/Admin/Notifications.php
```diff
@@ -212,4 +213,42 @@ public function thresholds_update_notice() {
...
+			return;
+		}
+		?>
+		<div class="notice notice-info is-dismissible classifai-dismissible-notice classifai-migation-notice" data-notice="<?php echo esc_attr( $key ); ?>">
+			<p>
+				<?php
+				echo wp_kses_post(
+					sprintf(
+						// translators: %1$s: <a> tag starting; %2$s: <a> tag closing.
+						__( '%1$sClassifAI 3.0.0%2$s has changed how AI service providers are integrated with AI features which impacted how settings were handled.  Those settings have been migrated automatically, and you can %3$sview them here%4$s.', 'classifai' ),
```
dkotter:
```suggestion
						__( '%1$sClassifAI 3.0.0%2$s has changed how AI providers are integrated with individual features. This changes how settings are stored and requires that existing settings be migrated. This migration has happened automatically and you can %3$sverify settings here%4$s.', 'classifai' ),
```

includes/Classifai/Plugin.php
```diff
@@ -232,4 +233,84 @@ public function filter_plugin_action_links( $links ): array {
...
+			$feature_id       = $feature_instance->get_option_name();
+
+			if ( method_exists( $feature_instance, 'migrate_settings' ) ) {
+				$migrated_settings = $feature_instance->migrate_settings();
+				update_option( $feature_id, $migrated_settings );
+			}
+		}
+
+		// Mark the migration as completed.
+		update_option( 'classifai_v3_migration_completed', true );
```
dkotter: Do we need these options autoloaded across the entire site? Wondering if we can be a little more performant here and set the autoload option to `false` on these?

includes/Classifai/Plugin.php
```diff
@@ -37,6 +37,7 @@ public function enable() {
 		add_action( 'init', [ $this, 'i18n' ] );
 		add_action( 'admin_init', [ $this, 'init_admin_helpers' ] );
 		add_action( 'admin_init', [ $this, 'add_privacy_policy_content' ] );
+		add_action( 'admin_init', [ $this, 'maybe_migrate_to_v3' ] );
```
dkotter, line 40: Have we explored running this on the WordPress update hook, instead of on `admin_init`? Maybe it doesn't matter but we could make this only run after an update has happened (so after someone updates ClassifAI)

dkotter: Tested things out and seems to work fine. I do think it's cleaner to have this only run on the update hook but this is fine

dkotter (approved):
Tested this and all seemed to work fine for me. There is potential for some confusion as we've rearranged some settings and changed how the role and user based access works (no longer showing extra checkboxes) but that all looks fine to me and the main things (like API keys) all come over correctly

dkotter:
One other thing to consider is we aren't deleting any of the old settings that are stored in options. I think that's okay for now, as it allows users to access that if something does go wrong in the migration process. But we probably want to look at cleaning that up in a release or two.

PR #717, opened 2024-02-14
Upgrade from DALL·E 2 to DALL·E 3

dkotter:
### Description of the Change

We currently utilize DALL·E 2 in our Image Generation Feature. OpenAI released DALL·E 3 to the public last fall and so this PR updates our integration to use DALL·E 3.

I had originally considered having an option to choose between the two models, as DALL·E 2 is still available and is cheaper but there are new options that only apply to DALL·E 3, making it harder to support both, and the quality of generated images is so much better with DALL·E 3 that it seemed the right idea to only support that.

### How to test the Change

1. Go to `Tools > ClassifAI > Image Processing > Image Generation`
2. Turn on the Feature and enter a valid API key
3. Change any other settings as desired
4. Go to the stand-alone image generation page by going to `Media > Generate Image`
5. Enter in a prompt and ensure images are generated
6. Create a new post and add an Image block. Go to the Media Library of the Image block and go to the Generate Images tab
7. Enter in a prompt and ensure images are generated

### Changelog Entry

> Changed - Upgrade from DALL·E 2 to DALL·E 3 in our Image Generation Feature 

### Credits

Props @dkotter, @jeffpaul 

### Checklist:


jeffpaul (commented):
Would be helpful in the setup/settings display of the `Image size` options to perhaps add a parenthetical descriptor for each whether the dimensions are (square), (landscape), or (portrait) to help clarify how each size will effectively render.

I also think there's a small enhancement on the `Generate images` modal to include the `Number of images`, `Image quality`, `Image size`, and `Image style` with each preset to its site-wide default but allow someone to tweak per that specific image generation request?

dkotter:
> Would be helpful in the setup/settings display of the `Image size` options to perhaps add a parenthetical descriptor for each whether the dimensions are (square), (landscape), or (portrait) to help clarify how each size will effectively render.

Good idea, I've added this

> I also think there's a small enhancement on the Generate images modal to include the Number of images, Image quality, Image size, and Image style with each preset to its site-wide default but allow someone to tweak per that specific image generation request?

One thing to consider is that the number of images, image quality and image size are all options that impact the price, so I think we would need the ability to not allow those to be changed at the Image Generation level. But yes, I agree that having the option to show those would be a nice enhancement though may be better handled in a follow up PR.

iamdharmesh (approved):
🚀

PR #733, opened 2024-03-02
Ensure both arguments passed to `merge_settings` are an `array`

### LINKED ISSUE #732: Update from 2.5.1 to 3.0.0 broke my website
### Describe your question

I updated the plugin from my current version 2.5.1 to 3.0.0 and it broke the website. I had to disable the plugin to recover my website and then return to the old version.
WP threw the following errors:

```
[01-Mar-2024 21:47:51 UTC] PHP Fatal error:  Uncaught TypeError: Argument 1 passed to Classifai\Features\Feature::merge_settings() must be of the type array, string given, called in C:\laragon\www\mysite\wp-content\plugins\classifai-stable\includes\Classifai\Features\Feature.php on line 399 and defined in C:\laragon\www\mysite\wp-content\plugins\classifai-stable\includes\Classifai\Features\Feature.php:394
Stack trace:
#0 C:\laragon\www\mysite\wp-content\plugins\classifai-stable\includes\Classifai\Features\Feature.php(399): Classifai\Features\Feature->merge_settings('1', Array)
#1 C:\laragon\www\mysite\wp-content\plugins\classifai-stable\includes\Classifai\Features\Feature.php(319): Classifai\Features\Feature->merge_settings(Array, Array)
#2 C:\laragon\www\mysite\wp-content\plugins\classifai-stable\includes\Classifai\Features\Feature.php(942): Classifai\Features\Feature->get_settings()
#3 C:\laragon\www\mysite\wp-content\plugins\classifai-stable\includes\Classifai\Features\Feature.php(57): Classifai\Features\Feature->is_feature_enabled()
#4 C:\laragon\www\mysite in C:\laragon\www\mysite\wp-content\plugins\classifai-stable\includes\Classifai\Features\Feature.php on line 394
```

### Code of Conduct

dkotter:
### Description of the Change

We have a `merge_settings` method that takes in stored settings as well as the default settings and merges those together. This method expects both of these values to be an `array`, so if non-array data is passed in, a PHP error occurs.

This method is called recursively to merge settings that contain array data. This was throwing a fatal error in #732 because the passed in settings were not an array but the default settings were.

This PR adds additional checks here to ensure we only ever call `merge_settings` if both the stored settings and default settings are an array. For instances were the stored settings aren't an array but the defaults are, we override the stored settings with the defaults as it seems the only time this happens is when data has been stored improperly (or not migrated properly in v3.0.0).

In addition, some followup in #732 pointed to some additional issues in the descriptive text generator feature where it expects array data but is instead passed in non-array data. In theory this should never happen but again, in scenarios where data is stored improperly or migrated wrong, I've added additional checks to avoid PHP errors and warnings.

Closes #732 

### How to test the Change

There are good details in #732 but I've not been able to replicate the problem yet without manually manipulating data.

Here's the steps I followed:

1. Install the 2.5.1 version of ClassifAI
2. Configure settings as seen in the screenshots from #732 
3. Deactivate the plugin and install and activate the 3.0.0 version of the plugin
4. Go to the ClassifAI settings screens and note the settings that have been migrated over and should see no errors
5. Manually change the `classifai_feature_descriptive_text_generator` option in the database. Change the `descriptive_text_fields` item to be the value `1` instead of an array
6. Go back to the settings you should now see the same errors as reported in #732 
7. Check out this branch and refresh, you should no longer see the errors

### Changelog Entry

> Fixed - Ensure we only pass in array data to our `merge_settings` method
> Fixed - Handle scenario where non-array data is passed into our descriptive text generator feature

### Credits

Props @ajaxthemestudios, @dkotter

### Checklist:

### DIFF
--- a/includes/Classifai/Features/DescriptiveTextGenerator.php
+++ b/includes/Classifai/Features/DescriptiveTextGenerator.php
@@ -357,7 +357,7 @@ class DescriptiveTextGenerator extends Feature {
 			[
 				'label_for'      => 'descriptive_text_fields',
 				'options'        => $checkbox_options,
-				'default_values' => is_array( $settings['descriptive_text_fields'] ) ? $settings['descriptive_text_fields'] : [],
+				'default_values' => $settings['descriptive_text_fields'],
 				'description'    => __( 'Choose image fields where the generated text should be applied.', 'classifai' ),
 			]
 		);
--- a/includes/Classifai/Features/DescriptiveTextGenerator.php
+++ b/includes/Classifai/Features/DescriptiveTextGenerator.php
@@ -307,12 +307,13 @@ class DescriptiveTextGenerator extends Feature {
 		$settings       = $this->get_settings();
 		$enabled_fields = array();
 
-		if (
-			! isset( $settings['descriptive_text_fields'] ) ||
-			! is_array( $settings['descriptive_text_fields'] )
-		) {
+		if ( ! isset( $settings['descriptive_text_fields'] ) ) {
+			return array();
+		}
+
+		if ( ! is_array( $settings['descriptive_text_fields'] ) ) {
 			return array(
-				'alt'         => 0,
+				'alt'         => 'no' === $settings['descriptive_text_fields'] ? 0 : 'alt',
 				'caption'     => 0,
 				'description' => 0,
 			);
--- a/includes/Classifai/Features/DescriptiveTextGenerator.php
+++ b/includes/Classifai/Features/DescriptiveTextGenerator.php
@@ -307,13 +307,12 @@ class DescriptiveTextGenerator extends Feature {
 		$settings       = $this->get_settings();
 		$enabled_fields = array();
 
-		if ( ! isset( $settings['descriptive_text_fields'] ) ) {
-			return array();
-		}
-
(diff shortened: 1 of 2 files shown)


dkotter:
I believe I've finally figured out proper reproduction steps here. The ones outlined in the PR description still work (manually changing data) but I'm pretty sure I've figured out that actual use case.

In #374, we changed how the descriptive text generator feature works. Previously it was either on or off and if on, it would save the text in the `alt` text field. In that PR we changed this so you can choose which field(s) this text gets saved to.

But this also means the settings we saved changed from being a `string` to being an `array`. This PR went out in the v2.0.0 release. So if someone were using ClassifAI prior to that release and they never re-saved the descriptive text generator settings, they would still have a `string` value saved there instead of an `array` value. And when updating to v3.0.0 and the changes introduced there (as well as stricter type checks) this ends up causing the problem.

So to reproduce:

1. Install the 1.8.1 version of ClassifAI (last version prior to v2.0.0)
2. Set things up to match the screenshot in #732 but also ensure to turn on the `Generate alt text` setting
3. Update to v3.0.0
4. Notice PHP errors happen
5. Checkout out this PR and ensure the errors go away

peterwilsoncc (approved):
LGTM and tests well.

PR #734, opened 2024-03-04
Amazon Polly as a provider for the text-to-speech feature.

### LINKED ISSUE #728: Add Amazon Polly as a Provider for Text to speech feature.

AWS provides a [Amazon Polly](https://aws.amazon.com/polly/) to turn text into high-quality, natural-sounding human voices. It could be a good alternative provider option to the current provider Azure AI Speech.

### Designs

_No response_

_No response_

### Code of Conduct

iamdharmesh:
### Description of the Change
PR Adds Amazon Polly as a provider for the Text-to-speech feature.

Closes #728 

@jeffpaul @dkotter Amazon Polly provides additional options such as [Newscaster Speaking Style](https://docs.aws.amazon.com/polly/latest/dg/ntts-speakingstyles.html) and [SSML](https://docs.aws.amazon.com/polly/latest/dg/ssml.html), which offers features like including breathing sounds and emphasizing specific words or phrases. I haven't implemented it in this PR, but we can consider integrating it in the future if there are specific client requirements around these features.

@dkotter, I haven't added E2E tests for this because I haven't figured out yet how to mock the API, given that we are using the AWS PHP SDK here. Please let me know if you have any ideas on this.

### How to test the Change
1. Go to Tools > ClassifAI > Language Processing > Text to Speech.
1. Select the "Amazon Polly" option as the provider.
1. Add AWS credentials and save settings.
1. Create/Edit a post and ensure that the Text-to-Speech feature is working as expected.

### Credits
Props @jeffpaul @iamdharmesh 

### Checklist:


includes/Classifai/Providers/AWS/AmazonPolly.php
```diff
@@ -0,0 +1,555 @@
...
+				'aws_polly_user_not_authorized',
+				esc_html__( 'Unauthorized user.', 'classifai' )
+			);
+		}
+
+		$normalizer          = new Normalizer();
+		$feature             = new TextToSpeech();
+		$settings            = $feature->get_settings();
+		$post                = get_post( $post_id );
+		$post_content        = $normalizer->normalize_content( $post->post_content, $post->post_title, $post_id );
```
dkotter, line 381: Looking at the [API docs](https://docs.aws.amazon.com/aws-sdk-php/v3/api/api-polly-2016-06-10.html#synthesizespeech), seems there's a limit on the length of content we can send. Wondering if we should have a check here that we can return early with a proper error message if the content length is too long or if the error that will be returned in the `try/catch` block below is good enough?

dkotter:
> @dkotter, I haven't added E2E tests for this because I haven't figured out yet how to mock the API, given that we are using the AWS PHP SDK here. Please let me know if you have any ideas on this.

I guess my first question would be do we need to use the SDK here? I know that can help simplify things but we haven't used any SDKs for the other Providers up to this point. I'm not opposed to it, just wondering if there was a specific reason.

But I can think of two approaches we can take to mock the requests:

1. Add a short-circuit filter right before we make the request to AWS, allowing us to return our own results. This is basically what WordPress does, my only concern is we'd basically be adding a filter for testing purposes only which I don't love
2. Because the main request goes through a custom REST endpoint, there is a filter there that fires before any callbacks are called: `rest_pre_dispatch`. We could use this and return a hardcoded result, similar to how we're currently using the `pre_http_request` filter. This wouldn't work for all scenarios (like triggering Text to Speech from the inline row action) but should work to test the main use case of publishing content

includes/Classifai/Providers/AWS/AmazonPolly.php
iamdharmesh: Yes, nice catch! However, while checking on this, I found that there is no error handling (in terms of displaying error messages) in place for the text-to-speech feature. I will investigate further on this.

iamdharmesh:
> I guess my first question would be do we need to use the SDK here? I know that can help simplify things but we haven't used any SDKs for the other Providers up to this point. I'm not opposed to it, just wondering if there was a specific reason.

The main reason for using the SDK was to keep things simple, especially concerning [signing and authenticating REST requests](https://docs.aws.amazon.com/AmazonS3/latest/userguide/RESTAuthentication.html). I believe we don't have this complex authentication process with existing providers. I'm open to getting rid of the SDK here and writing a custom class for handling authentication and REST operations (similar to what we did for the OpenAPI). Please let me know if you think we should remove the SDK here.

> I can think of two approaches we can take to mock the requests:

Approach #1 seems like the better choice to me as it allows us to cover all scenarios.

Thanks

dkotter:
> The main reason for using the SDK was to keep things simple, especially concerning [signing and authenticating REST requests](https://docs.aws.amazon.com/AmazonS3/latest/userguide/RESTAuthentication.html). I believe we don't have this complex authentication process with existing providers. I'm open to getting rid of the SDK here and writing a custom class for handling authentication and REST operations (similar to what we did for the OpenAPI). Please let me know if you think we should remove the SDK here.

I think we're fine to proceed with keeping the SDK here. It does increase the size of the final release zip due to all the code the SDK brings but that should be fine.

> Approach #1 seems like the better choice to me as it allows us to cover all scenarios.

That works for me

includes/Classifai/Providers/AWS/AmazonPolly.php
iamdharmesh: @dkotter This is added and PR is not ready for the review.

includes/Classifai/Providers/AWS/AmazonPolly.php
```diff
@@ -0,0 +1,595 @@
...
+			 * Filters the return value of the connect to services function.
+			 *
+			 * Returning a non-false value from the filter will short-circuit the describe voices request and return early with that value.
+			 * This filter is useful for E2E tests.
+			 *
+			 * @since 3.1.0
+			 *
+			 * @param false|mixed $response        A return value of connect to service. Default false.
+			 * @param array       $synthesize_data HTTP request arguments.
+			 */
```
dkotter: We'll want to ensure this docblock matches what JSDoc expects so it ends up in our documentation. See the `classifai_aws_polly_synthesize_speech_args` filter for an example

includes/Classifai/Providers/AWS/AmazonPolly.php
```diff
@@ -0,0 +1,595 @@
...
+			 * Filters the return value of the synthesize speech function.
+			 *
+			 * Returning a non-false value from the filter will short-circuit the synthesize speech request and return early with that value.
+			 * This filter is useful for E2E tests.
+			 *
+			 * @since 3.1.0
+			 *
+			 * @param false|mixed $response        A return value of synthesize speech. Default false.
+			 * @param array       $synthesize_data HTTP request arguments.
+			 */
```
dkotter: We'll want to ensure this docblock matches what JSDoc expects so it ends up in our documentation. See the `classifai_aws_polly_synthesize_speech_args` filter for an example

dkotter (dismissed):

dkotter (approved):

PR #754, opened 2024-04-01
fix/708: fixes issue with processing `.png` images with large file sizes.

### LINKED ISSUE #708: Image processing > When 5 MB png type image is uploaded and selected for processing then getting errors
### Describe the bug

Image processing > When a 5 MB png type image is uploaded and selected for processing then getting errors .

### Steps to Reproduce

1. Go to media upload
2. Upload 5 MB size PNG type image 
3. Try to generate Descriptive text, Image tags, Smart thumbnail, Scan image for text
4. Observe error messages

image (5)

### Environment information

QA Environment - WordPress 6.4.3

### WordPress information

_No response_

### Code of Conduct

Sidsector9:
### Description of the Change

This PR:
- sets the image scanning timeout to 60 seconds.
- adds the filter hook `'classifai_ms_computer_vision_scan_image_timeout'`

Closes #708 

### How to test the Change

This can be tested with the following sample images:

[5mb-png-image](https://github.com/10up/classifai/assets/17757960/a92aa3da-bad5-4e36-977d-945789b84f19)
[10mb-png-image](https://github.com/10up/classifai/assets/17757960/77a38862-8c3a-4a00-8407-86284a64cb41)

### Changelog Entry

> Added - filter hook `'classifai_ms_computer_vision_scan_image_timeout'` to increase image processing timeout.
> Fixed - timeout issue with processing `.png` images of large file sizes.

### Credits

Props @QAharshalkadu @Sidsector9 

### Checklist:

### DIFF
--- a/includes/Classifai/Providers/Azure/ComputerVision.php
+++ b/includes/Classifai/Providers/Azure/ComputerVision.php
@@ -612,9 +612,15 @@ class ComputerVision extends Provider {
 				],
 				/**
 				 * Filters the timeout for the image scan request.
+				 *
 				 * Default: 60 seconds.
 				 *
-				 * @param int $timeout Timeout in seconds.
+				 * @since 3.1.0
+				 * @hook classifai_ms_computer_vision_scan_image_timeout
+				 *
+				 * @param {int} $timeout Timeout in seconds.
+				 *
+				 * @return {int} Timeout in seconds.
 				 */
 				'timeout' => apply_filters(
 					'classifai_' . self::ID . '_scan_image_timeout',
--- a/includes/Classifai/Providers/Azure/ComputerVision.php
+++ b/includes/Classifai/Providers/Azure/ComputerVision.php
@@ -610,6 +610,16 @@ class ComputerVision extends Provider {
 					'Ocp-Apim-Subscription-Key' => $settings['api_key'],
 					'Content-Type'              => 'application/json',
 				],
+				/**
+				 * Filters the timeout for the image scan request.
+				 * Default: 60 seconds.
+				 *
+				 * @param int $timeout Timeout in seconds.
+				 */
+				'timeout' => apply_filters(
+					'classifai_' . self::ID . '_scan_image_timeout',
+					60
+				),
 				'body'    => '{"url":"' . $image_url . '"}',
 			]
 		);


includes/Classifai/Providers/Azure/ComputerVision.php
```diff
@@ -610,6 +610,16 @@ protected function scan_image( string $image_url, \Classifai\Features\Feature $f
 					'Ocp-Apim-Subscription-Key' => $settings['api_key'],
 					'Content-Type'              => 'application/json',
 				],
+				/**
+				 * Filters the timeout for the image scan request.
+				 * Default: 60 seconds.
+				 *
+				 * @param int $timeout Timeout in seconds.
+				 */
```
dkotter: We'll want to ensure this doc block follows the JS Hook standards so it shows properly in our documentation. See the `classifai_should_smart_crop_image` filter in this file for an example

dkotter (changes requested):
One minor change request on the documentation side.

But also a question that may need looked at further. According to the API docs, the max image supported is 4MB. We have some [code in place](https://github.com/10up/classifai/blob/fe28ee7c7971fecdc4de246a5b0beb8b4baf2747/includes/Classifai/Providers/Azure/ComputerVision.php#L742) to return an error if the image size is larger than that so in theory, a 5MB image shouldn't even work, though it sounds like that's not the case.

Some time should be spent here to figure out how images larger than 4MB are making it through and making any needed adjustments. I know we check the file size in the `rest_endpoint_callback` method but maybe that check needs added other places as well?

Sidsector9:
> According to the API docs, the max image supported is 4MB. We have some [code in place](https://github.com/10up/classifai/blob/fe28ee7c7971fecdc4de246a5b0beb8b4baf2747/includes/Classifai/Providers/Azure/ComputerVision.php#L742) to return an error if the image size is larger than that so in theory, a 5MB image shouldn't even work, though it sounds like that's not the case.

@dkotter what you're saying is correct and the check we have is working as expected. When we upload an image size > 4MB, the check selects the largest image variant that is < 4MB to process. 

This particular issue is a weird one pertaining to just the `.png` format. From what I am understanding, the AI service takes more time to process a `.png` compared to a `.jpeg`. So, even with a `.png` file which has a size of 3.5MB, it still fails.

This works well when the timeout is increased.

dkotter (approved):

PR #760, opened 2024-04-16
Update the `@10up/cypress-wp-utils` package to 0.3.0

dkotter:
### Description of the Change

The [v0.3.0](https://github.com/10up/cypress-wp-utils/releases/tag/0.3.0) release of the `@10up/cypress-wp-utils` is now out so this PR upgrades to that and removes a couple custom commands we were using and instead uses the commands that are in that package.

### How to test the Change

Ensure E2E tests pass on this PR

### Changelog Entry

> Changed - Update the `@10up/cypress-wp-utils` package to 0.3.0

### Credits

Props @dkotter

### Checklist:


jeffpaul (approved):
Changes look good, just waiting for tests to pass

PR #770, opened 2024-05-17
Conditional Display of Helper Text for AI Provider Account Creation

### LINKED ISSUE #715: Conditional Display of Helper Text for AI Provider Account Creation
### Describe the bug

**Description:** 
The helper text, such as 'Don't have an IBM Cloud account yet? Register for one and set up a Natural Language Understanding Resource to get your API key.', should be displayed conditionally. It should only appear when the credentials fields are empty or contain incorrect credentials. 

**Use Case:**
When the service is successfully configured, it implies that the admin already has an account, allowing them to set up the service correctly. Therefore, if the admin has already configured an account, this helper text is not useful on the screen. 

### Steps to Reproduce

**Steps:**

1. Begin setting up the plugin.
2. Enter the correct credentials for the AI provider.
3. Save the settings.

Branch- `develop`

When fields are empty - this text will help the user to create an account and setup a plugin.

image

When Setting is configured successfully, this can be hidden:

image

Example - 

image

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

kmgalanakis:
### Description of the Change

Closes #715

### How to test the Change
* Enable/disable the several features with all the available providers and verify that help texts related to the configuration of each provider are hidden/displayed respectively.

### Changelog Entry

> Changed - Hide the provider's configuration help text when the provider is configured, for all providers.

### Credits

Props @kmgalanakis

### Checklist:

### DIFF
--- a/includes/Classifai/Providers/Watson/NLU.php
+++ b/includes/Classifai/Providers/Watson/NLU.php
@@ -121,7 +121,7 @@ class NLU extends Provider {
 				'input_type'    => 'password',
 				'large'         => true,
 				'class'         => 'classifai-provider-field provider-scope-' . static::ID, // Important to add this.
-				'description'   => $this->feature_instance->is_feature_enabled() ?
+				'description'   => $this->feature_instance->is_configured() ?
 					'' :
 					sprintf(
 						wp_kses(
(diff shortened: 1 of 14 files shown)


includes/Classifai/Providers/Watson/NLU.php
```diff
@@ -121,41 +121,45 @@ public function render_provider_fields() {
...
-					wp_kses(
-						/* translators: %1$s is the link to register for an IBM Cloud account, %2$s is the link to setup the NLU service */
-						__( 'Don\'t have an IBM Cloud account yet? <a title="Register for an IBM Cloud account" href="%1$s">Register for one</a> and set up a <a href="%2$s">Natural Language Understanding</a> Resource to get your API key.', 'classifai' ),
-						[
-							'a' => [
-								'href'  => [],
-								'title' => [],
-							],
-						]
+				'description'   => $this->feature_instance->is_feature_enabled() ?
```
dkotter: So the `is_feature_enabled` method checks three things:

1. Is the Provider configured properly?
2. Does the current user have access to this Feature?
3. Is the Feature turned on?

This is probably fine when someone is setting things up through the onboarding wizard but if changing things from the settings page, this means the helper text won't show if the Feature isn't turned on. This leads to the scenario where the only way to see the helper text is turning the Feature on and then saving the settings.

I think it may be better to use the `is_configured` method, so we aren't running a user check or enabled check but just checking if the Provider is configured.

includes/Classifai/Providers/Watson/NLU.php
```diff
@@ -121,41 +121,45 @@ public function render_provider_fields() {
...
-						: esc_html__( 'Use an API Key instead?', 'classifai' ) // phpcs:ignore WordPress.Security.EscapeOutput.OutputNotEscaped
-				);
-			},
-			$this->feature_instance->get_option_name(),
-			$this->feature_instance->get_option_name() . '_section',
-			[
-				'class' => 'classifai-provider-field hidden provider-scope-' . static::ID, // Important to add this.
-			]
-		);
+		if ( ! $this->feature_instance->is_feature_enabled() ) {
```
dkotter: I don't think we want to hide this toggle as someone may want to switch from using an API key to using a Username and Password, which this makes that impossible unless they remove credentials first or disable the Feature first.

dkotter (changes requested):
Left a few comments but also noting this only addresses the helper text for the IBM Watson Provider. Most of our other Providers also show similar helper text so probably best to tackle all of those in a single PR

includes/Classifai/Providers/Watson/NLU.php
kmgalanakis: Good point @dkotter, thank you. This should be fine now.

includes/Classifai/Providers/Watson/NLU.php
kmgalanakis: Hello @dkotter.

Good point. However, instead of just checking if the feature is configured I preferred to check if the feature is configured with the specified provider. The problem was that when using just `is_configured` the feature might have been configured with provider A and that would result in hiding the help texts of provider B for the same service. With my approach, the help texts of a provider are only hidden if the feature is configured with the specified provider.

Please let me know what you think.

Thank you

Sidsector9 (approved):
Tested and works as expected 👍

PR #774, opened 2024-05-29
enhance/748: Updates to expand/condense text functionality

### LINKED ISSUE #748: Updates to expand/condense text functionality

See mockups below for some suggested tweaks to the Expand/Condense options in the Paragraph block.  In summary:
- Update modal heading text to specifically reference what action was taken (e.g. expand/condense)
- Update the action button text from `Select` to `Insert` to make it more clear what action will happen when the button is pressed
- Add a "retry" action line at the bottom of the modal to give an option to expand/condense again without having to close the modal + click the AI button in the toolbar + click expand/condense.  Effectively saves several steps in trying the expand/condense option again.
- If we're not already providing feedback to the AI service provider / model, then I'd suggest when someone clicks the `Insert` button that we let the model know that option was selected.  Alternatively if someone closes out of the modal or clicks the retry action link that we let the model know that none of the suggestions were selected.

### Designs

Mockup showing updates to the "Expand this text" option

expand-updates

Mockup showing updates to the "Condense this text" option

condense-updates

Leave as-is, though the changes suggested here have come out of demos I've done over the last ~6 months and appear to (1) help clarify how these modals function, (2) provide an option to retry expand/condense options more quickly, and (3) could help further train models based on 👍🏼 / 👎🏼 feedback.

### Code of Conduct

Sidsector9:
### Description of the Change

1. Added feature to refresh Content Resizing results without closing the results modal.
2. 'Select' label is changed to 'Replace' within Content Resizing suggestions modal.
3. Changed the Content Resizing modal title from static to dynamic depending on the type of resizing.

Closes #748 

### How to test the Change

Please follow the issue.

### Changelog Entry

> Added - Feature to refresh Content Resizing results without closing the results modal.
> Changed - 'Select' label is changed to 'Replace' within Content Resizing suggestions modal.
> Changed - Content Resizing modal title from static to dynamic depending on the type of resizing.

### Credits

Props @jeffpaul @faisal-alvi @Sidsector9 

### Checklist:


jeffpaul:
@Sidsector9 I think @faisal-alvi's suggesting to use `Replace` instead of `Insert` might be better here... mind making that tweak?

src/js/gutenberg-plugins/content-resizing-plugin.js
```diff
@@ -330,6 +360,26 @@ const ContentResizingPlugin = () => {
 					</tbody>
 				</table>
 			</div>
+			<p>
+				{ __(
+					'None of these suggestions are ideal, please',
```
dkotter: I agree with your thought here: https://github.com/10up/classifai/issues/748#issuecomment-2137301539. To me this reads as if something went wrong and we're giving an error message. It also doesn't make sense if someone has it set to only show a single result (`suggestions` vs `suggestion`).

I also wonder if a button makes more sense here, as that would stand out more, especially when the disable functionality is turned on, we already show a link here:

Content resizing modal

Something like this:

Refresh results in modal

I also think `Retry expanding/condensing the text` looks good here, so I'm fine with either text:

Retry button in modal

src/js/gutenberg-plugins/content-resizing-plugin.js
```diff
@@ -142,6 +146,12 @@ const ContentResizingPlugin = () => {
 				await resizeContent();
 			} )();
 		}
+
+		if ( 'grow' === resizingType ) {
+			setModalTitle( __( 'Expanded text suggestions', 'classifai' ) );
+		} else {
+			setModalTitle( __( 'Condensed text suggestions', 'classifai' ) );
```
dkotter: This may be more complicated than what we care about but in the case where someone has it set to only display a single suggestion, `suggestions` doesn't make sense at that point. Very minor thing so may be fine to just leave

dkotter (changes requested):

src/js/gutenberg-plugins/content-resizing-plugin.js
Sidsector9: Added here [6e9884d](https://github.com/10up/classifai/pull/774/commits/6e9884df745ba93cb4e464ede838dee64c95b934)

dkotter (approved):

PR #775, opened 2024-06-05
fix/659: strip off HTML subscript and superscript tags for Text to Speech

### LINKED ISSUE #659: Removing HTML superscript and subscript tags with its contents for text to speech

Screenshot 2024-01-14 at 1 59 31 PM

When the post content has `<sup>` and `<sub>` tags, the tags are stripped but the content, in this case the number `3` is left out. When the text to speech transform is done, it reads **Ancient Egyptians three**, which changes the meaning of the sentence.

This is an opinionated enhancement request and I think it'll be good to strip the content along with these tags. Or at least we can provide a setting to toggle this behaviour. 

### Designs

_No response_

_No response_

### Code of Conduct

CacheMeOwside:
### Description of the Change
Added logic to strip off HTML subscript and superscript tags, along with its content for text-to-speech generation.

Closes #659

### How to test the Change

1. Add or Edit a post.
2. Add text which contains subscript or superscript.
ClassifAI

3. Click on "Generate Audio".
4. Publish or update the post.
5. Preview the generated audio. The subscript and superscript content should be skipped in the audio.

### Changelog Entry

> Added - `classifai_normalize_content_before_strip_all_tags` filter hook.
> Fixed - Stripped off superscript and subscript tags with content for text to speech generation.

### Credits

### Checklist:

### DIFF
--- a/includes/Classifai/Features/TextToSpeech.php
+++ b/includes/Classifai/Features/TextToSpeech.php
@@ -859,11 +859,11 @@ class TextToSpeech extends Feature {
 	 * @return string The normalized post content.
 	 */
 	public function normalize_post_content( int $post_id ): string {
-		add_filter( 'classifai_normalize_content_before_strip_all_tags', [ $this, 'strip_sub_sup_tags' ] );
+		add_filter( 'classifai_pre_normalize', [ $this, 'strip_sub_sup_tags' ] );
 		$normalizer   = new Normalizer();
 		$post         = get_post( $post_id );
 		$post_content = $normalizer->normalize_content( $post->post_content, $post->post_title, $post_id );
-		remove_filter( 'classifai_normalize_content_before_strip_all_tags', [ $this, 'strip_sub_sup_tags' ] );
+		remove_filter( 'classifai_pre_normalize', [ $this, 'strip_sub_sup_tags' ] );
 
 		return $post_content;
 	}
@@ -876,7 +876,7 @@ class TextToSpeech extends Feature {
 	 *
 	 * @return string The filtered post content.
 	 */
-	public function strip_sub_sup_tags( $post_content ) {
+	public function strip_sub_sup_tags( string $post_content ): string {
 		$post_content = preg_replace( '/<sub>.*?<\/sub>|<sup>.*?<\/sup>/', '', $post_content );
 		return $post_content;
 	}
--- a/includes/Classifai/Features/TextToSpeech.php
+++ b/includes/Classifai/Features/TextToSpeech.php
@@ -859,10 +859,17 @@ class TextToSpeech extends Feature {
 	 * @return string The normalized post content.
 	 */
 	public function normalize_post_content( int $post_id ): string {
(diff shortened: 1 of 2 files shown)


includes/Classifai/Normalizer.php
```diff
@@ -49,6 +49,17 @@ public function normalize_content( $post_content, $post_title = '', $post_id = f
...
+		/**
+		 * Hook to filter post content before stripping HTML tags.
+		 *
+		 * @hook classifai_normalize_content_before_strip_all_tags
+		 *
+		 * @param {string} $post_content The post content.
+		 *
+		 * @return {string} The filtered Post content.
+		 */
+		$post_content = apply_filters( 'classifai_normalize_content_before_strip_all_tags', $post_content );
```
dkotter: We have a filter that runs after all normalization happens (`classifai_normalize`). I'd suggest instead of adding this filter prior to tags being stripped, let's add a filter at the very beginning of this method to allow modifying the content before any normalization happens. Maybe `classifai_pre_normalize` could be the filter name?

includes/Classifai/Normalizer.php
```diff
@@ -49,6 +49,17 @@ public function normalize_content( $post_content, $post_title = '', $post_id = f
 		/* Replace HTML linebreaks with newlines */
 		$post_content = preg_replace( '#<br\s?/?>#', "\n\n", $post_content );
 
+		/**
+		 * Hook to filter post content before stripping HTML tags.
+		 *
+		 * @hook classifai_normalize_content_before_strip_all_tags
```
dkotter:
```suggestion
		 *
		 * @since 3.1.0
		 * @hook classifai_normalize_content_before_strip_all_tags
```

includes/Classifai/Features/TextToSpeech.php
```diff
@@ -859,13 +859,28 @@ public function get_audio_generation_subsequent_state( $post = null ): bool {
...
 
+	/**
+	 * Filters the post content by stripping off HTML subscript and superscript tags
+	 * with its content for text to speech generation.
+	 *
+	 * @param string $post_content The post content.
+	 *
+	 * @return string The filtered post content.
+	 */
+	public function strip_sub_sup_tags( $post_content ) {
```
dkotter:
```suggestion
	public function strip_sub_sup_tags( string $post_content ): string {
```

dkotter (changes requested):

includes/Classifai/Normalizer.php
```diff
@@ -43,12 +43,26 @@ public function normalize( $post_id, $post_content = '' ) {
...
+		 */
+		$post_content = apply_filters( 'classifai_pre_normalize', $post_content );
+
 		/* Strip HTML entities */
 		$post_content = preg_replace( '/&#?[a-z0-9]{2,8};/i', '', $post_content );
 
 		/* Replace HTML linebreaks with newlines */
 		$post_content = preg_replace( '#<br\s?/?>#', "\n\n", $post_content );
 
+
```
dkotter:
```suggestion
```

dkotter (approved):

PR #776, opened 2024-06-07
Improvements to the Embedding functionality

dkotter:
### Description of the Change

In testing out another plugin that uses the Embedding functionality from ClassifAI, ran into a few issues that could be improved.

1. The OpenAI API allows you to [generate Embeddings](https://platform.openai.com/docs/api-reference/embeddings/create#embeddings-create-input) from either a single string or an array of strings. In #758 we made some major improvements to Embeddings, one of which was to break content down into smaller chunks and embed each chunk. This was done with individual API requests as I didn't realize you could send array data. This PR fixes that by checking the size of the content and if it's below the token limit, we send that as a single request. For any bulk processing, this reduces the number of API requests significantly, helping with performance and helping to avoid rate limit errors
2. Also discovered a bug in our chunking code where each chunk would be larger than the last. We want an overlap with each chunk but there was a bug here where each iteration we would increase that overlap. This wasn't noticeable for smaller posts but for long posts, you'd end up with huge chunks that were over the token limit, resulting in API failures. This also meant more tokens were being used that necessary. For any bulk processing, this helps reduce the tokens used significantly
3. Ensure we set a higher timeout for the Azure OpenAI Embedding requests, as these can timeout at the default 5 seconds

### How to test the Change

1. Setup the Classification Feature to use either OpenAI or Azure OpenAI (or try both)
2. Preview a post and ensure results show
3. (Optional) Test classifying a post and ensure results show

### Changelog Entry

> Added - Higher timeout for Azure OpenAI Embedding requests
> Changed - Add the ability to send chunked content in a single request to the Embedding API
> Fixed - Ensure we chunk content correctly, keeping each chunk roughly the same size

### Credits

Props @dkotter

### Checklist:

### DIFF
--- a/includes/Classifai/Providers/Azure/Embeddings.php
+++ b/includes/Classifai/Providers/Azure/Embeddings.php
@@ -6,6 +6,7 @@
 namespace Classifai\Providers\Azure;
 
 use Classifai\Providers\OpenAI\EmbeddingCalculations;
+use Classifai\Providers\OpenAI\Tokenizer;
 use Classifai\Normalizer;
 use Classifai\Features\Classification;
 use Classifai\Features\Feature;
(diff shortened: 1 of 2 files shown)


iamdharmesh (approved):
Thanks for this improvement @dkotter Code Looks good and it tests well.

PR #779, opened 2024-06-17
fix/759: Add background processing for Embeddings Classification

### LINKED ISSUE #759: Need a more performant way to bulk generate embeddings for terms
### Describe the bug

In v2.2.0 of ClassifAI we added the ability to classify content within your own terms using OpenAI Embeddings. In order for this to work, we need embedding data to be generated for each term and for the post we are comparing those terms to.

The post embedding data is generated on the fly when the comparison is triggered but we don't want to do that for terms as we may have hundreds or thousands. So these are generated in bulk when the feature is first set up. This has always been a known limitation, that if you have lots of terms, this process will probably run into either timeouts, memory issues or rate limit issues with OpenAI.

In #758, we are doing some changes to how OpenAI Embeddings work but we have not yet fixed this issue, so ideally that is fixed and added to the same release (as these changes require all embeddings to be regenerated).

There are two issues I'm currently aware of:

1. We generate these embeddings when the settings are saved. But we only generate embeddings for taxonomies that are turned on. So the first time you save, the taxonomy settings aren't saved yet so we don't run anything. You have to save again for things to work
2. We generate embeddings for each term that doesn't currently have an embedding saved during this process (which again, fires when the settings are saved). For sites with 1000+ terms, this will almost certainly lead to timeouts or memory issues. Sites with far fewer terms will probably run into OpenAI rate limits

Ideally we would introduce some sort of queue management system to address this, ideally making this a general enough solution that it can be used by other features that may come in the future. There are tools out there we could look to use, like [Action Scheduler](https://actionscheduler.org/) or [Cavalcade](https://github.com/humanmade/Cavalcade), but we may be fine just building a lightweight system on top of the scheduled event system in WordPress.

### Steps to Reproduce

1. Setup the Classification Feature with OpenAI Embeddings as the Provider
2. Turn on at least one taxonomy and hit save
3. Notice that no embeddings are actually generated
4. Hit save again and notice the embeddings get generated

Can also generate 1000+ terms and try running this process again, though note this will cost money since it makes API requests. I've tested locally using an embeddings model run through Ollama and at around 1000 terms, I run into memory issues

_No response_

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

Sidsector9:
### Description of the Change

This PR implements Action Scheduler to perform Embeddings Classification with 20 terms per job. (The count is opinionated, open to change it).

Closes #759 

### How to test the Change

1. Visit `/wp-admin/tools.php?page=classifai&tab=language_processing`
2. Click save.
3. Observe the notice "**OpenAI Embeddings**: Classification is in progress in the background."
4. Wait until message updates to "**OpenAI Embeddings**: Classification completed."
5. Verify the embeddings generated in the `wp_termmeta` table.

### Changelog Entry

> Added - `classifai_openai_embeddings_max_terms` filter hook to change the number of terms to process during Embeddings classification per job.
> Changed - Embeddings classification now uses background processing to improve performance.

### Credits

Props @dkotter @Sidsector9 

### Checklist:

### DIFF
--- a/classifai.php
+++ b/classifai.php
@@ -84,6 +84,7 @@ require_once __DIR__ . '/config.php';
 function classifai_autoload() {
 	if ( file_exists( CLASSIFAI_PLUGIN_DIR . '/vendor/autoload.php' ) ) {
 		require_once CLASSIFAI_PLUGIN_DIR . '/vendor/autoload.php';
+		require_once CLASSIFAI_PLUGIN_DIR . '/vendor/woocommerce/action-scheduler/action-scheduler.php';
 
 		return true;
 	} else {
--- a/classifai.php
+++ b/classifai.php
@@ -84,7 +84,6 @@ require_once __DIR__ . '/config.php';
 function classifai_autoload() {
 	if ( file_exists( CLASSIFAI_PLUGIN_DIR . '/vendor/autoload.php' ) ) {
 		require_once CLASSIFAI_PLUGIN_DIR . '/vendor/autoload.php';
-		require_once CLASSIFAI_PLUGIN_DIR . '/vendor/woocommerce/action-scheduler/action-scheduler.php';
 
 		return true;
 	} else {
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,8 @@
     "php": ">=7.4",
     "yahnis-elsts/plugin-update-checker": "5.1",
     "ua-parser/uap-php": "dev-master",
-    "aws/aws-sdk-php": "^3.300"
+    "aws/aws-sdk-php": "^3.300",
+    "woocommerce/action-scheduler": "dev-trunk"
   },
   "autoload": {
     "psr-4": {
--- a/composer.json
+++ b/composer.json
@@ -14,7 +14,7 @@
     "yahnis-elsts/plugin-update-checker": "5.1",
     "ua-parser/uap-php": "dev-master",
     "aws/aws-sdk-php": "^3.300",
-    "woocommerce/action-scheduler": "dev-trunk"
+    "woocommerce/action-scheduler": "3.8.1"
   },
   "autoload": {
     "psr-4": {
--- a/includes/Classifai/Providers/OpenAI/Embeddings.php
+++ b/includes/Classifai/Providers/OpenAI/Embeddings.php
@@ -808,6 +808,18 @@ class Embeddings extends Provider {
 			}
 		}
 
+		/**
+		 * Filter the maximum number of terms to process.
+		 *
+		 * @since x.x.x
+		 * @hook classifai_openai_embeddings_max_terms
+		 *
+		 * @param {int} $number Number of terms to process per job.
+		 *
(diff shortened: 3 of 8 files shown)


composer.json
```diff
@@ -13,7 +13,8 @@
     "php": ">=7.4",
     "yahnis-elsts/plugin-update-checker": "5.1",
     "ua-parser/uap-php": "dev-master",
-    "aws/aws-sdk-php": "^3.300"
+    "aws/aws-sdk-php": "^3.300",
+    "woocommerce/action-scheduler": "dev-trunk"
```
dkotter: We should lock this to a specific version, otherwise this will end up potentially pulling in new changes anytime we build a release, which could cause unexpected problems

classifai.php
```diff
@@ -84,6 +84,7 @@ function classifai_define( $name, $value ) {
 function classifai_autoload() {
 	if ( file_exists( CLASSIFAI_PLUGIN_DIR . '/vendor/autoload.php' ) ) {
 		require_once CLASSIFAI_PLUGIN_DIR . '/vendor/autoload.php';
+		require_once CLASSIFAI_PLUGIN_DIR . '/vendor/woocommerce/action-scheduler/action-scheduler.php';
```
dkotter: I wonder if we'd be better off only loading this if we need it. For instance, if a site isn't using a Feature that utilizes this library (in this case, just the Embeddings Feature is using it) seems better to not have the overhead of loading this

includes/Classifai/Providers/OpenAI/Embeddings.php
```diff
@@ -801,21 +809,68 @@ private function trigger_taxonomy_update( string $taxonomy = '', bool $all = fal
...
+		}
+
+		$job_args = [
+			'taxonomy' => $taxonomy,
+			'all'      => $all,
+			'args'     => $default_args,
+			'user_id'  => $user_id,
+		];
+
+		\as_enqueue_async_action( 'classifai_schedule_generate_embedding_job', $job_args );
```
dkotter: One thing to consider. If we have no terms to generate embeddings for, we still schedule this action and show a message to the user that generation is in progress. This means anytime the settings are saved, that message shows. Wondering if we want to run the actual term query here and then only proceed if we have results?

dkotter (changes requested):
Left a few comments but overall this looks really good and seems to work well.

One other thing we'll need to do before this is ready to merge is bring these same changes over to the Azure OpenAI Embeddings Provider, as right now we only added this to the OpenAI Embeddings Provider

dkotter (commented):
This looks good and tests well but still need to account for this piece:

> bring these same changes over to the Azure OpenAI Embeddings Provider

@Sidsector9 Let me know if you'd like me to tackle that

Sidsector9:
@dkotter sorry for the delay, I couldn't find time to pick this up. I'll work on Azure and revert soon for review.

dkotter (approved):

PR #786, opened 2024-08-01
Add Azure Language Services provider

### LINKED ISSUE #159: Auto-populate missing meta tags and descriptions

psorensen:
### Description of the Change

This PR adds Azure Language Services as a provider and adds it to the Excerpt Generation feature.

Closes #159 

### How to test the Change

Testing Steps:
1. Create new Azure resource for Language Services
2. Enter Credentials into the Excerpt Generation feature screen
3. Confirm valid credentials allow feature to be enabled
4. Test 'Generate Excerpt' tag for allowed post types.

### Changelog Entry

> Added - Azure Language Services provider with Excerpt Generation.

### Credits

Props @psorensen @cadic @jeffpaul @Sidsector9 @dkotter 

### Checklist:


psorensen:
@dkotter How should I handle the VIP flag for `wp_remote_get` usage? Should I conditionally use it if defined, or ignore this rule?

dkotter:
> @dkotter How should I handle the VIP flag for `wp_remote_get` usage? Should I conditionally use it if defined, or ignore this rule?

So we've not been consistent with this so far in the codebase. We've basically done both, where we use `vip_safe_wp_remote_get` if it exists (see [here](https://github.com/10up/classifai/blob/develop/includes/Classifai/Providers/Azure/Speech.php#L216)) and otherwise use `wp_remote_get` and in other places we just use `wp_remote_get` and add an ignore statement.

Personally I'd prefer we go with the first approach, as this is technically a VIP approved plugin so it's nice to match their standards. We have had issues though a few times with timeouts, as some of these services take a bit to return. `vip_safe_wp_remote_get` only allows you to set a timeout of 3 seconds so if we need higher than that here, we'll want to stick with `wp_remote_get` only

psorensen:
sounds good @dkotter - I've included a helper function to facilitate a refactor of all the existing usage of `wp_remote_get`

jeffpaul:
@psorensen is this still WIP or can this formally go through review/merge?

psorensen:
@jeffpaul Good to review - removed the WIP from the title

includes/Classifai/Providers/Azure/Language.php
```diff
@@ -0,0 +1,391 @@
...
+	 *
+	 * @return bool|WP_Error
+	 */
+	protected function authenticate_credentials( string $url, string $api_key ) {
+		$rtn = false;
+
+		$endpoint = trailingslashit( $url ) . '/text/analytics/v3.1/languages';
+		$endpoint = add_query_arg( 'api-version', static::API_VERSION, $endpoint );
+
+		$request = safe_wp_remote_get(
```
Sidsector9: This is throwing error. Should it use `wp_remote_post()` instead?

includes/Classifai/Providers/Azure/Language.php
```diff
@@ -0,0 +1,391 @@
...
+		$summary = '';
+
+		if ( ! is_wp_error( $request ) ) {
+			$response = json_decode( wp_remote_retrieve_body( $request ) );
+			if ( ! empty( $response->error ) ) {
+				return new WP_Error( 'auth', $response->error->message );
+			}
+
+			while ( 'succeeded' !== $response->status ) {
+				sleep( .5 );
```
Sidsector9: A downside of this approach is that if you navigate away from the page, the process will quit.
We can use Action Scheduler for background processing which we're using for the Classification feature. (#779)

@dkotter do you have any thoughts on this?

includes/Classifai/Providers/Azure/Language.php
```diff
@@ -0,0 +1,391 @@
...
+						 * Filter the summary length.
+						 * Possible values are 'oneSentence', 'short', 'medium', 'long'.
+						 * Default is 'oneSentence'.
+						 *
+						 * @since 3.2.0
+						 * @hook classifai_azure_language_summary_length
+						 * @param string $summary_length The summary length.
+						 * @return string
+						 */
+						'summaryLength' => apply_filters( 'classifai_azure_language_summary_length', 'oneSentence' ),
```
Sidsector9, line 239: We'll need to format the doc block

```php
/**
 * Filter the summary length.
 * Possible values are 'oneSentence', 'short', 'medium', 'long'.
 * Default is 'oneSentence'.
 *
 * @since 3.2.0
 * @hook classifai_azure_language_summary_length
 * @param {string} $summary_length The summary length.
 * @return {string} The summary length
 */
```

includes/Classifai/Providers/Azure/Language.php
```diff
@@ -0,0 +1,391 @@
...
+						'headers' => [
+							'Ocp-Apim-Subscription-Key' => $api_key,
+							'Content-Type'              => 'application/json',
+						],
+					]
+				);
+				$response = json_decode( wp_remote_retrieve_body( $request ) );
+			}
+
+			$summary = $response->tasks->items[0]->results->documents[0]->summaries[0]->text;
```
Sidsector9: Just want to confirm that the array indexes at `[0]` will always contain those properties, or if we should check via `isset()` and `count() > 0`?

Sidsector9 (dismissed):

includes/Classifai/Providers/Azure/Language.php
dkotter: I think the problem here just in quickly looking at the code is that this method gets called from a REST request. The easy fix is what we're doing here, just pausing and retrying until we get a successful response.

I do wonder what happens if the response never succeeds? Would this code just be stuck in this loop?

Action Scheduler would be a nicer fix here but would be more complicated with how everything else is set up in ClassifAI. Basically when the `Generate Excerpt` button is clicked, the initial request would trigger Action Scheduler and then on the javascript side, we would need to continue polling until the request is done (and show the loading indicator the entire time).

That's doable but doesn't match how we handle things elsewhere so would need to think through how best to achieve that, especially in a way that would work for other Features and Providers in the future

includes/Classifai/Providers/Azure/Language.php
psorensen: ~~See this [comment](https://github.com/10up/classifai/pull/786#issuecomment-2269844794) and the follow-up comments for context. What error are you seeing?~~

Sidsector9: I am seeing the following when trying to connect to the service:

Screenshot 2024-09-06 at 2 10 22 PM

Sidsector9: Here's the neater version:

```
[06-Sep-2024 08:40:15 UTC] PHP Fatal error:  Uncaught TypeError: http_build_query(): Argument #1 ($data) must be of type array, string given in /Users/work/Sites/classifai/app/public/wp-includes/Requests/src/Transport/Curl.php:578
Stack trace:
#0 /Users/work/Sites/classifai/app/public/wp-includes/Requests/src/Transport/Curl.php(578): http_build_query('{"documents": [...', '', '&')
#1 /Users/work/Sites/classifai/app/public/wp-includes/Requests/src/Transport/Curl.php(391): WpOrg\Requests\Transport\Curl::format_get('https://classif...', '{"documents": [...')
#2 /Users/work/Sites/classifai/app/public/wp-includes/Requests/src/Transport/Curl.php(171): WpOrg\Requests\Transport\Curl->setup_handle('https://classif...', Array, '{"documents": [...', Array)
#3 /Users/work/Sites/classifai/app/public/wp-includes/Requests/src/Requests.php(469): WpOrg\Requests\Transport\Curl->request('https://classif...', Array, '{"documents": [...', Array)
#4 /Users/work/Sites/classifai/app/public/wp-includes/class-wp-http.php(397): WpOrg\Requests\Requests::request('https://classif...', Array, '{"documents": [...', 'GET', Array)
#5 /Users/work/Sites/classifai/app/public/wp-includes/class-wp-http.php(638): WP_Http->request('https://classif...', Array)
#6 /Users/work/Sites/classifai/app/public/wp-includes/http.php(184): WP_Http->get('https://classif...', Array)
#7 /Users/work/Sites/classifai/app/public/wp-content/plugins/classifai/includes/Classifai/Helpers.php(686): wp_remote_get('https://classif...', Array)
#8 /Users/work/Sites/classifai/app/public/wp-content/plugins/classifai/includes/Classifai/Providers/Azure/Language.php(164): Classifai\safe_wp_remote_get('https://classif...', Array)
#9 /Users/work/Sites/classifai/app/public/wp-content/plugins/classifai/includes/Classifai/Providers/Azure/Language.php(122): Classifai\Providers\Azure\Language->authenticate_credentials('https://classif...', '95193e52eec647f...')
#10 /Users/work/Sites/classifai/app/public/wp-content/plugins/classifai/includes/Classifai/Features/Feature.php(258): Classifai\Providers\Azure\Language->sanitize_settings(Array)
#11 /Users/work/Sites/classifai/app/public/wp-includes/class-wp-hook.php(326): Classifai\Features\Feature->sanitize_settings(Array)
#12 /Users/work/Sites/classifai/app/public/wp-includes/plugin.php(205): WP_Hook->apply_filters(Array, Array)
#13 /Users/work/Sites/classifai/app/public/wp-includes/formatting.php(5115): apply_filters('sanitize_option...', Array, 'classifai_featu...', Array)
#14 /Users/work/Sites/classifai/app/public/wp-includes/option.php(866): sanitize_option('classifai_featu...', Array)
#15 /Users/work/Sites/classifai/app/public/wp-admin/options.php(344): update_option('classifai_featu...', Array)
#16 {main}
  thrown in /Users/work/Sites/classifai/app/public/wp-includes/Requests/src/Transport/Curl.php on line 578

```

psorensen: > Should it use wp_remote_post() instead?

Ignore my previous comment. Yes it should. 😄  Updated

includes/Classifai/Providers/Azure/Language.php
psorensen: @dkotter @Sidsector9  Knowing the action scheduler would be a heavier lift, how about a simple counter for the short term? See
* https://github.com/10up/classifai/pull/786/commits/a34e4f20cb5a313014860793c66fd94095ab4605
* https://github.com/10up/classifai/pull/786/commits/656b5454d562d1d33d68072fd74af66fbe2c6b66

jeffpaul:
@dkotter @Sidsector9 back to you for review

jeffpaul:
Let's hold this for the 4.0.0 release where we may have things timed that if the WP AI Client gets into WordPress 7.0 that we could publish this as a composer package, add test coverage, and ensure this covers all respective features and not just excerpt generation.

dkotter:
Are plan now is to move over to the new AI Client and Connectors API that is in WordPress 7.0. As such, the idea of Providers here in ClassifAI will change and will likely be removed in favor of individual AI Connectors. Going to close this out for now though this could be extracted out into it's own AI Provider plugin if desired

PR #771, opened 2024-05-17
Onboarding and settings refactor to React

iamdharmesh:
### Description of the Change
The PR refactors the settings and onboarding to a React-based implementation and deprecates the existing settings panel. The existing settings panel is deprecated but can still be enabled using the `classifai_use_legacy_settings_panel` filter.

screencapture-classifai-local-wp-admin-tools-php-2024-09-19-16_17_44

Closes #502
Closes #449 

### How to test the Change
**Settings:**
1. Go to `Tools` > `ClassifAI`.
1. Ensure the settings page loads correctly and that navigation works as expected.
1. Update feature settings, such as enabling/disabling features or changing/configuring providers, and verify that everything functions as expected.
1. **Existing settings:** Configure the ClassifAI plugin using the `develop` branch, then switch to this PR branch. Confirm that no settings are lost and that all data is correctly reflected in the new settings page.
1. **Use Legacy Settings:** Add the following code to the theme’s functions.php file or plugin: `add_filter( 'classifai_use_legacy_settings_panel', '__return_true' );`. Verify that the old settings page loads and works as expected.

**OnBoarding**
1. Deactivate the ClassifAI plugin and activate it again.
2. You will see a notice to start the ClassifAI setup. Click the "Start Setup" button.
3. Follow the setup process and ensure everything works as expected.
4. **Existing settings:** Visit the setup page from the settings page and verify that it correctly reflects the existing configured features, with the option to modify the settings accordingly.
5. **Use Legacy Settings:** Add the following code to the theme’s functions.php file or plugin: `add_filter( 'classifai_use_legacy_settings_panel', '__return_true' );`. Verify that the old onboarding/setup page loads and works as expected.

**Add additional feature fields**
1. Add a new feature to the plugin by following the guide [here](https://github.com/10up/classifai/blob/enhancement/react-settings/hookdocs/useful-snippets.md#add-a-new-feature).
2. Follow step 4 to add additional feature fields to the settings and validate that everything works as expected.

**Add custom Provider fields**
1. Add a custom provider to any feature by following the guide [here](https://github.com/10up/classifai/blob/enhancement/react-settings/hookdocs/useful-snippets.md#add-custom-provider-to-an-existing-feature).
2. Add the required provider fields to the settings by following step 5 in the guide and validate that everything works as expected.

**General**
Run a full regression test to ensure everything is working as expected.

### Credits

Props @jeffpaul @dkotter @iamdharmesh @Sidsector9 

### Checklist:

[...]

PR #790, opened 2024-08-02
[WIP] Add AWS Amazon Personalize as a Provider for the Recommended Content Feature

dkotter:
### Description of the Change

Currently the only Provider we have for the Recommended Content Feature is Azure Personalizer. That Provider has been deprecated by Azure and is no longer available to use. In doing lots of research, a fairly close drop-in replacement is AWS Amazon Personalize. This PR adds a new Provider for that.

Worth noting there are lots of use cases that Amazon Personalize provides, some more out of the box and some that require a bit more setup. This PR is focusing on the [Personalized Ranking recipe](https://docs.aws.amazon.com/personalize/latest/dg/native-recipe-personalized-ranking-v2.html) as that very closely matches what we were doing with Azure Personalizer. We could look to support other use cases in the future though.

Closes #

### How to test the Change

More to come

### Changelog Entry

> Added - AWS Amazon Personalize as a new Provider for the Recommended Content Feature

### Credits

Props @dkotter, @Sidsector9 

### Checklist:

### DIFF
--- a/includes/Classifai/Providers/AWS/AmazonPersonalize.php
+++ b/includes/Classifai/Providers/AWS/AmazonPersonalize.php
@@ -468,7 +468,8 @@ class AmazonPersonalize extends Provider {
 	public function get_default_recommended_content( array $attributes ) {
 		$post_type      = $attributes['contentPostType'];
 		$key_attributes = [
-			'terms' => $attributes['taxQuery'] ?? [],
+			'terms'    => $attributes['taxQuery'] ?? [],
+			'excluded' => $attributes['excludeId'] ?? 0,
 		];
 		$transient_key  = 'classifai_actions_' . $post_type . md5( maybe_serialize( $key_attributes ) );
 		$content        = get_transient( $transient_key );
--- a/includes/Classifai/Providers/AWS/AmazonPersonalize.php
+++ b/includes/Classifai/Providers/AWS/AmazonPersonalize.php
@@ -115,8 +115,26 @@ class AmazonPersonalize extends Provider {
 				'label_for'     => 'event_tracker_id',
 				'input_type'    => 'text',
 				'default_value' => $settings['event_tracker_id'], // TODO: could make an API request to get all event trackers and populate a dropdown. Or make an API request to get the dataset and automatically get the tracker ID from the chosen dataset.
+				'placeholder'   => '4282ac5f-1681-53fg-8g35-1acf2gae4125',
 				'class'         => 'large-text classifai-provider-field hidden provider-scope-' . static::ID,
 				'description'   => esc_html__( 'Enter the event tracker ID associated with your dataset', 'classifai' ),
+			]
+		);
+
+		add_settings_field(
+			'campaign_arn',
+			esc_html__( 'Campaign ARN', 'classifai' ),
+			[ $this->feature_instance, 'render_input' ],
+			$this->feature_instance->get_option_name(),
+			$this->feature_instance->get_option_name() . '_section',
+			[
+				'option_index'  => static::ID,
+				'label_for'     => 'campaign_arn',
+				'input_type'    => 'text',
+				'default_value' => $settings['campaign_arn'],
+				'placeholder'   => 'arn:aws:personalize:us-east-1:12345:campaign/name',
+				'class'         => 'large-text classifai-provider-field hidden provider-scope-' . static::ID,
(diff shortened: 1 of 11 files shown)


dkotter:
@jeffpaul The dependency review workflow is failing on this PR: https://github.com/10up/classifai/actions/runs/10221897513/job/28285404935?pr=790#step:3:11.

Looking at the licenses it's flagging (Apache-2.0 and MIT) seems both of those should be allowed based off our config: https://github.com/10up/.github/blob/trunk/.github/dependency-review-config.yml

I'm not familiar enough with this action to know if this is an issue with that, an issue with our config or something else that needs fixed up

includes/Classifai/Providers/AWS/AmazonPersonalize.php
```diff
@@ -0,0 +1,740 @@
...
+	 */
+	public function track_event( $post_id, array $args = [] ) {
+		$settings = $this->feature_instance->get_settings( static::ID );
+		$client   = $this->get_client( 'personalize-events' );
+
+		if ( ! $client ) {
+			return new WP_Error( 'client_not_found', esc_html__( 'Client not found.', 'classifai' ) );
+		}
+
+		$id     = uniqid(); // TODO: Is this the best way to generate a session ID?
```
Sidsector9, line 672: This will create a new ID for every event. We'll need to generate a session ID that remains the same through the session. We can do something like the following?

```php
add_action( 'init', function() {
    if ( ! session_id() ) {
        session_start();
    }
} );
```

Few points from the docs:

1. Your application generates a unique `sessionId` when a user first visits your website or uses your application. You must use the same `sessionId` in all events throughout the session. Amazon Personalize uses the `sessionId` to associate events with the user before they log in (is anonymous).
2. To provide real-time personalization for anonymous users, specify the `sessionId` as the `userId` in your `GetRecommendations` or `GetActionRecommendations` request.

includes/Classifai/Providers/AWS/AmazonPersonalize.php
```diff
@@ -0,0 +1,740 @@
...
+			'eventList'  => [
+				[
+					'eventType' => $args['event']['type'] ?? 'click',
+					'itemId'    => (string) $post_id,
+					'sentAt'    => time(),
+				],
+			],
+			'sessionId'  => $id,
+			'trackingId' => $settings['event_tracker_id'] ?? '',
+			'userId'     => $id, // TODO: We need a user ID that can be tracked across page views.
```
Sidsector9, line 683: For logged-in users, can this be the value returned by `get_current_user_id()`, something like:

```php
'userId' => get_current_user_id() ?? session_id(),
```

iamdharmesh:
>@jeffpaul The dependency review workflow is failing on this PR: https://github.com/10up/classifai/actions/runs/10221897513/job/28285404935?pr=790#step:3:11.
>
 >Looking at the licenses it's flagging (Apache-2.0 and MIT) seems both of those should be allowed based off our config: https://github.com/10up/.github/blob/trunk/.github/dependency-review-config.yml
> 
> I'm not familiar enough with this action to know if this is an issue with that, an issue with our config or something else that needs fixed up

@jeffpaul, I'm seeing a similar failure in PR https://github.com/10up/classifai/pull/771 as well. I'm not sure of the exact reason for this. Could it be because `Apache-2.0` is compatible with GNU GPL-v3 but not with GPL-v2, as mentioned [here](https://www.gnu.org/licenses/license-list.html#apache2)?

iamdharmesh:
**Update:**

Regarding the dependency review workflow failing, it was due to an [issue](https://github.com/actions/dependency-review-action/issues/809) in the latest version (thanks to @jeffpaul for identifying this). Currently, I have downgraded dependency-review-action to v4.3.3, and it's working as expected. Once the issue is resolved upstream, I will update the workflow file to use the latest version.
cc: @dkotter

jeffpaul:
I believe the prior testing we did here showed our testing to be rather expensive (~$100-200), correct?  If so, probably best to close out this PR as unlikely to be something we think would be a good fit for the broad spectrum of ClassifAI users?

dkotter:
> I believe the prior testing we did here showed our testing to be rather expensive (~$100-200), correct? If so, probably best to close out this PR as unlikely to be something we think would be a good fit for the broad spectrum of ClassifAI users?

It was more expensive than we had expected (especially for just running locally) though I do think some of that could be addressed with some code updates and more aggressive caching. But it was also super difficult to setup on AWS, to the point that I don't think anyone would actually go through all those steps. If we do want this integration in the future, I'd suggest we look at automatting (via their API) as much of the setup process as possible.

PR #801, opened 2024-09-12
Add a Smart 404 Feature

### LINKED ISSUE #794: Add ability to craft "Smart 404" pages

Let's investigate pulling in a Smart 404 feature here based on explorations done elsewhere.  Primary concern for me is how that integrates to our setup/settings flows and otherwise presented within the admin/frontend.  Similarly will want to explore how we require/include Elasticsearch/ElasticPress here alongside ClassifAI.  Will also want to ensure we update classifaplugin.com with this feature.

### Designs

_No response_

_No response_

### Code of Conduct

dkotter:
### Description of the Change

Adding in a new Feature, Smart 404, that uses the power of embeddings, combined with vector search capabilities in Elasticsearch, to recommend content when users land on a 404 page, with the hope of recommending the actual item they were hoping to reach.

As mentioned, this Feature requires two services in order to function:

1. A Provider that can generate embeddings (right now we support OpenAI and Azure OpenAI)
2. A data storage engine that can also run vector queries (currently only support Elasticsearch powered by ElasticPress)

Once both are in place, all of your existing content (and any new content added in the future) is automatically sent off to the embeddings Provider. This vector information is then stored in elasticsearch and we can then run vector queries on this data. In particular, we've introduced new helper methods that can be dropped into a site's 404 template that will render these results.

At a high level, here's how the process works:

1. When an item is saved, the content is broken down into smaller chunks. Each chunk is then sent off to an embeddings Provider (like OpenAI)
2. The returned vector data is synced over to elasticsearch using ElasticPress
3. When a user lands on a 404 page, we pull the last slug out of the URL. For instance, if someone is trying to go to `example.com/news/interesting-news-article`, we take the slug `interesting-news-article`
4. This slug is then sent off to the embeddings Provider and the returned vector data is used to make a vector query against elasticsearch
5. There are multiple configuration options here that can be used to modify this query but we eventually end up with results back which are then rendered to the user

Closes #794

### How to test the Change

1. Follow the instructions in the [README](https://github.com/10up/classifai/pull/801/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R445) to get things set up. This will include setting up an embeddings Provider and ElasticPress + Elasticsearch
2. Run an ElasticPress sync from scratch. This can be done in the admin or using the WP-CLI command that ElasticPress provides. Important to run this from scratch so the proper mapping is sent
3. Utilize one of the new helper functions to output results on your 404 template. Ensure this renders properly and that all various settings work as expected

Note it's helpful to install [Query Monitor](https://wordpress.org/plugins/query-monitor/) and the [ElasticPress Debugging Add-On](https://wordpress.org/plugins/debug-bar-elasticpress/) to see the actual elasticsearch queries and results

### Changelog Entry

> Added - New Feature, Smart 404, which provides the ability to render recommended content on a 404 page

### Credits

Props @dkotter, @iamdharmesh, @berkod 

### Checklist:

### DIFF
--- a/includes/Classifai/Features/Smart404EPIntegration.php
+++ b/includes/Classifai/Features/Smart404EPIntegration.php
@@ -617,4 +617,62 @@ class Smart404EPIntegration {
 
 		return $source;
 	}
+
+	/**
+	 * Convert Elasticsearch results to WP_Post objects.
+	 *
+	 * @param array $results Document results from Elasticsearch.
+	 * @return array
+	 */
+	public function convert_es_results_to_post_objects( array $results ): array {
+		$new_posts = [];
+
+		// Turn each ES result into a WP_Post object.
+		// Copied from ElasticPress\Indexable\Post\QueryIntegration::format_hits_as_posts.
+		foreach ( $results as $post_array ) {
+			// Don't convert if not needed.
+			if ( is_a( $post_array, 'WP_Post' ) ) {
+				$new_posts[] = $post_array;
+				continue;
+			}
+
+			$post = new \stdClass();
+
+			$post->ID      = $post_array['post_id'];
+			$post->site_id = get_current_blog_id();
+
+			if ( ! empty( $post_array['site_id'] ) ) {
+				$post->site_id = $post_array['site_id'];
+			}
+
+			$post_return_args = [
+				'post_type',
+				'post_author',
+				'post_name',
+				'post_status',
+				'post_title',
+				'post_content',
+				'post_excerpt',
+				'post_date',
+				'post_date_gmt',
+				'permalink',
+			];
+
+			foreach ( $post_return_args as $key ) {
+				if ( 'post_author' === $key ) {
+					$post->$key = $post_array[ $key ]['id'];
+				} elseif ( isset( $post_array[ $key ] ) ) {
+					$post->$key = $post_array[ $key ];
+				}
+			}
+
+			$post->elasticsearch = true;
+
+			if ( $post ) {
+				$new_posts[] = $post;
+			}
+		}
+
+		return $new_posts;
+	}
 }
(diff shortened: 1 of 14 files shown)


dkotter:
Note there are a couple other things we could consider doing here, either in this PR or in follow up PRs:

1. Right now we check if ElasticPress is active before we allow the Feature to be configured (and we show an error message if not). But we never check to see if ElasticPress is actually configured properly and connected to elasticsearch. There may be a helper function we can utilize in ElasticPress for this
2. We have a couple helper functions that can be used to render the results on a 404 page but this does require someone to manually add those to their template. May be worth adding a Smart 404 block that can be used in FSE themes to add this functionality directly to their 404 template

includes/Classifai/Features/Smart404EPIntegration.php
```diff
@@ -0,0 +1,681 @@
...
+			}
+		}
+	}
+
+	/**
+	 * Inintialize the class and register the needed hooks.
+	 */
+	public function init() {
+		// Vector support was added in Elasticsearch 7.0.
+		if ( version_compare( $this->es_version, '7.0', '<=' ) ) {
```
iamdharmesh:
```suggestion
		if ( ! $this->es_version || version_compare( $this->es_version, '7.0', '<=' ) ) {
```

As this gives PHP deprecated notice if Elasticsearch is not configured.
`PHP Deprecated:  version_compare(): Passing null to parameter #1 ($version1) of type string is deprecated in /Users/****/wp-content/plugins/classifai/includes/Classifai/Features/Smart404EPIntegration.php on line 82`

iamdharmesh (dismissed):
Thanks for the Great work here @dkotter. The code looks great and it tests well. Just added one minor suggestion otherwise this is ready for merge.

> Note there are a couple other things we could consider doing here, either in this PR or in follow up PRs:


I think better to handle it in separate follow up PRs.

PR #803, opened 2024-09-16
[WIP] feat/489: Add Rewrite Tone feature

### LINKED ISSUE #489: Rewrite content to match brand standards

Consider publishers who have established style guides and other documents and policies that describe how content published on their site should be presented/written.  Alongside that is also the back catalog of published posts.  Those guiding docs/policies and existing content could be used to train a custom model on the brand standard (aka "tone") for a site.  That model could then be used to rewrite portions of a post or an entire post to match those standards/tone.  Similar to how #488 would add expand/reduce options for post content, this use case would be an option to `Rewrite tone` or similar descriptor.

### Designs

We'll want to consider what the option is actually called/labeled, what any associated icon with it should be, as well as where it should appear in the block controls. But for now a "Rewrite tone" option as a primary toolbar option would be a perfect start as a prototype PR and allow us time to fine tune copy/icon/location.

One side effect to be aware of here is if someone selects multiple blocks and some are NOT paragraph blocks that that content (e.g., images) will likely not be returned by ChatGPT and thus be removed when the resulting ChatGPT response overwrites that content.  Perhaps by default this feature could ignore all content except text content (e.g. paragraph, heading, list blocks) and rewrite that content only?

### Code of Conduct

Sidsector9:
### Description of the Change

Closes #489 

### Screencast

The Tone Rewriting options are preserved in local storage so that the admin does not have to perform the same selection every time.

https://github.com/user-attachments/assets/c8624576-caba-4090-ac22-eb3400d772bf

### Changelog Entry

> Added - Rewrite Tone feature.

### Credits

Props @jeffpaul @Sidsector9 

### Checklist:


Sidsector9:
@jeffpaul I've updated the PR with a screencast showing how the UI will be for the admin. Can you please share your thoughts on it?

dkotter:
> @jeffpaul I've updated the PR with a screencast showing how the UI will be for the admin. Can you please share your thoughts on it?

I know I wasn't pinged but going to provide some thoughts anyway :)

Overall I think this looks good so no real complaints with this approach. What I do think is worth discussing is how best to ensure we have consistency across all of our different features.

As an example, right now someone could have a `Generate titles` button in the status panel and a `Generate excerpt` button in the excerpt panel and a custom ClassifAI panel with some options in it. Throw in this new experience and it starts to get a little confusing on where to go to manage things.

I don't think we need a single experience for everything as some things do make sense where they're at (for instance, I think the `Generate excerpt` button makes sense to stay). And I don't really have an answer here but bringing this up for discussion. Would it make sense to bring most of our options within this new plugin panel? Or are we okay with having things spread across multiple areas?

Also worth considering in progress features like #859 that adds a new button that opens a new modal. This feels like something that would be worth consolidating into what's done here, if we decide this approach is the right one

Sidsector9:
Thanks for reviewing @dkotter 

The main reason why this popup was introduced was to support multiple block selection of different types.
Gutenberg does not allow to add a Block Toolbar button in this case. In fact, as discussed with @jeffpaul, we're going to move the Content Resizing options from the Block Toolbar to this new popup as well, as we're looking forward to Resize Content across multiple blocks.

Once we move the Content Resizing options, we'll only have 2 different Plugin Panel areas – The sidebar and this new Popup in the editor header (unless we plan to re-introduce a custom Block Toolbar in the future).

> Would it make sense to bring most of our options within this new plugin panel?

I've added a collapsible panel within the Popup for this reason, however that adds an extra step. If we move all settings here, then it'll require thorough testing, as Gutenberg doesn't really allow to extend this area. I've added the button in the header through some heavy workarounds and a custom Slot-Fill.

@10up/open-source-practice pinging ya'll to weigh in your opinions.

dkotter:
> I've added the button in the header through some heavy workarounds and a custom Slot-Fill.

So I know Gutenberg has the `PluginSidebar` SlotFill, which similarly adds a button to that top header. I know the experience there is slightly different (opens a custom sidebar rather than the popup you have now) but curious if that was explored as an option?

Sidsector9:
> curious if that was explored as an option?

Yes. The `PluginSidebar` adds a custom Sidebar and a button for it to open/close the sidebar. We can't have a button there without the sidebar.

dkotter:
> > curious if that was explored as an option?
> 
> Yes. The `PluginSidebar` adds a custom Sidebar and a button for it to open/close the sidebar. We can't have a button there without the sidebar.

To be more clear, my confusion was around this piece:

> If we move all settings here, then it'll require thorough testing, as Gutenberg doesn't really allow to extend this area

I guess I would assume if we're using the `PluginSidebar` SlotFill, we'd be able to add whatever settings we would want into that sidebar without much issue (and I say all this without having looked at the code here so looking through that may answer my questions)

Sidsector9:
> I guess I would assume if we're using the PluginSidebar SlotFill, we'd be able to add whatever settings we would want into that sidebar without much issue (and I say all this without having looked at the code here so looking through that may answer my questions)

This is possible. If we do this, would we also move controls for Excerpt and Title Generation too?

dkotter:
> This is possible. If we do this, would we also move controls for Excerpt and Title Generation too?

That's really the main question here that I'm bringing up for conversation as I'm not sure what's best. My concern is if someone has this Rewrite Tone Feature on plus Title Generation plus Excerpt Generation plus Classification each of those adds their own way to trigger functionality and wondering if we should be concerned with that fractured approach.

My own opinion (not saying we should make changes yet, just to be clear) would be to get rid of the custom panel we currently have and all the settings in there and add those to this new custom plugin section. I'd probably leave the Excerpt Generation button alone and maybe Title Generation as well but could also see benefit to putting those in the same place.

Ultimately I don't think any of that needs to happen on this PR, just want to ensure we're thinking through the best UI/UX for new Features like this and ensuring we're not backing ourselves into a corner with any decisions we make.

jeffpaul:
I talked through this a bit with @Sidsector9 earlier today and my knee-jerk reaction is:
1. Any interactive element that is contextually alongside where an author/editor would otherwise be should stay as-is (e.g. Generate Excerpt, Generate Image).
2. All other items should probably get moved into a new ClassifAI-specific `PluginSidebar` with the ClassifAI `AI` icon as the button for the top of the editor as there are a multitude of cases where someone may not immediately "see" the existing ClassifAI sidebar panel (e.g. the have the sidebar collapsed, they're in a cover block's style or settings) and having the ClassifAI `AI` button always present that would get them into the ClassifAI `PluginSidebar` regardless where they are in the editor.  The one downside here is that could potentially mean at least one extra click (perhaps two) depending on where an author/editor happens to be in the editor and wants to interact with something from ClassifAI's `PluginSidebar`.

@fabiankaegy I'm curious if you might recommend something different/similar here?

Sidsector9:
@jeffpaul , @dkotter and I discussed this PR along with #859 in today's 1:1.

We broadly classified features into 2 parts:
1. Generation (Title, Excerpt, Key Takeaways, Content Generation)
2. Modification (Content Resizing and Rewrite Tone).

@fabiankaegy's implementation is well suited for the first one where content will be generated from scratch, where, once the data is generated, the user can then format the content.

In case of the second one, where the formatted content already exists:
- Modification should preserve the current HTML and formatting.
- Should work with multiple block selection where blocks may be nested.
- User should be able to preview the modified content with the current structure, which is not possible with the new Chatbox.

We concluded that once #859 is merged, I'll move the Rewrite Tone settings within the Chatbox by adding a tab, and the Rewrite result will be shown on a full-screen modal. This functionality will also be ported to the Content Resizing feature as well.

jeffpaul:
> Generation (Title, Excerpt, Key Takeaways, Content Generation)

Note that many of these can also be Modification scenarios wherein there's pre-existing text in any of those places.

Sidsector9:
### Update

The Rewrite Tone feature is integrated inside the Chat UI:

https://github.com/user-attachments/assets/8a6e93d8-9cdc-48fb-a612-8e18cb80d979

fabiankaegy:
@Sidsector9 still love this :) 

Just wanted to quickly share my thoughts:

https://share.cleanshot.com/V6GCYlJ8

Sidsector9:
@fabiankaegy thanks for the input.
I'm having a difficult time integrating Rewrite Tone as per your suggestion, while also keeping the ChatUI extendible for any future features that might want to integrate within the ChatUI component.

Currently, a feature can integrate into the ChatUI using the following filter:

```JS
const featureTabs = applyFilters( 'classifai.chatUI', [] )
```

- The ChatUI then uses the `TabPanel` component to loop over `featureTabs` and render `Slot` with a unique name.
- Each feature can then use the unique `name` and render the content in the `Slot` using a `Fill`.

Your suggestion can be implemented by introducing a `Slot` inside the Content Generation feature, and the Rewrite Tone feature can Fill within that Slot. My question is, should all future-features leveraging the ChatUI populate the same way?

But this creates another issue, if Content Generation Feature is disabled, but the Rewrite Tone feature is enabled, then the Slot inside Content Generation won't work.

jeffpaul:
So I _think_ there are two separate concerns we're talking about here.  (1) visually how the UX/UI works/looks in the Chat UI interface.  (2) how the ChatUI component exists within the codebase and can be extended.

On (1), I believe this is more generally where @fabiankaegy is commenting and providing guidance.  I think sorting out the optimal UX/UI handling within the Chat UI interface is something we should precise about as this PR gets merged in (as we're now potentially using that interface for more than one feature and now is the right time to sort things out better).

On (2), I believe this is more generally where @Sidsector9 is commenting and expressing concern.  I agree that the Chat UI experience should be part of a top-level ClassifAI plugin component that features could inherit into.  And if no features are enabled that utilize the ChatUI, then it won't appear in the editor.  But having the ChatUI exist within the Content Generation feature that other features like Rewrite Tone need to inject into feels sub-optimal.
>But this creates another issue, if Content Generation Feature is disabled, but the Rewrite Tone feature is enabled, then the Slot inside Content Generation won't work.

Sid's comment here expresses things fairly well on this (2) topic.

cc: @dkotter for review/input.

dkotter:
> I agree that the Chat UI experience should be part of a top-level ClassifAI plugin component that features could inherit into. And if no features are enabled that utilize the ChatUI, then it won't appear in the editor. But having the ChatUI exist within the Content Generation feature that other features like Rewrite Tone need to inject into feels sub-optimal.

Yes, this is correct. We'll need to refactor the code that renders the experience right now to make it not dependent on the Content Generation Feature. My thought would be to go with the slotfill approach where each Feature can register what they need within the experience (and properly enqueue that) and then each Feature will also be responsible for enqueuing the default ChatUI code (with WordPress being smart enough to not load that multiple times if multiple Features are in use)

Sidsector9:
> My thought would be to go with the slotfill approach where each Feature can register what they need within the experience (and properly enqueue that) and then each Feature will also be responsible for enqueuing the default ChatUI code (with WordPress being smart enough to not load that multiple times if multiple Features are in use)

@dkotter This is exactly how it is currently implemented. If you disable Content Generation but keep Rewrite Tone enabled, the ChatUI would still show, but the downside to this approach is that it makes it difficult to implement the UI that Fabian is suggesting.

jeffpaul:
Noting optimal approach for 3.8.0 is to add this functionality in via button iteration on block toolbar / top toolbar / sidebar panel.  In future release we can consider a chat-based experience, but for now looking to have this function in-context within the authoring/editing experience is the preference.

dkotter:
@Sidsector9 Curious if you remember where progress was left here? And what may be left to get this done?

PR #806, opened 2024-09-26
Add the WordPress Plugin Check Action

dkotter:
### Description of the Change

WordPress has an official Plugin Check plugin that runs a handful of different checks. They've brought the same functionality into a [GitHub Action](https://github.com/wordpress/plugin-check-action/) and this PR adds that as a workflow.

Also made some minor updates to our Build Release Zip Action (which we use here to run tests on the final build) and fix the few issues the Plugin Check found.

Note there are quite a [few checks](https://github.com/wordpress/plugin-check-action/?tab=readme-ov-file#supported-checks) the Plugin Check does, some of which we already do separately (like PHPCS). For now I've left all checks in place other than the Plugin Readme and Plugin Updater checks (both fail here) but we may want to trim the list down if we think some of these are overkill, though they run fast enough I think it's fine. 

### How to test the Change

Verify the [Plugin Check Action](https://github.com/10up/classifai/actions/runs/11060040009?pr=806) is passing on this PR

### Changelog Entry

> Fixed - Ensure all strings have translator comments
> Developer - Add the WordPress Plugin Check GitHub Action

### Credits

Props @dkotter

### Checklist:

### DIFF
--- a/.github/workflows/plugin-check.yml
+++ b/.github/workflows/plugin-check.yml
@@ -1,0 +1,42 @@
+name: WordPress Plugin Checks
+
+on:
+  push:
+    branches:
+      - develop
+      - trunk
+  pull_request:
+    branches:
+      - develop
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Set PHP version
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: '7.4'
+          coverage: none
+          tools: composer:v2
+
+      - name: Setup node
+        uses: actions/setup-node@v4
+        with:
+          node-version-file: .nvmrc
+          cache: npm
+
+      - name: Install dependencies and build
+        run: |
+          npm ci --no-optional
+          npm run build
+          composer install
+
+      - name: Run plugin check
+        uses: wordpress/plugin-check-action@v1
+        with:
+          exclude-directories: 'node_modules,vendor'
--- a/.github/workflows/plugin-check.yml
+++ b/.github/workflows/plugin-check.yml
@@ -11,7 +11,7 @@ on:
 
 jobs:
   build:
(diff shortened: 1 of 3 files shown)


.github/workflows/plugin-check.yml
```diff
@@ -0,0 +1,39 @@
...
+    branches:
+      - develop
+      - trunk
+  pull_request:
+    branches:
+      - develop
+
+jobs:
+  build:
+    uses: 10up/classifai/.github/workflows/build-release-zip.yml@feature/plugin-check-action # TODO: Update branch before merge.
```
dkotter: Note this should be updated to `develop` before merging in (though I always forget and end up having to fix after)

.github/workflows/plugin-check.yml
```diff
@@ -0,0 +1,39 @@
...
+        with:
+          name: ${{ github.event.repository.name }}
+          path: ${{ github.event.repository.name }}
+
+      - name: Display structure of downloaded files
+        run: ls -R
+        working-directory: ${{ github.event.repository.name }}
+
+      - name: Run plugin check
+        uses: wordpress/plugin-check-action@v1
```
jeffpaul, line 35: Are there any options to know specifically what checks we're running or is it all-or-nothing?

dkotter: By default it will run all tests but you can exclude certain checks or certain categories of checks. If you see a few lines below, we're using the `exclude-checks` setting to remove a couple checks but we run everything else.

You can see all the checks and categories here: https://github.com/wordpress/plugin-check-action/?tab=readme-ov-file#supported-checks

jeffpaul:
Looks good to me, though we might suggest an upstream enhancement to have some sort of simple report post-run to see what's passed/failed as it wasn't immediately clear where that info was in action run.

jeffpaul (dismissed):

.github/workflows/plugin-check.yml
dkotter:
```suggestion
    uses: 10up/classifai/.github/workflows/build-release-zip.yml@develop
```

PR #808, opened 2024-09-26
Add a PHPStan configuration and run it as a GitHub Action

dkotter:
### Description of the Change

This PR adds in PHPStan as a dev dependency, along with the configuration to run that and a GitHub Action Workflow that will run on all PRs. I've also fixed up all issues that have been reported.

Right now I have this set to level 1 (lowest is 0, highest is 9). If I change this to level 2, there are 85 new things that get flagged (though most look to be pretty similar so solving a few should solve most of them). Ideally we bump this up to a higher level but I think starting at 1 is better than nothing.

### How to test the Change

Verify the PHPStan Action [runs and passes on this PR](https://github.com/10up/classifai/actions/runs/11061912553?pr=808)

### Changelog Entry

> Fixed - Addressed all issues raised by running PHPStan
> Developer - Added PHPStan into our workflow with the ability to run that locally and in a GitHub Action

### Credits

Props @dkotter

### Checklist:

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -36,6 +36,7 @@ LICENSE.md export-ignore
 package-lock.json export-ignore
 package.json export-ignore
 phpcs.xml.dist export-ignore
+phpstan.neon.dist export-ignore
 phpunit.xml.dist export-ignore
 README.md export-ignore
 webpack.config.js export-ignore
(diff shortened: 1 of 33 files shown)


iamdharmesh (approved):
Thanks for adding this @dkotter. LGTM.

PR #811, opened 2024-09-27
Add the new OpenAI Moderation model

dkotter:
### Description of the Change

In #531 we added a Feature to moderate comments using the OpenAI Moderation tool. They recently released a new Moderation model (`omni-moderation-latest`, see https://openai.com/index/upgrading-the-moderation-api-with-our-new-multimodal-moderation-model/) so this PR ensures we are using that model. In addition, added in a helper method (similar to what we do in other places) making it easy to filter that model if someone wants to use the older one.

Note: this new moderation model is multi-modal, which means it supports images. I've not added any new functionality in this PR but this new model would open up other possibilities, like moderating image uploads as an example.

### How to test the Change

1. Enable and configure the Moderation Feature
2. Enable comments on your site
3. Post a comment with hateful language in it and ensure the comment is saved but is set to a Pending status
4. With the Feature disabled, post another hateful comment and see that it doesn't get set to Pending
5. Go to the comments list in the admin (`wp-admin/edit-comments.php`) and find the new comment. Using the bulk dropdown or the inline links, send the comment to moderation
6. Ensure it gets set to a Pending status

### Changelog Entry

> Changed - Use the new OpenAI Moderation model, `omni-moderation-latest`, in our comment moderation Feature 
> Developer - New filter, `classifai_openai_moderation_model`, allowing you to change the moderation model

### Credits

Props @dkotter

### Checklist:


jeffpaul:
@dkotter are there any changes to the moderation flags that the newer OAI model categorizes moderated comments into or are they the same as before?

dkotter:
> @dkotter are there any changes to the moderation flags that the newer OAI model categorizes moderated comments into or are they the same as before?

There are two new categories it will look for: `illicit` and `illicit/violent` but the rest are the same

iamdharmesh (approved):
The code looks good to me and it tests well. Thank you @dkotter

Issue #502, opened 2023-06-20
Onboarding and settings refactor to React


jeffpaul:
### Is your enhancement related to a problem? Please describe.

The new user experience for onboarding and settings pages was released in [2.1.0](https://github.com/10up/classifai/releases/tag/2.1.0), but could be more easily extended as we add new service providers, new features, and even more so when we add a new service provider for an existing feature.  This tracking issue will cover the refactoring to React for the onboarding/settings experiences.  We'll also want to reduce the heavy, primary references to service providers during onboarding and more on the specific feature being enabled/configured.  We can include a note in the UI to the effect of "_Powered by <Service Provider Name>_" and only force a user to consider which service provider to select if there are multiple to choose from for a given feature.

On top of this refactoring, we'll also want to consider the following items as related requirements to this work:
- [x] #306
- [x] #449
- [ ] #50

### Designs

TBD

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

dkotter:
Nice guide that we may learn from as far as how to use the built-in WordPress React components for settings pages: https://developer.wordpress.org/news/2024/03/26/how-to-use-wordpress-react-components-for-plugin-pages/

dkotter:
Another guide around using custom SlotFills may be helpful: https://developer.wordpress.org/news/2024/03/26/how-to-use-wordpress-react-components-for-plugin-pages/#setting-up-the-menu-and-settings-page

jeffpaul:
Noting that as part of this, it would be good to assess the WP Core WIP Admin update Figma to see where there are missing components/etc in those designs compared to what we're planning for ClassifAI's settings refactor.

PR #819, opened 2024-10-16
Add Chrome AI as a Provider for our text generation Features

dkotter:
### Description of the Change

Chrome is experimenting with adding [built-in AI](https://developer.chrome.com/docs/ai/built-in) to their browser. The model they include is fairly lightweight and doesn't give quite as good results as larger models, but since it's built-in, there's no cost and all data stays private (also allows easy testing of some AI features without having to sign up for something).

At the moment, this feature is still being worked on and is only available in Chrome Canary, so there's a chance this never releases publicly or it may change quite a bit before the public release.

This PR integrates this browser AI API into our Title Generation, Excerpt Generation and Content Resizing Features. Because this API is 100% javascript, the way we integrate is a little unique compared to our other Providers:

- We still register a new Provider for Chrome AI
- This Provider still listens for REST requests on the same endpoint as our other Providers
- Instead of sending off API requests when a REST request comes in, we return the prompt and content
- I've added a new helper JS method that checks if the Chrome AI API is available
- This helper method is now used if the REST request returns the above data instead of the existing data structure. The AI request is then made at the javascript level, passing in the prompt and content returned from the REST request

### How to test the Change

1. Install Chrome Canary v128.0.6545.0 or newer
2. Open a new tab in Chrome, go to `chrome://flags/#optimization-guide-on-device-model` and select Enabled BypassPerfRequirement
3. Go to `chrome://flags/#prompt-api-for-gemini-nano` and select Enabled
4. Relaunch Chrome
5. Open DevTools and send `(await ai.languageModel.capabilities()).available;` in the console
6. If this returns “readily”, then you are all set
7. If it fails, force Chrome to recognize that you want to use this API. To do so, open DevTools and send  `await ai.languageModel.create();` in the console. This will likely fail but it’s intended
8. Relaunch Chrome
9. Open a new tab in Chrome, go to `chrome://components`
10. Confirm that Gemini Nano is either available or is being downloaded. You'll want to see the `Optimization Guide On Device Model` present with a version greater or equal to 2024.5.21.1031. If there is no version listed, click on Check for update to force the download
11. Once the model has downloaded and has reached a version greater than shown above, open DevTools and send `(await ai.languageModel.capabilities()).available;` in the console. If this returns “readily”, then you are all set
12. Within the Title Generation, Excerpt Generation and Content Resizing Features, select Chrome AI as the Provider
13. Test that all three Features still work as expected. Note I had to massage the prompts a bit to get better results but as long as you're seeing results show up, things are working

### Changelog Entry

> Added - Chrome AI as a new text generation Provider, allowing completely on-device AI usage

### Credits

Props @dkotter, @jeffpaul, @iamdharmesh, @faisal-alvi 

### Checklist:


includes/Classifai/Features/ExcerptGeneration.php
```diff
@@ -45,6 +46,7 @@ public function __construct() {
 			ChatGPT::ID   => __( 'OpenAI ChatGPT', 'classifai' ),
 			GeminiAPI::ID => __( 'Google AI (Gemini API)', 'classifai' ),
 			OpenAI::ID    => __( 'Azure OpenAI', 'classifai' ),
+			ChromeAI::ID  => __( 'Chrome AI', 'classifai' ), // TODO: only add this Provider if the browser supports AI.
```
dkotter: Right now this Provider shows up even if your browser doesn't support Chrome AI. Would be great to remove it but doesn't cause any problems if you do select it. With the settings refactor in #502, I think will be much easier to remove this since all settings are registered in javascript at that point, so may leave this until it's integrated in there

iamdharmesh: > Right now this Provider shows up even if your browser doesn't support Chrome AI. Would be great to remove it

Maybe keep showing it and display the steps of how to activate it (Like you provided in test steps)?

src/js/helpers.js
```diff
@@ -63,3 +63,57 @@ export const handleClick = ( {
...
+	) {
+		const session = await window.ai.languageModel.create( {
+			initialPrompts: [
+				{
+					role: 'system',
+					content: prompt,
+				},
+			],
+		} );
+		result = await session.prompt( `"""${ content }"""` );
```
iamdharmesh: I came across some random errors while testing scenarios. We should consider adding error handling here to inform the user about failures. Additionally, we need to set a limit on the length of the content being sent in the prompt (the limit on my browser model appears to be 6,144 tokens).

image

iamdharmesh:
@dkotter I did a quick check, and it works fine for me. I think once we have error handling in place, this will be ready to merge. Thanks for working on this. Great work!!

includes/Classifai/Features/ExcerptGeneration.php
iamdharmesh: @dkotter Updated as follows:  
- The Chrome AI option will be completely removed if `window.ai` is undefined (e.g., in browsers other than Chrome).  
- If built-in AI is not supported, "Not supported" will be displayed along with a link to documentation on how to enable AI in the browser.  
Screenshot 2024-12-16 at 10 52 06 PM

- Built-in AI is supported.
Screenshot 2024-12-16 at 10 52 30 PM

Please let me know if this looks good to you or if any updates are needed here.

dkotter: Looks great! Was actually working on something similar at the same time but this is better than what I had

src/js/helpers.js
dkotter: I've added a token limit now to the Provider, so we trim the content we use if we are above the limit. I've also added some additional error handling, just outputting an `alert` if something goes wrong. Not the best UX but for this Chrome AI that's still under development, I think that's good enough for now.

iamdharmesh (dismissed):
@dkotter This looks good to me and is ready for the merge. I'm just curious if we should add any note that Chrome AI is still in the experiment, though most users will know it when they attempt to enable it.

src/js/helpers.js
iamdharmesh: Yes, good enough for now.

PR #829, opened 2024-11-26
Update from v3.2 to v4.0 of the Azure AI Vision API

### LINKED ISSUE #827: Update Azure AI Vision from 3.2 to 4.0

Related to #826 and alongside adding OpenAI to alt text generation, let's look to update the Azure API version in ClassifAI.

While there are some differences from the features that are / are not available in 4.0 (which we'll want to validate ClassifAI features that are available in v4), let's look to [update to 4.0](https://learn.microsoft.com/en-us/azure/ai-services/computer-vision/overview-image-analysis?tabs=4-0#image-analysis-versions) where that version includes coverage for specific ClassifAI image processing features using Azure.

Version 4.0 features available: Read text, Captions, Dense captions, Tags, Object detection, Custom image classification / object detection, People, Smart crop
> Better models; use version 4.0 if it supports your use case.

Version 3.2 features available: Tags, Objects, Descriptions, Brands, Faces, Image type, Color scheme, Landmarks, Celebrities, Adult content, Smart crop
> Wider range of features; use version 3.2 if your use case is not yet supported in version 4.0

Additional context from Azure:
> We recommend you use the Image Analysis 4.0 API if it supports your use case. Use version 3.2 if your use case is not yet supported by 4.0.

> You'll also need to use version 3.2 if you want to do image captioning and your Vision resource is outside the supported Azure regions. The image captioning feature in Image Analysis 4.0 is only supported in certain Azure regions. Image captioning in version 3.2 is available in all Azure AI Vision regions. See [Region availability](https://learn.microsoft.com/en-us/azure/ai-services/computer-vision/overview-image-analysis#region-availability).

### Designs

_No response_

_No response_

### Code of Conduct

dkotter:
### Description of the Change

In #559, we switched over to using the Azure AI Vision v3.2 API for all Features relying on that. We decided not to switch to the v4.0 of that API as it was still in public preview and had some breaking changes.

That API seems to be more stable now so this PR switches over to that for the following Features:

- Descriptive Text Generator
- Image Tags Generator
- Image Text Extraction

It does not change the following Features:

- Image Cropping: the v4.0 API is fairly different (doesn't actually return a cropped image but returns the image coordinates that need cropped) and will require additional work
- PDF Text Extraction: this doesn't exist in the v4.0 API but has been moved to an entirely new API, [Azure AI Document Intelligence](https://azure.microsoft.com/en-us/products/ai-services/ai-document-intelligence), so will look to tackle that in a separate PR

Things to note:

- the v4.0 API supports images up to 20MB, up from previous of 4MB and larger dimensions, up to 16000x16000px
- Image Text Extraction (OCR) used to be two separate API requests. That can all be done in the v4.0 API so the code for this has been simplified (we've removed the OCR class entirely)
- The v4.0 API has less regions supported, in particular for the `captions` feature, which is used for the Descriptive Text Generator Feature. See https://learn.microsoft.com/en-us/azure/ai-services/computer-vision/overview-image-analysis?tabs=4-0#region-availability for that list
- Also seems the v4.0 API has fixed confidence scores. In v3.0, we recommended a threshold of 70-75%. In v3.2, we saw scores drop to 50-55% and used that as our recommendation. In testing multiple images, it seems 70% is again a good default so that has been updated

In addition, we now output an error message if a valid caption is returned but the confidence score is lower than our threshold. Previously we would just silently discard that, which can lead to people thinking things aren't working. We still don't save that caption but we show an error letting the user know what happened.

Partially closes #827

For some tests, here's some results I got:

| Image | v3.2 Caption | v3.2 Confidence Score |  v4.0 Caption |  v4.0 Confidence Score |
|----|------|------|------|------|
| A scientist with a microscope | a woman wearing a white coat and white lab coat sitting at a desk  | 36.85% | a woman in a lab coat and gloves holding a pen and looking at a microscope | 70.76% |
| Stop sign | a stop sign with a cloudy sky  | 49.51% | a stop sign with clouds in the background | 82.06% |
| A dog | a dog with its mouth open  | 57.06% | a dog with its mouth open | 83.40% |

You could argue on if these captions are better or not but they are definitely not worse and the confidence scores are back to being more realistic, which is great as that's an issue that trips up a lot of people

### How to test the Change

1. Setup the Descriptive Text Generator Feature with Azure AI. Ensure it works as expected
2. Setup the Image Tags Generator Feature with Azure AI. Ensure it works as expected
2. Setup the Image Text Extraction Feature with Azure AI. Ensure it works as expected

### Changelog Entry

> Changed - Migrate from the Azure AI Vision v3.2 API to the v4.0 API

### Credits

Props @dkotter, @jeffpaul 

### Checklist:


iamdharmesh (approved):
Thanks @dkotter. PR looks good to me and it tests well.

PR #838, opened 2024-12-23
Add a Snackbar notice to inform users that settings have been saved successfully.

### LINKED ISSUE #837: Scroll to notification messages when settings are saved

With the new settings UI that was launched in v3.2.0, we no longer have page refreshes when settings are saved. What this means though is on longer settings pages, when someone saves, they have to scroll up manually to see the notification message. This makes it hard to know if things saved successfully or if something went wrong.

I can see two approaches here:

1. Show these messages closer to the save button
2. Automatically scroll up to the message when a save happens

### Designs

_No response_

_No response_

### Code of Conduct

iamdharmesh:
### Description of the Change
This PR adds a Snackbar notice to inform users that settings have been saved successfully.

Dec-23-2024 22-31-06

Closes #837 

### How to test the Change
1. Navigate to the ClassifAI settings page (Tools > ClassifAI).
1. Enable or disable any feature and confirm that the notice is displayed properly.
1. Navigate to the settings page of any feature and save the settings.
1. Validate that the Snackbar notice confirming the settings have been saved is displayed.

### Credits
Props @iamdharmesh @dkotter @jeffpaul 

### Checklist:

### DIFF
--- a/src/js/settings/components/classifai-registration/index.js
+++ b/src/js/settings/components/classifai-registration/index.js
@@ -201,6 +201,10 @@ export const SaveSettingsButton = ( {
 					);
 					setSettings( res.settings );
 					setIsSaving( false );
+					window.scrollTo( {
+						top: 0,
+						behavior: 'smooth',
+					} );
 					return;
 				}
 
@@ -226,6 +230,10 @@ export const SaveSettingsButton = ( {
 					}
 				);
 				setIsSaving( false );
+				window.scrollTo( {
+					top: 0,
+					behavior: 'smooth',
+				} );
 			} );
 	};
 
--- a/src/js/settings/components/classifai-registration/index.js
+++ b/src/js/settings/components/classifai-registration/index.js
@@ -174,7 +174,8 @@ export const SaveSettingsButton = ( {
 	setSettings,
 	onSaveSuccess = () => {},
 } ) => {
-	const { createErrorNotice, removeNotices } = useDispatch( noticesStore );
+	const { createErrorNotice, createSuccessNotice, removeNotices } =
+		useDispatch( noticesStore );
 	const notices = useSelect( ( select ) =>
 		select( noticesStore ).getNotices()
 	);
@@ -205,6 +206,12 @@ export const SaveSettingsButton = ( {
 
 				setSettings( res.settings );
 				onSaveSuccess();
+				createSuccessNotice(
+					__( 'Settings saved successfully.', 'classifai' ),
+					{
+						type: 'snackbar',
+					}
+				);
 				setIsSaving( false );
 			} )
 			.catch( ( error ) => {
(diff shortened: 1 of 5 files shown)


dkotter (commented):
Overall this works nice and code looks good here. Couple things worth discussing:

1. Is it possible to reduce the time the notice shows? Right now the snackbar sticks around for quite a while and I think it would be nice if we could hide it quicker
2. This solves success messages but doesn't solve error messages. If you save settings for a Feature and an error occurs, the error shows up top but it is often out of view. Would be great if we scroll to the top when that happens so the error can be seen

jeffpaul:
Note that there's discussion in WP/GB via https://github.com/WordPress/gutenberg/issues/16391 on overall accessibility of the snackbar, so we might perhaps review the consensus from a11y folks there and ensure we utilize the snackbar for error & success states in a solid a11y manner.

jeffpaul:
Looks like https://github.com/WordPress/gutenberg/issues/67662 is an updated proposal in WP/GB that we might consider here as well.

iamdharmesh:
@dkotter I have added a scroll-to-top to make error notices more visible. Regarding the Snackbar notice, as @jeffpaul pointed out, it has open accessibility issues. Since we have already added the scroll-to-top, I think we can replace it with a standard success notice combined with the scroll-to-top. However, I am not too concerned about this since the message we are displaying is not high priority. Additionally, the core itself uses Snackbar notices, and I have seen some plugins, such as ElasticPress, doing the same. what do you think?

dkotter:
>  Additionally, the core itself uses Snackbar notices, and I have seen some plugins, such as ElasticPress, doing the same. what do you think?

I'm fine with using the Snackbar here. We can always change this once there's a better option in Core.

> I have added a scroll-to-top to make error notices more visible.

Thanks, this works great.

One last thing is how long the snackbar shows. If we can't reduce the time that shows (not sure if that's a setting that is supported by Core) I think we should at least remove the snackbar (if this is easy enough) anytime the screen changes.

As an example, if I enable a Feature, the snackbar shows. If I then click in to configure the settings for that Feature, ideally we remove the snackbar at that point.

Then when settings are saved, we show the snackbar again. If I go back to the main screen, that snackbar should be removed again.

iamdharmesh:
>One last thing is how long the snackbar shows. If we can't reduce the time that shows (not sure if that's a setting that is supported by Core) I think we should at least remove the snackbar (if this is easy enough) anytime the screen changes.

It seems the time limit is [hardcoded](https://github.com/WordPress/gutenberg/blob/trunk/packages/components/src/snackbar/index.tsx#L29). We may be able to reduce the time by setting a timeout function on our side, but I think removing the snackbar on screen change looks more appropriate here. I have updated the code to remove notices on location change. Could you please help review it once? Thanks.

dkotter (approved):

PR #839, opened 2025-01-03
Fix typos

szepeviktor:
### Description of the Change

Introduce https://github.com/crate-ci/typos to ClassifAI.

### DIFF
--- a/.typos.toml
+++ b/.typos.toml
@@ -1,0 +1,23 @@
+[files]
+extend-exclude = [
+    ".git/",
+    "tests/test-plugin/text-to-speech-voices.json"
+]
+ignore-hidden = false
+
+[default]
+extend-ignore-re = [
+    "\\bTung Du\\b",
+]
+
+[default.extend-words]
+# Typos
+"weather" = "whether"
+
+[default.extend-identifiers]
+# Typos
+"Github" = "GitHub"
+"Wordpress" = "WordPress"
+"Woocommerce" = "WooCommerce"
+"get_post_stati" = "get_post_stati"
+"Automattic" = "Automattic"
--- a/includes/Classifai/Features/Classification.php
+++ b/includes/Classifai/Features/Classification.php
@@ -458,7 +458,7 @@ class Classification extends Feature {
 			</label>
 		</p>
 
-		<div class="classifai-clasify-post-wrapper" style="display: none;">
+		<div class="classifai-classify-post-wrapper" style="display: none;">
 			<a href="<?php echo esc_url( wp_nonce_url( admin_url( 'admin-post.php?action=classifai_classify_post&post_id=' . $post->ID ), 'classifai_classify_post_action', 'classifai_classify_post_nonce' ) ); ?>" class="button button-classify-post">
 				<?php esc_html_e( 'Suggest terms & tags', 'classifai' ); ?>
 			</a>
--- a/src/js/features/classification/previewer.js
+++ b/src/js/features/classification/previewer.js
@@ -177,7 +177,7 @@ import './previewer.scss';
 		 * Filters response data depending on the threshold value.
 		 *
 		 * @param {Object} data       Response data from NLU.
-		 * @param {Object} thresholds Object containing threshold values for various taxnomy types.
+		 * @param {Object} thresholds Object containing threshold values for various taxonomy types.
 		 * @return {Array} Sorted data.
 		 */
 		function filterByScoreOrRelevance( data = {}, thresholds ) {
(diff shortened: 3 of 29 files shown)


.typos.toml
```diff
@@ -0,0 +1,23 @@
+[files]
+extend-exclude = [
+    ".git/",
+    "tests/test-plugin/text-to-speech-voices.json"
+]
+ignore-hidden = false
+
+[default]
+extend-ignore-re = [
+    "\\bTung Du\\b",
```
szepeviktor, line 10: Not `variable_names` go here.

.typos.toml
```diff
@@ -0,0 +1,23 @@
...
+]
+
+[default.extend-words]
```
szepeviktor, line 13: These are parts of variable names.

.typos.toml
```diff
@@ -0,0 +1,23 @@
...
+# Typos
+"weather" = "whether"
+
+[default.extend-identifiers]
```
szepeviktor, line 17: Complete variable names.

.typos.toml
```diff
@@ -0,0 +1,23 @@
...
+# Typos
+"Github" = "GitHub"
+"Wordpress" = "WordPress"
+"Woocommerce" = "WooCommerce"
+"get_post_stati" = "get_post_stati"
+"Automattic" = "Automattic"
```
szepeviktor, line 23: This means: DO NOT CORRECT

includes/Classifai/Features/Classification.php
```diff
@@ -458,7 +458,7 @@ public function render_meta_box( \WP_Post $post ) {
 			</label>
 		</p>
 
-		<div class="classifai-clasify-post-wrapper" style="display: none;">
+		<div class="classifai-classify-post-wrapper" style="display: none;">
```
dkotter, line 461: While this is a typo, because this is used as a class name, we'll need to decide if we want to leave this as-is, as others may be relying on this class for styling purposes and updating this will break things for them. Not overly concerned but something for us to consider

src/scss/admin.scss
```diff
@@ -870,7 +870,7 @@ input.classifai-button {
 }
 
 // Start: Classify Post Modal
-.classify-post-componenet {
+.classify-post-component {
```
dkotter, line 873: In doing a quick search to see if the `classify-post-componenet` class was used elsewhere (and thus may need updated) it doesn't appear so. Seems all other places use the proper class name. What this means though is this style has not been applying so we'll need to do some QA here to ensure this style doesn't cause any problems

src/js/features/classification/previewer.js
```diff
@@ -513,7 +513,7 @@ document.addEventListener( 'DOMContentLoaded', function () {
 	if ( classifaiNLUCheckbox ) {
 		classifaiNLUCheckbox.addEventListener( 'change', function () {
 			const classifyButton = document.querySelector(
-				'.classifai-clasify-post-wrapper'
+				'.classifai-classify-post-wrapper'
```
dkotter, line 516: If we decide we don't want to change this class name (mentioned in another comment) we'll want to revert this change as well

src/js/features/classification/taxonomy-controls.js
```diff
@@ -158,7 +158,7 @@ const TaxonomyControls = ( { onChange, query } ) => {
 								// Append newTerm to taxoInfo.terms.
 								const terms = {
 									...taxoInfo.terms,
-									entitites: [
+									entities: [
```
dkotter, line 161: It seems like this typo may have caused problems but would be curious to know if that was the case

szepeviktor: `typos` is a very useful tool. Not every day but the benefits are great on those days.

szepeviktor: BTW `tsc` from Typescript does perform [static analysis on JavaScript](https://github.com/szepeviktor/byte-level-care/issues/245) too.

src/scss/admin.scss
iamdharmesh: This class is not in use anywhere, so we can update it or possibly remove it entirely.

src/js/features/classification/taxonomy-controls.js
iamdharmesh: @dkotter, I just checked this. Due to a typo, the `terms` object was updated with a new key, but since we are not using the `entities` data, it hasn't caused any issues yet.

includes/Classifai/Features/Classification.php
iamdharmesh: Just adding additional details here that this is used in the classic editor only.

iamdharmesh (approved):
Thanks for all the help here @szepeviktor. LGTM. I have added GitHub Action workflow for the spelling check on PR.

@dkotter sending this to you for another review. Thanks.

dkotter (approved):

szepeviktor:
Glad to contribute.

PR #845, opened 2025-01-31
Add Ollama as a Provider, allowing some Features to work with locally hosted LLMs

### LINKED ISSUE #772: Integration with open source, local LLMs

There are varying options on open source, local LLMs that we could look to integrate with ClassifAI.  These come on the market with some regularity, so we might want to consider not just direct integration with individual models but perhaps a middleware layer that would then integrate with new OSS models as the become available.  This would allow ClassifAI users to have a free, open source option for a service provider that would also allow for controlling where their site data goes.  So a data privacy and cost perspective win.

### Designs

_No response_

No immediate options that I'd recommend we start with, so some exploration on options would be best before proceeding with any integrations.

### Code of Conduct

dkotter:
### Description of the Change

[Ollama](https://ollama.com/) allows you to easily run various LLMs on your own computer. This provides a couple key benefits:

1. Can test these LLMs without any worry about cost
2. No concerns with data privacy as all requests stay on your own computer

This PR integrates Ollama with a number of our existing Features, depending on which model you use:

If using a standard model:

- Title Generation
- Excerpt Generation
- Content Resizing
- Key Takeaways

If using a vision model:

- Descriptive Text Generator
- Image Tags Generator
- Image Text Extraction

If using an embedding model:

- Classification

This allows you to test quite a few of the Features ClassifAI provides without any cost or data concerns. This could also be used on production sites, though worth noting a few downsides to running these models locally:

- Performance is slower than using an externally hosted LLM, like OpenAI. This is especially true if you don't use the model often, as Ollama has to load that model for the first request, then stores that model in memory for a short time. So if you're doing multiple requests close together, the first request is fairly slow and the ones after actually return pretty fast
- While these "open source" models are getting better and better, quality still doesn't quite match models from OpenAI (or competitors). This is especially true as you typically need to run the smaller versions of these models locally, due to memory requirements for the full size models

That said, there really isn't any reason you couldn't use Ollama as a Provider on an actual production site. You'd just need to ensure that any user that wants to use those Features has Ollama installed and configured on their individual computers.

Closes #772 

### How to test the Change

1. Download and install [Ollama](https://ollama.com/)
2. Install the models you want to use. I'd suggest `ollama pull llama3.1` as a base LLM; `ollama pull nomic-embed-text` as the embedding LLM; `ollama pull llava` as the vision LLM but can see all available models [here](https://ollama.com/search)
3. Once Ollama is running and models are installed, configure the following Features to use Ollama with the proper model selected: 
- Classification
- Title Generation
- Excerpt Generation
- Content Resizing
- Key Takeaways
- Descriptive Text Generator
- Image Tags Generator
- Image Text Extraction
4. Ensure each Feature works as expected. Note for any Image Processing Features, the image needs to be publicly available to work, so typically won't work on local environments. Quick fix for testing is to hardcode an external image URL in each of those callbacks to test

### Changelog Entry

> Added - Integration with Ollama as a Provider, allowing the following Features to use locally hosted LLMs: Classification; Title Generation; Excerpt Generation; Content Resizing; Key Takeaways; Descriptive Text Generator; Image Tags Generator; Image Text Extraction
> Changed - Bump WordPress minimum from 6.5 to 6.6

### Credits

Props @dkotter

### Checklist:


dkotter:
There's a few things that I've captured while working on this that I think would be good followups:

1. This PR only adds support for Ollama. Originally was going to also add support for [GPT4All](https://www.nomic.ai/gpt4all) and [LMStudio](https://lmstudio.ai/) but with the size of this PR already, decided it was best to handle those separately, if we think those would be worth adding. Both give an actual UI, which can be nice, compared to Ollama which is all command line based
2. Added support for the Classification Feature using embeddings models from Ollama but have not added support for other Features that use embeddings, like the Smart 404 and Term Cleanup Features. Would be nice to look at adding support for those in followup PRs
3. We added support for OpenAI Vision models for the Descriptive Text Generator Feature in #828 but didn't add support for any other Image Processing. In this PR, I added support for Descriptive Text Generator, Image Tags Generator and Image Text Extraction when using vision models. Would be nice for a followup PR to add OpenAI as a Provider for Image Tags Generator and Image Text Extraction

jeffpaul:
@vikrampm1 would you please help open issues for the 3 items above and drop into the 3.4.0 milestone?

src/js/settings/components/provider-settings/ollama-base.js
```diff
@@ -0,0 +1,98 @@
...
+	if (
+		providerSettings?.models &&
+		! Array.isArray( providerSettings.models )
+	) {
+		for ( const [ key, value ] of Object.entries(
+			providerSettings.models
+		) ) {
+			models.push( { label: value, value: key } );
+		}
+	}
```
iamdharmesh, line 64: Optional note: How about loading models by making an API call if provider settings don't have models saved or the endpoint URL is updated by the user in settings? This way, we don't need to save settings to populate models during first-time setup.

iamdharmesh (dismissed):
This looks great and works really well, @dkotter! Thanks a lot for all your great work here.

src/js/settings/components/provider-settings/ollama-base.js
dkotter: Ah, sorry, missed this in fixing a few other things. I do think either a refresh button or maybe just always pulling in the current models when on the settings page would be a nice enhancement here. Would make it easier to see new models after you install them

PR #858, opened 2025-02-27
Add embedding similarity hooks

dkotter:
### Description of the Change

When using an embedding model to run classification, we don't store the full similarity results and there's currently no easy way for someone to store that themselves if they want to use those results for something else.

This PR introduces a few new hooks allowing you to do just that:

- After embedding similarity has been run for an individual post embedding -> term chunk embedding, a new hook fires that allows you to do whatever you want with those results (like store them in post meta).
- After the full embedding similarity results are done, a new hooks fires that allows you to do whatever you want with those results
- And then another hook fires after we sort the results, allowing you to do whatever you want with those results

### How to test the Change

If desired, can add a custom plugin that uses one or more of these new hooks to ensure they work but mostly just need a code review here

### Changelog Entry

> Added - New hooks that allow extra processing of embedding similarity calculations

### Credits

Props @dkotter, @tylercherpak 

### Checklist:

### DIFF
--- a/includes/Classifai/Providers/Azure/Embeddings.php
+++ b/includes/Classifai/Providers/Azure/Embeddings.php
@@ -481,6 +481,19 @@ class Embeddings extends OpenAI {
 			return new WP_Error( 'invalid', esc_html__( 'No matching terms found.', 'classifai' ) );
 		}
 
+		/**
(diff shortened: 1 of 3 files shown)


iamdharmesh (approved):
Code LGTM and it tests well. Thanks @dkotter

PR #859, opened 2025-02-28
Add new Content Generation Feature

dkotter:
### Description of the Change

This PR adds in a new Feature to Generate Content, meant to be used to create initial article drafts. Added OpenAI, Azure OpenAI and Ollama as Providers for this Feature, but can add our other generative AI Providers in the future (Gemini, Grok).

A new UI was created in #864 that this Feature utilizes, and other Features in the future can use as well.

It's currently setup where you enter in a summary of what you want written and the LLM will provide a full length article. It also provides the ability to iterate on the article the LLM sends back, if for instance you want to request changes like increasing the length or changing the tone. Once you're happy, you can insert the content into the block editor, overwriting any existing content.

Here's a more detailed step-by-step of how it works:

1. Enable the Feature then create a new Post
2. If desired, add a title to this post (will help with results)
3. In the bottom right corner of the editor, click the new sparkle icon
4. This will open a modal where you can enter in the summary of the article you want written. Once ready, click `Send` and this will send that summary off to the LLM (along with our base prompt)
5. Once results are returned, the UI is updated to show the summary you provided and the results the AI provided. You can then choose to insert the content, which will overwrite any content you may have had in the block editor with this generated content
6. If you don't like the results returned, you can click a `Start over` button which will keep the modal open but clear everything out, allowing you to enter in a new summary
7. You can also copy the results if you want to manually paste that somewhere
8. If the results returned are close to what you want but you'd like to iterate a bit, you can add additional instructions in the textarea, like `Make the content longer/shorter` or `Change the tone of the content to be more professional`
9. Submitting this new instruction will send the original base prompt, the original summary and the first content AI returned, along with the new instructions. Once results come back, we render these results in the modal along with any previous results (trying to have a UI similar to what ChatGPT provides, where you can see your requests along with the AI responses)
10. You can iterate as many times as you want, though eventually you'll run into token limits with the context size

Partially closes #841. This Feature does solve the `Generate draft post content from simple input` but does not do anything in regards to `Ensuring the generated drafts align closely with the publication’s tone and style`. A follow up PR will be created to help with that piece

### Screenshots

New sparkle icon to initiate content generation

After clicking the new sparkle icon

Waiting for the AI to respond

Reviewing the content AI provided

Available action buttons

Iterating on the content with AI

### How to test the Change

Follow the details steps above

### Changelog Entry

> Added - New Content Generation Feature, allowing the creation of draft content from a summary

### Credits

Props @dkotter, @jeffpaul, @fabiankaegy 

[...]

PR #860, opened 2025-02-28
Fix Gutenberg deprecation warnings

dkotter:
### Description of the Change

As of the last release our minimum supported version of WordPress is 6.6. There were quite a few [deprecation warnings added](https://make.wordpress.org/core/2024/06/18/editor-unified-extensibility-apis-in-6-6/) in that version that we've previously been ignoring but we can now safely update those to the proper components to fix those.

In addition, while fixing those up and testing to ensure everything worked, found a few other things I thought worth fixing:

- In both our title generation and excerpt generation Features, we grab the content and/or title of the item to send along. But we grab this on page load so if someone were to update the content and then click the button, the most recent version of the content isn't used. This has been fixed
- Took some learnings from #859 to make the modal we show when generating titles a bit nicer. It's now wider and if generating multiple titles, those end up in columns instead of stacked

**Before**

Generate title modal before changes

**After**

Generate title modal after changes

### How to test the Change

Minimal changes made across multiple Features so ideal is to test to ensure those Features still work and no WordPress 6.5 or 6.6 deprecation warnings are shown. Note there are some deprecation notices introduced in 6.7 around bottom margin styles being deprecated since version 6.7 and will be removed in version 7.0 but decided not to address those yet

### Changelog Entry

> Changed - Update the layout of the generate titles modal
> Fixed - Fix various Gutenberg deprecation notices added in WordPress 6.5 and 6.6
> Fixed - Ensure we pass the most recent version of the content in our title and excerpt generation Features

### Credits

Props @dkotter

### Checklist:

### DIFF
--- a/src/js/features/classification/pre-publish-classify-post.js
+++ b/src/js/features/classification/pre-publish-classify-post.js
@@ -3,7 +3,7 @@
  */
 import { __ } from '@wordpress/i18n';
 import { withSelect } from '@wordpress/data';
-import { PluginPrePublishPanel } from '@wordpress/edit-post';
+import { PluginPrePublishPanel } from '@wordpress/editor';
 import { Component } from '@wordpress/element';
 
 /**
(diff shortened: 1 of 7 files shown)


iamdharmesh (approved):
Thanks for the PR @dkotter. LGTM

PR #862, opened 2025-03-04
Add ability to only allow manual processing of our various image Features

dkotter:
### Description of the Change

Right now most of our Image Processing Features automatically trigger when an image (or PDF) is uploaded. While this can be nice as it means things just magically happen without any extra effort, it does mean every single upload goes through this flow.

There's very likely a scenario where someone doesn't want an image to be processed. Also, because of the extra processing that happens (sending this image off to AI and waiting for the response), it makes the upload process slower. I could see a situation where someone wants uploads to be as fast as possible but still allow images to be manually processed after the fact (**side note**, there is a potential feature change we could look at where we change our image processing to fire after the image has been fully uploaded instead of blocking that process).

Similar to how we introduced a mode toggle in our Classification Feature, this PR adds in a new `Processing mode` setting to each of our `Image Processing` Features that have extra processing (all but `Image Generation`). This will default to `Automatically on upload` (which is the current behavior) but can be changed to `Manually trigger`, which won't process at all unless someone requests it at the individual media level.

Worth noting that you can still trigger things manually even if the mode is set to automatic.

### Screenshots

New Processing mode setting

### How to test the Change

1. Check out this PR
2. Go to the `Descriptive Text Generator`, `Image Tags Generator`, `Image Cropping`, `Image Text Extraction` or `PDF Text Extraction` Features
3. Ensure the `Processing mode` settings shows for each and is defaulted to `Automatically on upload`
4. Test each Feature to ensure it still works as expected
5. Change the setting to be `Manually trigger` and test again, ensuring no processing happens until it's triggered on an individual media item

### Changelog Entry

> Added - New Processing Mode setting to our Image Processing Features, allowing you to only have those Features trigger when someone manually requests it, rather than automatically on upload

### Credits

Props @dkotter

### Checklist:

### DIFF
--- a/includes/Classifai/Features/DescriptiveTextGenerator.php
+++ b/includes/Classifai/Features/DescriptiveTextGenerator.php
@@ -390,6 +390,7 @@ class DescriptiveTextGenerator extends Feature {
 				'caption'     => 0,
 				'description' => 0,
 			],
+			'processing_mode'         => 'automatic',
 			'provider'                => ComputerVision::ID,
 		];
 	}
@@ -444,6 +445,8 @@ class DescriptiveTextGenerator extends Feature {
 		$settings = $this->get_settings();
 
 		$new_settings['descriptive_text_fields'] = array_map( 'sanitize_text_field', $new_settings['descriptive_text_fields'] ?? $settings['descriptive_text_fields'] );
+
+		$new_settings['processing_mode'] = sanitize_text_field( $new_settings['processing_mode'] ?? $settings['processing_mode'] );
 
 		return $new_settings;
 	}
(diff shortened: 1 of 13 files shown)


jeffpaul:
> side note, there is a potential feature change we could look at where we change our image processing to fire after the image has been fully uploaded instead of blocking that process

This feels like a worthwhile update, perhaps as an additional option for Processing mode of "Automatic after upload" with the current default iterating to "Automatic during upload" perhaps?

dkotter:
> This feels like a worthwhile update, perhaps as an additional option for Processing mode of "Automatic after upload" with the current default iterating to "Automatic during upload" perhaps?

I agree, this would definitely be a nice follow up enhancement here. I do think some thought will need to be put into the UX in this scenario though. For instance, if we send off the image for processing async, very likely someone may upload an image and then click in to edit the image and no alt text (or description or tags, etc) will have returned yet. Seems like we should at least show a message letting them know processing is happening. Potentially even preventing updates until processing is finished (i.e. disable the alt input)?

jeffpaul:
FYI I created #863 to try and provide an option for follow-up after this feature lands.

Sidsector9 (approved):
The following works as expected ✅ 
- Descriptive Text Generator
- Image Tags Generator
- Image Cropping
- PDF Text Extraction
- Image Text Extraction

PR #871, opened 2025-03-13
[Enhancement] Support WooCommerce Product-Based Title & Excerpt Generation

### LINKED ISSUE #98: WooCommerce product tagging
I received the following via our ClassifAI web form registration and wanted to capture the idea here for future consideration.

"WooCommerce product tagging would also be a benefit."

**Designs**
n/a

none.

**Additional context**
n/a

faisal-alvi:
### Description of the Change

This PR enhances the title and excerpt generation feature to support WooCommerce products. It fetches key product details such as name, type, SKU, categories, attributes, price, stock status, and descriptions to generate optimized and high-converting titles and excerpts tailored for e-commerce. The implementation ensures compatibility with existing functionality while adding WooCommerce-specific logic for improved accuracy and relevance.

Closes #98 (Yet to determine if Woo Content & Tagging Generations will be included in this PR or handled separately).

### How to test the Change

1. Install and activate the WooCommerce plugin if not already enabled.
2. Navigate to Products → Add New or edit an existing product.
3. Verify Title & Excerpt Generation.
4. Check Behavior for Non-Product Posts.

### Changelog Entry

> Added - Support for WooCommerce product-based title and excerpt generation
> Changed - Bump WordPress minimum from 6.6 to 6.7

### Credits

### Checklist:


dkotter (changes requested):
Overall this looks good. Made a few minor changes and fixed a bug but there's still one major thing that needs fixed.

Generating titles and inserting those work and generating an excerpt works but inserting it does not. I think our JS that handles this needs updated to account for the TinyMCE editor Woo uses, whereas the normal excerpt input is just a textarea. Would also be nice to update the button text to make it clear it's generating a short product description and not an excerpt.

Also, if we could get some E2E tests here that would be great. Don't know if we need tests for each individual Provider but at least need one test to ensure titles can be generated and inserted on a product and that excerpts can be generated and inserted on a product (will need to install Woo to test that)

faisal-alvi:
@dkotter The newly added tests are passing, but a few of the old ones have started failing. I’m skipping resolving them here for now as they seem unrelated.

tests/bin/initialize.sh
```diff
@@ -2,3 +2,4 @@
 npm run env run tests-wordpress chmod -- -c ugo+w /var/www/html
 npm run env run tests-cli wp rewrite structure '/%postname%/' -- --hard
 npm run env run tests-cli wp plugin deactivate classic-editor
+npm run env run tests-cli wp plugin activate woocommerce
```
dkotter: Since we only need Woo activated when running a few tests, we should start with it deactivated and then only activate it when we need it (and then deactivate it again after the tests are done)

dkotter (changes requested):

dkotter:
> @dkotter The newly added tests are passing, but a few of the old ones have started failing. I’m skipping resolving them here for now as they seem unrelated.

Tests are currently passing on `develop` which implies the test failures are due to changes in this PR. I can also reproduce those failures locally. It does seem like an issue with having WooCommerce active. If I disable that, the tests start to pass. I think addressing my comment here: https://github.com/10up/classifai/pull/871#discussion_r2093629578 will fix those tests

dkotter (approved):

PR #873, opened 2025-03-18
Update all actions we use in our workflows to pull from specific pinned commits

dkotter:
### Description of the Change

In order to help protect against compromised actions, instead of including actions based on their major version (like v4), this PR switches all the actions we use to pull based on the commit hash from the latest release.

While this does impact maintenance a bit going forward, it ensures that we we're always using actions that we (hopefully) trust and if an action gets compromised (which [happens](https://semgrep.dev/blog/2025/popular-github-action-tj-actionschanged-files-is-compromised/)) we don't have to worry that we're using a compromised action. This also goes along with what [GitHub suggests](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions).

### How to test the Change

Ensure all our workflows still run as expected

### Changelog Entry

> Developer - Update all third-party actions our workflows rely on to use versions based on specific commit hashes.

### Credits

Props @dkotter, @jeffpaul 

### Checklist:

### DIFF
--- a/.github/workflows/plugin-check.yml
+++ b/.github/workflows/plugin-check.yml
@@ -19,10 +19,10 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
       - name: Download built zip
-        uses: actions/download-artifact@v4
+        uses: actions/download-artifact@b14cf4c92620c250e1c074ab0a5800e37df86765 # v4.2.0
         with:
           name: ${{ github.event.repository.name }}
           path: ${{ github.event.repository.name }}
@@ -32,7 +32,7 @@ jobs:
         working-directory: ${{ github.event.repository.name }}
 
       - name: Run plugin check
-        uses: wordpress/plugin-check-action@v1
+        uses: wordpress/plugin-check-action@70f38272b2beee386e973f319591baa2fc7dff16 # v1.1.2
         with:
           build-dir: ${{ github.event.repository.name }}
           exclude-checks: 'plugin_readme,plugin_updater' # Plugin isn't on .org so excluding these for now.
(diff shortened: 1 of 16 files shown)


.github/workflows/plugin-check.yml
```diff
@@ -32,7 +32,7 @@ jobs:
         working-directory: ${{ github.event.repository.name }}
 
       - name: Run plugin check
-        uses: wordpress/plugin-check-action@v1
+        uses: wordpress/plugin-check-action@70f38272b2beee386e973f319591baa2fc7dff16 # v1.1.2
```
jeffpaul, line 35: Noting that I think this one is still updating fairly regularly, but probably something we're ok trailing a bit on and just updating occasionally as they hit major updates

.github/workflows/stable.yml
```diff
@@ -28,7 +28,7 @@ jobs:
         npm run archive
 
     - name: Release to Stable
-      uses: s0/git-publish-subdir-action@develop
+      uses: s0/git-publish-subdir-action@ac113f6bfe8896e85a373534242c949a7ea74c98 # develop
```
jeffpaul, line 31: Looks like there are tagged releases on this one and we could pick the latest to use there instead of develop?  https://github.com/s0/git-publish-subdir-action/releases

jeffpaul (commented):
One comment that's mostly me reminding myself, but another potential update / question on a workflow file a bit further down the listing

.github/workflows/stable.yml
dkotter: I saw that but seems they haven't been putting out recent releases and they recommend pinning to a commit as well: https://github.com/s0/git-publish-subdir-action/issues/104.

My worry here is if we do go to the latest release (2.6.0 it looks like) because that release doesn't contain everything that is in `develop` and we've been relying on `develop` here, that may cause some breakage

jeffpaul: Ah ok, that makes sense then

jeffpaul (approved):

PR #875, opened 2025-03-21
Add support for new OpenAI Speech to Text models

dkotter:
### Description of the Change

OpenAI recently introduced two new [audio transcription models](https://platform.openai.com/docs/models#transcription): GPT-4o Transcribe and GPT-4o mini Transcribe. We currently use the Whisper model for our Audio Transcripts Generation Feature but according to OpenAI, the new GPT-4o mini Transcribe model is more accurate and similar in terms of cost.

We historically have not surfaced settings to choose which model to use and instead have various hooks in place allowing someone to change that if desired, while defaulting to models that both work well and are a reasonable cost. Initially I was going to do the same here but ultimately decided it's nice to provide users an easy way to choose the model they want, so there's now a setting to choose the model, which defaults to GPT-4o mini Transcribe but allows you to choose the older Whisper model or the new GPT-4o Transcribe model as well.

In addition, we originally had this Provider named Whisper, which isn't accurate, especially now with new model options. So I've renamed most things to be SpeechToText, matching our TextToSpeech Provider. I've left things like the Provider ID and hook names alone to preserve backwards compat.

### How to test the Change

1. Configure the Audio Transcripts Generation Feature, connecting it to OpenAI
2. Try transcribing an audio file and ensure it works
3. Try changing the model to the other options and ensure things still work

### Changelog Entry

> Added - Support for selecting new OpenAI Audio Transcription models, GPT-4o mini Transcribe and GPT-4o Transcribe
> Added - New filters around the Speech to Text model and API URL, allowing those to easily be changed
> Changed - Rename the OpenAI `Whisper` Provider to `SpeechToText`. If anyone extended the `Whisper` class, you'll need to update your code to account for this.

### Credits

Props @dkotter

### Checklist:


jeffpaul:
> We historically have not surfaced settings to choose which model to use and instead have various hooks in place allowing someone to change that if desired, while defaulting to models that both work well and are a reasonable cost. Initially I was going to do the same here but ultimately decided it's nice to provide users an easy way to choose the model they want, so there's now a setting to choose the model, which defaults to GPT-4o mini Transcribe but allows you to choose the older Whisper model or the new GPT-4o Transcribe model as well.

Sounds good to me, thanks for surfacing this context and decision.

iamdharmesh (approved):
Thanks for the PR, @dkotter! The code looks good and tests well. 

I just wanted to check if we should deprecate the `Whisper` class by extending the new `SpeechToText` class and removing it in a later release instead of completely removing (renaming) it as part of this PR. People generally don’t extend the provider classes and use hooks instead, so renaming should be fine. However, I just wanted to double-check before we merge the PR. Thanks!

dkotter:
> I just wanted to check if we should deprecate the `Whisper` class by extending the new `SpeechToText` class and removing it in a later release instead of completely removing (renaming) it as part of this PR. People generally don’t extend the provider classes and use hooks instead, so renaming should be fine. However, I just wanted to double-check before we merge the PR. Thanks!

That's a good point. I am of the opinion here that it's very unlikely anyone is extending the class so I'm not concerned with removing it without any deprecation (but making sure that's called out in the changelog). But open to differing opinions on that

PR #876, opened 2025-03-21
Add support for new OpenAI Text to Speech model

dkotter:
### Description of the Change

OpenAI recently introduced a new text to speech model, [GPT-4o mini TTS](https://platform.openai.com/docs/models/gpt-4o-mini-tts). We currently support the TTS-1 and TTS-1-HD models but seems this new model is better than those, both in terms of accuracy, speed and price.

This is the one Feature where we do already allow someone to choose the model in the settings so I've added GPT-4o mini TTS as an option there and have set that as the new default. I've also updated the available voice options as it seems those have changed (some new ones added, a few removed). And I've added new filters around the model and API URL, matching what we're doing in other Providers and making it easy for developers to modify those.

### How to test the Change

1. Configure the Text to Speech Feature, connecting to OpenAI
2. Try generating an audio file and ensure it works
3. Ensure this works with the new model and existing models

### Changelog Entry

> Added - Support for selecting the new GPT-4o mini TTS model in our Text to Speech Feature
> Added - New filters around the Text to Speech model and API URL, allowing those to easily be changed
> Changed - Set the GPT-4o mini TTS model as our new default in our Text to Speech Feature

### Credits

Props @dkotter

### Checklist:

### DIFF
--- a/includes/Classifai/Providers/OpenAI/TextToSpeech.php
+++ b/includes/Classifai/Providers/OpenAI/TextToSpeech.php
@@ -31,6 +31,49 @@ class TextToSpeech extends Provider {
 	}
 
 	/**
+	 * Get the API url.
+	 *
+	 * @return string
+	 */
+	public function get_api_url(): string {
+		/**
+		 * Filter the API URL.
+		 *
+		 * @since x.x.x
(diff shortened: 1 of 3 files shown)


iamdharmesh (approved):
🚀

PR #881, opened 2025-04-04
Recommended Content Block powered by AI embeddings

### LINKED ISSUE #884: Add Option for Backfilling Results When Threshold Isn't Met
Implement a feature to backfill results if fewer items meet the threshold than desired. This could either be the default behavior or introduced as a new setting. Ref: #881

dkotter:
### Description of the Change

We currently have a `Recommendation Service` section with a single Feature within that, `Recommended Content`. At the moment, the only Provider for this is Azure AI Personalizer though that has been deprecated by Microsoft and while we have kept the code in place (for now) we don't recommend anyone uses that.

We've looked into various alternatives(see #392) but haven't found a similar service to Azure AI Personalizer that we've liked. We have been using AI embeddings in other places to get content that is similar to each other (term cleanup, smart 404, classification) and I've had the idea for awhile to use that same concept to power content recommendations in a Recommended Content block. That's what this PR introduces.

Worth noting this is different from the recommendations Azure AI Personalizer provides, which are supposed to be more personalized content recommendations rather than just similar content recommendations.

This new Provider does the following:

- When configured, embeddings are generated for all existing Posts that don't already have embedding data
- In addition, anytime a Post is published (either initially or a published Post is saved) we generate embedding data for that item
- We've introduced a new Content Recommendation block that is a block variation of the core Query Loop block. I've left the settings here pretty simple, you can change the number of posts that show and the sort order
- This variation block uses a default template of the featured image, title, date and excerpt, though this can be modified after the block is inserted
- When this block renders, we modify the query that is run to first get the embedding data for the current item, run a similarity calculation against the 5000 most recent posts (this is to try and keep performance decent) and their embeddings and then return the posts that match our threshold (defaults to 75% but can be changed in the settings for the Feature). These results are cached to help with performance, with that cache expiring every hour. We then display however many items have been set (default is 3). Worth noting that if less items meet the threshold, we only display those. So if someone chooses to display 5 items but only 2 meet the threshold, we only display 2

While I see this as being a complete feature, there are additional improvements we could look to make, potentially in this PR but ideally in follow up PRs so we can ship this sooner:

- Right now we hardcode to Posts only. Ideal would be to allow someone to choose the post types they want, both at the settings level for the Feature but also in the block itself. Becomes a little tricky to sync those but is solvable
- We could look to backfill the results if less items meet the threshold than what we want. This could either be the default or introduce a new setting for this
- The Query Loop block has additional settings we can look to support, like filtering by taxonomy, filtering by author, etc
- As mentioned above, we default to a template that uses the featured image, title, date and excerpt. This can be changed after the block is added but we may want to add global settings to set the default here (or just change the default if we think there's a better standard template)
- Right now I only added support for OpenAI. Ideally we add our other embedding Providers (Azure OpenAI, Ollama)
- As I was working through our embedding code, there's quite a bit here that's duplicated across other embedding Providers. Potential here to do some code cleanup and either move shared code into the Feature or move shared code into some other shared class
- Probably worth discussion around removing Azure AI Personalizer now

### Screenshots

Settings for the Recommended Content Feature

Admin rendering of the Recommended Content block

### How to test the Change

1. Check out this PR and then go to `Tools > ClassifAI > Recommendation Service > Recommended Content`
2. Turn the Feature on and select and configure the OpenAI Embeddings Provider
3. After saving, you should see a message that embeddings are being generated. This may take some time depending on how much content you have but eventually this process should finish and a success message will show
4. Go to an existing Post and add the Recommended Content block
5. Ensure this is rendering some content, both in the editor and the front-end

### Changelog Entry

> Added - New Recommended Content block powered by the OpenAI Embeddings Provider

### Credits

Props @dkotter, @fabiankaegy 

### Checklist:


tests/cypress/integration/recommendation-service/recommended-content.test.js
```diff
@@ -0,0 +1,40 @@
...
+		cy.enableFeature();
+		cy.selectProvider( 'openai_embeddings' );
+		cy.get( '#openai_api_key' ).clear().type( 'password' );
+		cy.get( '#embedding-threshold' ).clear().type( '50' );
+
+		cy.allowFeatureToAdmin();
+		cy.saveFeatureSettings();
+	} );
+
+	it.skip( 'Can add the Recommended Content block in a post', () => {
```
dkotter, line 23: Note I'm skipping this test for now as it fails in the pipeline. Works fine for me locally but fails consistently when run here. In doing some debugging, seems the block can't be found but couldn't figure out why

jeffpaul:
Making a note here @dkotter @vikrampm1 to remind us once this is released that we'll likely want to adjust our microsite and ClassifAI overview deck with how this feature now works.  Probably also worth a quick scan through the dev docs site to see if there are any updates there that are worth including in this PR.

jeffpaul:
@vikrampm1 please open individual follow-up issues for the items that Darin called out:

> - Right now we hardcode to Posts only. Ideal would be to allow someone to choose the post types they want, both at the settings level for the Feature but also in the block itself. Becomes a little tricky to sync those but is solvable
> - We could look to backfill the results if less items meet the threshold than what we want. This could either be the default or introduce a new setting for this
> - The Query Loop block has additional settings we can look to support, like filtering by taxonomy, filtering by author, etc
> - As mentioned above, we default to a template that uses the featured image, title, date and excerpt. This can be changed after the block is added but we may want to add global settings to set the default here (or just change the default if we think there's a better standard template)
> - Right now I only added support for OpenAI. Ideally we add our other embedding Providers (Azure OpenAI, Ollama)
> - As I was working through our embedding code, there's quite a bit here that's duplicated across other embedding Providers. Potential here to do some code cleanup and either move shared code into the Feature or move shared code into some other shared class
> - Probably worth discussion around removing Azure AI Personalizer now

jeffpaul:
> As I was working through our embedding code, there's quite a bit here that's duplicated across other embedding Providers. Potential here to do some code cleanup and either move shared code into the Feature or move shared code into some other shared class

If we don't do that in this PR, then let's make that a fast-follow to ensure we're not dragging along some tech debt for too long (so @vikrampm1 please drop this follow-up issue in the next major milestone).

> Probably worth discussion around removing Azure AI Personalizer now

Similar to the above, if not in this PR let's do as a fast-follow (not just for tech debt, but also to avoid user confusion).  @vikrampm1 make this follow-up issue in the current milestone and tag whoever has an open source week coming up next week.

includes/Classifai/Features/RecommendedContent.php
```diff
@@ -47,10 +65,163 @@ public function get_enable_description(): string {
 	 */
 	public function get_feature_default_settings(): array {
 		return [
-			'provider' => PersonalizerProvider::ID,
+			'provider' => OpenAIEmbeddings::ID,
```
jeffpaul: One question that came to me for some reason in scanning this section of code, what if (and I recognize the probability here is likely an asymptote of 0) someone has a properly configured Personalizer account and is actually using the existing Recommended Content block?  I know that provider stops working (I think later this October?) but if we swap in OAI will any previously functioning Personalizer integrations continue to work?

jeffpaul:
_^^^Testing out Copilot code review^^^_

dkotter:
> If we don't do that in this PR, then let's make that a fast-follow to ensure we're not dragging along some tech debt for too long

Yeah, not planning on doing that here but can tackle that as a follow up. Also probably won't remove Azure AI Personalizer here but can do that in a follow up as well.

includes/Classifai/Features/RecommendedContent.php
dkotter: So this PR doesn't remove Azure AI Personalizer, just adds OpenAI as a new Provider and sets that as the new default. So no impact from this PR on existing users.

The plan was to open a follow up PR to remove Azure AI Personalizer all together though we will need to consider that, if someone is using the old block, how do we want to handle that? I guess I'm okay with just removing all that code, knowing that anywhere that block is used it will no longer work, as long as we are very explicit about that in the changelog so anyone using that can defer updating.

Another alternative would be to remove the Provider but leave all the code in place, so no one can use that going forward but it will still work for those that have set it up but that then leaves all that legacy code in place which would be nice to clean up.

But no real way to migrate from the old to the new, as the way these work is so different and require someone to configure OpenAI

Sidsector9:
@dkotter I'm not sure if I'm testing this incorrectly, but I'm unable to make this work.

This is the ZIP of posts I'm testing with. [classifai-posts.xml.zip](https://github.com/user-attachments/files/20437223/classifai-posts.xml.zip)

The Recommended Block is added to the post: `Senate Republicans vote to revoke California’s right to set its own tailpipe pollution rules`. This is the post which is unique and unrelated to other posts, but still the block renders posts which are unrelated.

I have confirmed with 10% and 100% threshold, and also confirmed that post embedding meta exists for all the posts.

fabiankaegy:
Just adding my twocents here. 

I am thinking of content recommendation in a query loop more like a sorting than a filtering. 

I always want to display some posts at the bottom of an article to keep visitors in my site. Ideally those are as relevant as possible. But even when we don't find anything relevant I still want there to be some posts.

dkotter:
> This is the post which is unique and unrelated to other posts, but still the block renders posts which are unrelated.

@Sidsector9 I tested this out and did find an issue, though I think unrelated to your results. I think what you're running into is by design (though we can change this) where if no related results are found we just run a normal query. This is to ensure we always show something instead of just having this section be blank.

This also matches @fabiankaegy's point:

> I always want to display some posts at the bottom of an article to keep visitors in my site. Ideally those are as relevant as possible. But even when we don't find anything relevant I still want there to be some posts.

I did find an issue though where if a post has this query block added (or any query block) the content from that is used when we generate embeddings. This can lead to unexpected matches because we have partial content from one or more posts that is used as embeddings for a different post, leading to high matches on those chunks. This has been resolved in [082f370](https://github.com/10up/classifai/pull/881/commits/082f3709e2eae6070e5e12743f46f5e61ece4455) by stripping any Query blocks out before we render content

dkotter:
> I always want to display some posts at the bottom of an article to keep visitors in my site. Ideally those are as relevant as possible. But even when we don't find anything relevant I still want there to be some posts.

As mentioned above, it does work this way right now, where it will run the default query if no relevant results are found. One thing it doesn't do (though I have a `TODO` statement in the code to consider this) is backfill results if we don't have enough relevant results. Meaning if someone wants to show 3 items but we only have 1 that matches our threshold, right now that is the only item that shows. In those situations we could run an extra query to get more results, ensuring we always have 3 items to show.

dkotter:
> One thing it doesn't do (though I have a TODO statement in the code to consider this) is backfill results if we don't have enough relevant results. Meaning if someone wants to show 3 items but we only have 1 that matches our threshold, right now that is the only item that shows. In those situations we could run an extra query to get more results, ensuring we always have 3 items to show

This has been added now, where we backfill the results using the default query if needed. Could look at making this a setting in the future but seems like good default behavior

includes/Classifai/Features/RecommendedContent.php
```diff
@@ -241,12 +241,36 @@ public function modify_block_query_vars( array $query_vars = [], int $post_id =
...
+					[
+						'posts_per_page' => (int) $query_vars['posts_per_page'] - $post_count,
+						'post__not_in'   => [ $post_id ],
+						'fields'         => 'ids',
+					]
+				)
+			);
+
+			// Add the backfilled posts to the post__in array.
+			$post__in   = array_merge( $post__in, $backfill_query->posts );
```
fabiankaegy, line 264: OMG why did I never think of doing it this way.... 🤦 I spent so much time hacking the internals of how queries are done to add backfilling support. This is so nice!

Sidsector9 (approved):
Tested, and works as per description and discussion 👍

PR #917, opened 2025-05-23
feat/886: Add Default Template global settings for Recommended Content Block

### LINKED ISSUE #886: Introduce Global Template Settings for Default Block Layout
Currently via #881 , the block defaults to using featured image, title, date, and excerpt. We may want to add global settings to set this default template or alter it to a better standard.

Sidsector9:
### Description of the Change

Closes #886 

### How to test the Change

1. Visit the page `/wp-admin/tools.php?page=classifai#/personalizer/feature_recommended_content`.
2. Select a **Default Template** and save the settings.
3. Follow the instructions from PR #881 to insert the Recommended Content Block.
4. Verify if the template set in the block is the same as the Default Template set in the settings.

### Screencast:

https://github.com/user-attachments/assets/29b2adfa-d016-498e-a04d-f031d572cf46

### Changelog Entry

> Added - Default Template global settings for Recommended Content Block

### Credits

Props @dkotter @Sidsector9 

### Checklist:

### DIFF
--- a/src/js/settings/components/feature-additional-settings/index.js
+++ b/src/js/settings/components/feature-additional-settings/index.js
@@ -3,8 +3,6 @@
  */
 import { Slot } from '@wordpress/components';
 import { PluginArea } from '@wordpress/plugins';
-import { registerPlugin } from '@wordpress/plugins';
-import { Fill } from '@wordpress/components';
 
 /**
  * Internal dependencies
--- a/src/js/settings/components/feature-additional-settings/index.js
+++ b/src/js/settings/components/feature-additional-settings/index.js
@@ -3,6 +3,8 @@
  */
 import { Slot } from '@wordpress/components';
 import { PluginArea } from '@wordpress/plugins';
+import { registerPlugin } from '@wordpress/plugins';
+import { Fill } from '@wordpress/components';
 
 /**
  * Internal dependencies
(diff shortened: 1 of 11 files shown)


src/js/settings/components/feature-additional-settings/index.js
```diff
@@ -3,6 +3,8 @@
  */
 import { Slot } from '@wordpress/components';
 import { PluginArea } from '@wordpress/plugins';
+import { registerPlugin } from '@wordpress/plugins';
+import { Fill } from '@wordpress/components';
```
dkotter: Seems neither of these is currently used?

dkotter (changes requested):
Thanks for adding this (and apologies on late review here). Left one comment but I think we need to be a bit more strict on the templates here. What I'm seeing now is pagination gets added, which isn't desired for a recommended content section. Is there a way to remove that piece? It also doesn't render in the way I'd expect, it renders stacked now instead of in a column layout.

src/js/settings/components/feature-additional-settings/index.js
Sidsector9: Fixed 👍

dkotter (approved):
Looks good to me, thanks for the work on this. I think we may want to register our own templates in the future, rather than relying on whatever default templates are registered, to provide a few more options but this looks good as a starting point

PR #939, opened 2025-06-18
Add a main Diagrams doc page and nest all diagrams under that

dkotter:
### Description of the Change

Noticed today that all of the new diagrams we added in #902 are showing as navigation items in our sidebar menu (see https://10up.github.io/classifai/). While this works it's not ideal, as these don't have proper titles added and they take more space in our sidebar.

This PR fixes that by adding a new Diagrams page and nesting all of these diagrams underneath that. The result will be a single link in the sidebar (Diagrams) that will take you to a page that lists out all of the individual diagrams:

Diagrams page and sidebar navigation

There are potential followups we could look at here if we want, though not sure how feasible these are:

1. The tutorials template automatically outputs a list of children pages and then the actual page title and content beneath that. Would be great to be able to modify that template to show the title and content above instead of below but I didn't find a quick way to accomplish this
2. The diagrams we have on these pages are mermaid charts and they don't render properly. In doing some quick searching, seems there's a somewhat hacky approach to fix this (basically loading in some custom JS on those pages) that if we can get it working, would make those diagram pages nicer

### How to test the Change

If desired, can build docs locally by running `npm run build:docs` and verify things look good.

Otherwise not much to review here

### Changelog Entry

> Developer - Add a main Diagrams doc page and nest all diagrams under that

### Credits

Props @dkotter

### Checklist:

### DIFF
--- a/hookdoc-conf.json
+++ b/hookdoc-conf.json
@@ -7,7 +7,7 @@
       "tutorials": "./hookdocs"
   },
   "source": {
-      "includePattern": ".+\\.(php|inc)?$"
+    "includePattern": ".+\\.(php|inc)?$"
   },
   "plugins": [
     "node_modules/wp-hookdoc/plugin",
--- a/hookdocs/meta.json
+++ b/hookdocs/meta.json
@@ -15,9 +15,41 @@
         "title": "Chrome built-in AI"
     },
     "data-flow": {
-        "title": "ClassifAI Data Flow"
+        "title": "Data Flow"
     },
     "feature-provider-support": {
         "title": "Providers supported per Feature"
+    },
+    "diagrams": {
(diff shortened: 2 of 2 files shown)


hookdocs/diagrams.md
dkotter, line 1: Purposely left this page empty as the built-in template will render a list of all children page at the top, so no need to render anything below that

jeffpaul (approved):

jeffpaul:
@dkotter I noticed in hindsight that there aren't any visuals added for this section, though I think you had created some for a previous ClassifAI client request.  Perhaps its worth ensuring those diagrams are generic enough that they could be added to the docs pages here where relevant to pair with the text descriptions you've added?

PR #949, opened 2025-06-30
Ensure we return API error when generating embeddings to avoid infinite loops

dkotter:
### Description of the Change

Was helping diagnose some issues on a client site and it appeared when generating embeddings for terms (when using OpenAI as the Provider for the Classification Feature) that we were getting stuck in an infinite loop with Action Scheduler.

In doing some testing locally, I think I figured out the issue. If we encounter an API error when generating an embedding for a term, we don't ever return that error and thus we continue to process that same term (or terms) over and over again with the same result. We previously fixed this for post embeddings but missed it for term embeddings, so this PR brings that same handling over.

In addition, we were missing both post and term error handling in our other embedding Providers so this PR fixes that as well.

### How to test the Change

Tricky to test this as it requires a situation where you have a valid API key (so you can configure the Provider properly) but the API key doesn't have access to make requests. Can also modify the code to always return an error for easier testing.

1. Enable the Classification Feature with OpenAI as the Provider
2. (Tricky part) Add an API key that is valid but doesn't have access to generate embeddings (I tested this with my personal account that has no credits associated with it)
3. May need to save twice to get embedding generation to start (this is another bug we need to fix). You should see a message saying embedding generation in progress
4. Validate within the Action Scheduler page that the process stops after processing all terms. Previously the same Action Scheduler event would be scheduled over and over

### Changelog Entry

> Fixed - Ensure if an API error occurs when generating embeddings, we don't try to keep processing that item and end up in an infinite loop

### Credits

Props @dkotter

### Checklist:

### DIFF
--- a/includes/Classifai/Providers/Azure/Embeddings.php
+++ b/includes/Classifai/Providers/Azure/Embeddings.php
@@ -366,11 +366,11 @@ class Embeddings extends OpenAI {
 	}
 
 	/**
-	 * Trigger embedding generation for content being saved.
+	 * Trigger embedding generation for a post.
 	 *
-	 * @param int  $post_id ID of post being saved.
+	 * @param int  $post_id ID of post.
 	 * @param bool $force Whether to force generation of embeddings even if they already exist. Default false.
-	 * @return array|WP_Error
+	 * @return array[]|WP_Error Array of embedding vectors on success, WP_Error on failure.
 	 */
 	public function generate_embeddings_for_post( int $post_id, bool $force = false ) {
 		// Don't run on autosaves.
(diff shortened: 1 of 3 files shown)


MiguelAxcar (dismissed):
@dkotter I tested it with a restricted OpenAI API key (`/v1/embeddings` set to `None`) from my personal account: authentication succeeds, but requests to the embeddings endpoint return a 403, which was useful to simulate the issue.

With `develop` code I was able to reproduce the bug: an invalid key caused a loop that kept generating pending actions tied to the `classifai_generate_term_embedding_job` hook. These flipped to complete as soon as an embeddings valid key was added on the tests I made.

After applying your changes, I confirmed the fix works: even with an invalid key, the actions now complete gracefully, and I saw no differences in behavior between valid and invalid keys in this context, as expected. 

One small nit: the docblock for `generate_embeddings_for_post` method might benefit from a tweak, maybe something like `@return array[]|WP_Error` and a short description along the lines, such as "Array of embedding vectors on success, WP_Error on failure" and adding details on the expected format/dimensions of the returned array/vectors could help too.

I also double checked how callers of `generate_embeddings_for_post` method were handling return values, to avoid fatal errors, and I found that all calls are robust, either checking for `is_wp_error` or propagating the error, so no immediate risk of fatal iteration over an error object.

Solid work from my perspective ✅

PR #958, opened 2025-07-15
Ensure we log errors appropriately in our image WP-CLI command

dkotter:
### Description of the Change

We have a number of WP-CLI commands that can be used to run functionality in ClassifAI. In recently testing our `image` command, noticed that we had code that would output any errors that were logged but we weren't actually logging errors if they occurred. This PR fixes that by ensure we save any errors so they can be output, making it easier to debug if something goes wrong.

### How to test the Change

1. Configure the Descriptive Text Generator Feature
2. Run the image WP-CLI command, ex `wp classifai image 4579`
3. Ensure this works properly
4. Disable the Feature and run the command again
5. Ensure an error message shows

### Changelog Entry

> Developer - Ensure error messages are properly logged when running the image WP-CLI command

### Credits

Props @dkotter

### Checklist:


faisal-alvi (approved):
@dkotter Thanks for the fix, LGTM.

Before:

image


After:

image

PR #962, opened 2025-07-18
Standalone Pre-Publish Classification Panel

### LINKED ISSUE #879: The "Classify Post" suggestion in the Pre-Publish panel is no longer visible.
### Describe the bug

While reviewing PR #860, I noticed that the suggestion for "Classify Post" is no longer visible in the Pre-Publish panel. (It was initially added in #609.) With the initial check didn’t find any useful information, but it seems this issue occurred due to a regression caused by recent feature-side JS improvements.

### Steps to Reproduce

1. Enable and configure the Classification feature.  
2. Create a new post and add a title and some content.  
3. Click on "Publish" to open the Pre-Publish panel.  
4. Notice that the "Classify Post" suggestion is missing. (Previously, it appeared as shown in the screenshot below.)

Image

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

faisal-alvi:
### Description of the Change

This PR fixes a regression where the "Classify Post" suggestion in the Pre-Publish panel was no longer visible after recent refactoring. The issue occurred because the `PrePubClassifyPost` component was embedded within the `ClassificationButton` component instead of being registered as a separate WordPress plugin.

**Solution:**
- Extracted the pre-publish panel into a separate plugin file (`pre-publish-panel.js`)
- Added proper webpack entry and PHP enqueue for the new component
- Removed the embedded usage from the classification button
- Reused existing code without rewriting to minimize risk

Closes #879

### How to test the Change

1. Enable and configure the Classification feature in WordPress admin
2. Create a new post and add a title and some content
3. Click on "Publish" to open the Pre-Publish panel
4. Verify the "Classify Post" suggestion appears in the Pre-Publish panel
5. Test the classification functionality by clicking "Suggest terms & tags"
6. Verify the classification button in the editor toolbar still works correctly
7. Check for any console errors in the browser developer tools

The "Classify Post" suggestion should be visible in the Pre-Publish panel, and both the pre-publish panel and classification button should function independently without conflicts.

### Changelog Entry

> Fixed - Restored "Classify Post" suggestion in Pre-Publish panel after regression

### Credits

Props @iamdharmesh @faisal-alvi 

### Checklist:

### DIFF
--- a/includes/Classifai/Features/Classification.php
+++ b/includes/Classifai/Features/Classification.php
@@ -348,6 +348,14 @@ class Classification extends Feature {
 			true
 		);
 
+		wp_enqueue_script(
+			'classifai-plugin-classification-pre-publish-js',
+			CLASSIFAI_PLUGIN_URL . 'dist/classifai-plugin-classification-pre-publish.js',
+			array_merge( get_asset_info( 'classifai-plugin-classification-pre-publish', 'dependencies' ), array( 'lodash' ), array( Feature::PLUGIN_AREA_SCRIPT ) ),
+			get_asset_info( 'classifai-plugin-classification-pre-publish', 'version' ),
+			true
+		);
+
 		wp_add_inline_script(
 			'classifai-plugin-classification-js',
 			sprintf(
--- a/includes/Classifai/Providers/Azure/Embeddings.php
+++ b/includes/Classifai/Providers/Azure/Embeddings.php
@@ -695,6 +695,11 @@ class Embeddings extends OpenAI {
 				$term_embedding = get_term_meta( $term_id, 'classifai_azure_openai_embeddings', true );
 
 				if ( ! empty( $term_embedding ) ) {
+					// Convert the term embedding to an array if it's not already.
+					if ( ! is_array( $term_embedding[0] ) ) {
+						$term_embedding = [ $term_embedding ];
+					}
+
 					// Loop through the chunks and run a similarity calculation on each.
 					foreach ( $term_embedding as $chunk ) {
 						$similarity = $calculations->cosine_similarity( $embedding, $chunk );
--- a/includes/Classifai/Providers/Azure/Embeddings.php
+++ b/includes/Classifai/Providers/Azure/Embeddings.php
@@ -695,11 +695,6 @@ class Embeddings extends OpenAI {
 				$term_embedding = get_term_meta( $term_id, 'classifai_azure_openai_embeddings', true );
 
 				if ( ! empty( $term_embedding ) ) {
-					// Convert the term embedding to an array if it's not already.
-					if ( ! is_array( $term_embedding[0] ) ) {
-						$term_embedding = [ $term_embedding ];
-					}
-
 					// Loop through the chunks and run a similarity calculation on each.
 					foreach ( $term_embedding as $chunk ) {
 						$similarity = $calculations->cosine_similarity( $embedding, $chunk );
--- a/includes/Classifai/Providers/Localhost/OllamaEmbeddings.php
+++ b/includes/Classifai/Providers/Localhost/OllamaEmbeddings.php
@@ -771,11 +771,6 @@ class OllamaEmbeddings extends Ollama {
 				$term_embedding = get_term_meta( $term_id, 'classifai_ollama_embeddings', true );
 
 				if ( ! empty( $term_embedding ) ) {
-					// Convert the term embedding to an array if it's not already.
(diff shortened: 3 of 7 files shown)


faisal-alvi:
Ready for review.

iamdharmesh (commented):
Thanks for the PR @faisal-alvi. Code looks good. However, while testing, I noticed that the suggested terms are not being assigned to the post when it is requested from the pre-publish panel. Another issue I observed is that the "Classify Post" suggestion always appears, even when the post has already been classified and has terms saved or suggested. It should only appear if classification hasn’t been done yet.

Could you please check it once. Thanks!

faisal-alvi:
@dharmesh Thanks for the feedback! I think I missed the Save button earlier. I’ve now added it, so you should be able to click Save, and the suggested terms will be correctly assigned to the post when requested from the pre-publish panel.

image

Regarding the other issue you mentioned, the “Classify Post” suggestion always appears regardless of whether the user has already added tags or categories. I believe this is intended behavior, but I think we can add some other enhancements around this area for sure. But for now, this PR focuses on bringing the ClassifAI Post tab back to the Pre-Publish panel. We can explore enhancements in future tickets.

---

@dkotter While working on this, I noticed that AI suggestions for tags and categories were intermittently failing. After debugging, I found that sometimes the $embeddings variable was populated with a single array, and other times with multiple nested arrays. I’ve added a potential fix to normalize this behavior. Although this fix isn’t originally part of the scope of this PR, it felt too minor to raise a separate PR.

The inconsistent array formation seems to stem from this part of the code:
https://github.com/10up/classifai/blob/97412809ae1ea44e71fa0855d9286de49713d92d/includes/Classifai/Features/Smart404EPIntegration.php#L214

I avoided changing this directly to prevent potential issues for existing users who may have legacy data structures in their DBs. Since it looks like you’ve worked on this area previously, I’d really appreciate your feedback on the change, especially [here](https://github.com/10up/classifai/pull/962/files/812ceb069bcaae988d3534db8bf3ba394dfb5b14..0d2564e626bf837ef36c6dc83000648ee4b54c65).

dkotter:
> While working on this, I noticed that AI suggestions for tags and categories were intermittently failing. After debugging, I found that sometimes the $embeddings variable was populated with a single array, and other times with multiple nested arrays. I’ve added a potential fix to normalize this behavior. Although this fix isn’t originally part of the scope of this PR, it felt too minor to raise a separate PR.
> 
> The inconsistent array formation seems to stem from this part of the code:
> 
> [classifai/includes/Classifai/Features/Smart404EPIntegration.php](https://github.com/10up/classifai/blob/97412809ae1ea44e71fa0855d9286de49713d92d/includes/Classifai/Features/Smart404EPIntegration.php#L214)
> 
> Line 214 in [9741280](https://github.com/10up/classifai/commit/97412809ae1ea44e71fa0855d9286de49713d92d)
> 
>  // Otherwise let's get all embeddings in a single request. 
> I avoided changing this directly to prevent potential issues for existing users who may have legacy data structures in their DBs. Since it looks like you’ve worked on this area previously, I’d really appreciate your feedback on the change, especially [here](https://github.com/10up/classifai/pull/962/files/812ceb069bcaae988d3534db8bf3ba394dfb5b14..0d2564e626bf837ef36c6dc83000648ee4b54c65).

So for terms, I believe we'll only ever have a single array of embeddings. But for posts, unless the post is really short, we will end up with multiple embeddings that are each stored in their own array. That should already be normalized prior to saving that data to post meta, whether we have a single array of embeddings or multiple arrays of embeddings. If that's not working properly, that is a bug we should fix when that embedding data is saved and not try and normalize that when we use the data. I haven't had time to test this myself but if you're able to reproduce this, I'd suggest we fix this during the save process and not add a normalization fix.

Depending on what is required to fix that, we can consider what to do for backwards compatibility but I'd prefer to address this in the right place if we can.

iamdharmesh:
@faisal-alvi I am able to see Save button now and it work as expected. So, PR is all good for me to be merged once discussion for the embeddings changes got concluded.

iamdharmesh:
@faisal-alvi @dkotter as pre-publish penal changes looks good, maybe we can handle the embeddings update things in separate PR to unblock the original work on this PR?

dkotter:
> @faisal-alvi @dkotter as pre-publish penal changes looks good, maybe we can handle the embeddings update things in separate PR to unblock the original work on this PR?

I agree, we can document this in an Issue and open a new PR to address this, and remove any code that was added to this PR to fix that

dkotter (approved):

PR #964, opened 2025-07-24
Address minor issues flagged in a SAST scan

dkotter:
### Description of the Change

We recently ran ClassifAI through a handful of security audits. One of those was a SAST scan and while no major issues were flagged, quite a few smaller issues were highlighted. This PR fixes quite a few of those, though there are still some we may want to look at fixing in follow-up PRs.

I know there's a lot of file changes here but most of these changes fall in the following buckets:

- Simplifying unnecessarily nested `if` statements (error message `Merge this if statement with the enclosing one`)
- Directly returning values instead of assigning those to a variable and then immediately returning that variable (error message `Immediately return this expression instead of assigning it to the temporary variable`)
- Removing variables that aren't used (error message `Remove this unused "$example" local variable`)

### How to test the Change

Ensure all tests pass

If desired, could use some smoke testing across the entire plugin to ensure all functionality works as expected

### Changelog Entry

> Fixed - Address minor issues reported by a SAST scan

### Credits

Props @dkotter

### Checklist:


dkotter:
Note I'm holding this for now. While I think some of these fixes are good to have, others I'm not entirely sure I agree with. There also may be other scans coming so going to wait on those before proceeding

PR #965, opened 2025-07-27
Fix | Conflicts between AWS SDK in this plugin and others

### LINKED ISSUE #822: Conflicts between AWS SDK in this plugin and others
### Describe your question

Hey y'all,

first of all thanks for a great plugin! :)

[Since Amazon Polly was added as a text-to-speech](https://github.com/10up/classifai/pull/734/files#diff-d2ab9925cad7eac58e0ff4cc0d251a937ecf49e4b6bf57f8b95aab76648a9d34) provider this plugin can conflict with other plugins (or other parts of wp-apps) that also uses the AWS SDK. In general this wouldn't be a big problem as long as both places that loaded the SDK where using a somewhat similar version. But since Classifai removes all services from the AWS SDK except for Polly this can lead to issues such as missing methods in the SDK.

I realize that loading all of the AWS SDK will make the plugin-zip really large, so I was wondering if maybe you could use something like https://github.com/coenjacobs/mozart to namespace the dependencies brought in via composer? Or if you've got any better ideas to solve the conflict.

### Code of Conduct

sksaju:
### Description of the Change

This PR implements Strauss dependency prefixing to namespace the AWS SDK, preventing conflicts with other WordPress plugins that may use the same AWS SDK package. The implementation creates a completely isolated version of the AWS SDK under the ClassifaiVendor\ namespace.

Closes https://github.com/10up/classifai/issues/822

### How to test the Change

First, let's get everything installed and built:

```bash
# Navigate to the plugin directory and install dependencies
composer install
npm install
npm run build
```

#### Check That Strauss Did Its Job

Strauss should have created a `vendor-prefixed` directory with our namespaced AWS SDK. Let's verify:

```bash
# See if the prefixed directory exists
ls -la vendor-prefixed/aws/

# Check that classes are properly namespaced
grep -r "ClassifaiVendor" vendor-prefixed/aws/
```

You should see something like this:
```
vendor-prefixed/aws/
├── aws-sdk-php/
│   ├── src/
│   │   ├── Aws/
│   │   │   ├── Polly/
│   │   │   └── ... (other AWS services)
│   └── autoload.php
```

#### Test the AWS Polly Feature

Now let's make sure our text-to-speech functionality still works:

1. Head to your WordPress admin panel
2. Navigate to **ClassifAI** → **Text to Speech**
3. Enter your AWS credentials:
   - Access Key ID
   - Secret Access Key  
   - AWS Region (like `us-east-1`)
4. Try generating speech from a post - you should get an audio file back

#### The Real Test - Conflict Prevention

This is where we see if Strauss actually prevents conflicts:

1. Install another plugin that uses AWS SDK (WooCommerce with AWS extensions works well)
2. Activate both plugins at the same time
3. Check for any PHP errors in your error logs
4. Test both plugins to make sure they still work properly

If everything's working, you shouldn't see any "class already exists" or similar errors!

#### Test the Build Process

Let's make sure our build process works correctly:

```bash
# Run the archive command
npm run archive

# Check what's in the release directory
ls -la release/
```

### Changelog Entry

> Added - Implement Strauss dependency prefixing for AWS SDK to prevent conflicts with other plugins
### Credits

### Checklist:

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2F10up%2Fclassifai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-965-a74da3f20e33fc3059eb58ed3f76428372858417.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


classifai.php
```diff
@@ -82,6 +82,13 @@ function classifai_define( $name, $value ) {
  * @return bool True or false if autoloading was successful.
  */
 function classifai_autoload() {
+
+	// Load the prefixed vendor autoloader (contains AWS SDK and all other dependencies)
```
dkotter, line 92: So currently this function will return `true` if the autoload file is loaded properly. Now that we are introducing another autoload file, I think we'll need additional changes here to ensure both are loaded properly and show an error message if not.

Something like this (untested):
```php
function classifai_autoload() {
	$loaded = false;

	// Load the prefixed vendor autoloader (contains AWS SDK and all other dependencies)
	if ( file_exists( CLASSIFAI_PLUGIN_DIR . '/vendor-prefixed/autoload.php' ) ) {
		require_once CLASSIFAI_PLUGIN_DIR . '/vendor-prefixed/autoload.php';
		$loaded = true;
	}

	// Load the main vendor autoloader (contains plugin classes and other dependencies)
	if ( $loaded && file_exists( CLASSIFAI_PLUGIN_DIR . '/vendor/autoload.php' ) ) {
		require_once CLASSIFAI_PLUGIN_DIR . '/vendor/autoload.php';
	} else {
		$loaded = false;
	}

	if ( ! $loaded ) {
		error_log( // phpcs:ignore WordPress.PHP.DevelopmentFunctions.error_log_error_log
			sprintf( 'Warning: Composer not setup in %s', CLASSIFAI_PLUGIN_DIR )
		);

		return false;
	}

	return true;
}
```

.github/workflows/build-release-zip.yml
```diff
@@ -48,6 +48,9 @@ jobs:
     - name: Install composer dependencies
       run: composer install --no-dev -o
 
+    - name: Run Strauss to namespace dependencies
+      run: php -d memory_limit=2048M -d max_execution_time=300 bin/strauss.phar
```
dkotter: Is this needed? Looking at the changes in `composer.json`, we run this same command on the `post-install-cmd` hook which I think means this will automatically run after the step above

dkotter (changes requested):

dkotter (changes requested):
One other thing that needs fixed here. If running `composer install` locally, you'll need to ensure your environment is running PHP 7.4 (which is our minimum for this plugin). If you're running PHP 8+, composer will set PHP 8.1 as our minimum which breaks things (you can see this in some of our failing GitHub Actions on this PR). So you'll want to install PHP 7.4 locally, run `composer install` and then commit the `composer.lock` file

jeffpaul:
@dkotter @peterwilsoncc we'll likely need to pick this one up to finish it off, to give us a bit more time I'm moving this out a release

dkotter (approved):

PR #966, opened 2025-08-01
Replace hookdoc with WP Hooks Documentor

iamdharmesh:
### Description of the Change
PR replace the existing hookdoc package with the WP Hooks Documentor package which is used for generate developer docs. The new package is uses the docusaurus and provide good flexibility on extending docs site.

image
  

### How to test the Change
1. Run `npm i && npm run build:docs`
2. Run `cd ./docs && npm run serve`
3. Visit http://localhost:3000 to check generated doc site.

For GH pages deployment testing I have try this on forked PR. You can check sample site here: https://iamdharmesh.github.io/classifai/ 

### Credits
Props @jeffpaul @dkotter @iamdharmesh 

### Checklist:


package.json
```diff
@@ -44,12 +45,10 @@
     "cypress-mochawesome-reporter": "^4.0.0",
     "cypress-plugin-tab": "^1.0.5",
     "husky": "^9.1.7",
-    "jsdoc": "^4.0.4",
     "lint-staged": "^16.1.2",
     "node-wp-i18n": "^1.2.8",
     "svg-react-loader": "^0.4.6",
-    "taffydb": "2.7.3",
-    "wp-hookdoc": "^0.2.0"
+    "wp-hooks-documentor": "github:10up/wp-hooks-documentor#build"
```
iamdharmesh, line 51: Using github branch to install, until the package get published on npm. This is not blocker for merging this PR and we can always update this later once package published.

.github/workflows/plugin-check.yml
```diff
@@ -39,3 +39,4 @@ jobs:
           exclude-directories: 'assets,dist,vendor'
           ignore-codes: 'PluginCheck.CodeAnalysis.DiscouragedFunctions.load_plugin_textdomainFound' # Plugin isn't on .org so we load the textdomain manually.
           ignore-warnings: true
+          ignore-codes: 'PluginCheck.CodeAnalysis.DiscouragedFunctions.load_plugin_textdomainFound'
```
dkotter: This was already added a couple lines above so this can be removed

wp-hooks-docs/docs/01.get-started/05.CHANGELOG.md
dkotter, line 1: So it looks like we have a pre-build command that copies the changelog to this file. Do we need content in this file then or can this file just be empty?

wp-hooks-docs/package.json
dkotter, line 1: Do we need this file?

.github/workflows/stable.yml
```diff
@@ -40,10 +45,12 @@ jobs:
...
+      run: |
+        rm -rf ./release
+        npm run build:docs
 
     - name: Deploy docs update
       uses: peaceiris/actions-gh-pages@4f9cc6602d3f66b9c108549d475ec49e8ef4d45e # v4.0.0
       with:
         github_token: ${{ secrets.GITHUB_TOKEN }}
-        publish_dir: './docs'
+        publish_dir: './docs/build'
```
dkotter: So right now our documentation site is at https://10up.github.io/classifai/. Will that URL be different after this or does it stay the same?

.github/workflows/stable.yml
```diff
@@ -14,6 +14,11 @@ jobs:
     - name: Checkout
       uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
 
+    - name: Setup proper PHP version
+      uses: shivammathur/setup-php@9e72090525849c5e82e596468b86eb55e9cc5401 # v2.32.0
+      with:
+        php-version: 8.3
```
dkotter: So I know if I run `composer install` locally and I'm running PHP 8+, composer will pull in dependencies that rely on PHP 8+, even though our minimum is PHP 7.4 (this has caught me a few times). I don't know if this will cause the same problems, I would assume not since we currently aren't setting a specific PHP version here so I'm assuming it defaults to 8+, but worth considering so we don't end up pushing out a final release that requires PHP 8+ unexpectedly

.github/workflows/plugin-check.yml
iamdharmesh: Yeah, I had added it long ago. So, might be due to result of updating branch with latest develop. fixed.

wp-hooks-docs/docs/01.get-started/05.CHANGELOG.md
iamdharmesh: We don't need that file. I had kept it initially, but realised that it would be duplicate. So, added that pre-build command but missed that file due to it wasn't added in gitignore file.

wp-hooks-docs/package.json
iamdharmesh: We are adding additional `@docusaurus/theme-mermaid` dependency on top of default deps for render flow diagram. So, we have to keep this file.

.github/workflows/stable.yml
iamdharmesh: It will remain same.

.github/workflows/stable.yml
iamdharmesh: Agree. I have seperated build docs workflow from this now. So, this is no longer concern now.

dkotter (dismissed):
This looks really good to me, nice work here.

One thing we should look to capture is the steps involved in building and running the documentation locally, just to make it easier for anyone looking to contribute to that in the future.

Otherwise I think once these changes are live, we can probably trim down our README quite a bit and just link to the relevant sections within the documentation site. That can happen in a follow up PR though

iamdharmesh:
> One thing we should look to capture is the steps involved in building and running the documentation locally, just to make it easier for anyone looking to contribute to that in the future.

Done.

dkotter (approved):

PR #969, opened 2025-08-08
Enhancement/921 Add Custom and ACF Field targeting for Excerpt Generation

### LINKED ISSUE #921: Allow Excerpt Generation to target alternate fields (e.g. custom excerpt, subtitle, or dek fields)

Expand the Excerpt Generation feature to allow generated excerpts to be stored in custom or alternate fields beyond the default WordPress excerpt.

This could include:
- Selecting a custom field (e.g. `editorial_subtitle`, `custom_excerpt`, `dek`)
- Option to define a fallback priority (e.g. try custom field first, then default excerpt)
- Support for ACF or post meta field targeting
- UI option in settings to specify the target field for excerpt generation

Alternatively, replicate excerpt generation logic for use with subtitle or dek-specific fields to better support editorial workflows using structured content blocks or custom meta.

### Designs

Some editorial workflows, particularly in enterprise publishing environments, do not use the native WordPress excerpt field. Instead, they store excerpts, subtitles, or deks in custom fields.  This enhancement would allow teams using tools like ACF or custom meta integrations to benefit from AI-powered excerpting in their preferred field layout.

- Could use the same logic and UI patterns as the Featured Image generation field selection
- Should integrate with WordPress `get_post_meta()` and `update_post_meta()` or `acf` APIs where appropriate
- A filter could allow further customization for custom implementations
- Consider frontend fallback display logic (e.g., if the target field is empty but a default excerpt exists)

### Code of Conduct

faisal-alvi:
### Description of the Change

This PR implements **custom field targeting** for the ClassifAI excerpt generation feature, allowing users to save generated excerpts to custom fields beyond the default WordPress excerpt field. This enhancement supports editorial workflows that use structured content blocks, custom meta fields, or ACF fields for storing excerpts, etc.

  - Added support for targeting custom meta fields (e.g., `editorial_subtitle`, `custom_excerpt`, `dek`)
  - Integrated ACF field support for seamless ACF integration
  - Maintained default excerpt field as primary option
  - Added fallback mechanisms for field detection and value retrieval
  - New "Target field" setting in excerpt generation configuration
  - Dropdown selection for field type (Default excerpt, Custom meta, ACF field)
  - Dynamic field selection based on chosen field type
  - ACF field integration with automatic field detection
  - **Classic Editor**: Full custom field support with real-time updates and field detection
  - **Block Editor**: Integrated custom field display and synchronization
  - Context-aware messaging for different field types
  - Multiple detection methods for ACF fields (data attributes, field keys, DOM structure)
  - Flexible meta field detection and updating

---

### New Setting Field:

image

### Block Editor (Custom Field):

image

### Block Editor (ACF):

image

### Classic Editor (ACF):

image

Closes #921

### How to test the Change

1. **Custom Meta Field Testing:**
   - Create custom meta fields (e.g., `editorial_subtitle`, `custom_excerpt`)
   - Configure excerpt generation to target custom meta fields
   - Generate excerpts and verify they save to the correct fields

2. **ACF Field Testing:**
   - Install/activate ACF plugin
   - Create ACF fields (text/textarea)
   - Configure excerpt generation to target ACF fields
   - Test excerpt generation and field updates

3. **Editor Testing:**
   - Test in both classic and block editors
   - Verify real-time field updates
   - Confirm proper field detection and synchronization

4. **Settings Testing:**
   - Navigate to ClassifAI settings > Language Processing > Excerpt Generation
   - Verify "Target field" dropdown with all options
   - Test field type switching and configuration saving

### Credits

Props @jeffpaul @faisal-alvi @iamdharmesh 

### Changelog Entry

> Added - Add custom field targeting for Excerpt Generation

### DIFF
--- a/includes/Classifai/Features/ExcerptGeneration.php
+++ b/includes/Classifai/Features/ExcerptGeneration.php
@@ -336,7 +336,7 @@ class ExcerptGeneration extends Feature {
 	 * @return string
 	 */
 	public function get_enable_description(): string {
-		return esc_html__( 'A button will be added to the excerpt panel that can be used to generate an excerpt.', 'classifai' );
+		return esc_html__( 'A button will be added to the excerpt panel that can be used to generate an excerpt. You can configure where the generated excerpt is saved (default excerpt field, custom meta fields, or ACF fields).', 'classifai' );
 	}
 
 	/**
--- a/includes/Classifai/Features/ExcerptGeneration.php
+++ b/includes/Classifai/Features/ExcerptGeneration.php
@@ -246,7 +246,9 @@ class ExcerptGeneration extends Feature {
 
 			// If generation was successful and we have a post ID, save to target field.
 			if ( ! is_wp_error( $result ) && $post_id ) {
-				$save_result = $this->save_excerpt_to_target_field( $result, $post_id );
+				$excerpt     = is_array( $result ) ? $result['content'] : $result;
+				$excerpt     = is_string( $excerpt ) ? $excerpt : '';
+				$save_result = $this->save_excerpt_to_target_field( $excerpt, $post_id );
 				if ( is_wp_error( $save_result ) ) {
 					return rest_ensure_response( $save_result );
 				}
--- a/includes/Classifai/Features/ExcerptGeneration.php
+++ b/includes/Classifai/Features/ExcerptGeneration.php
@@ -562,6 +562,15 @@ class ExcerptGeneration extends Feature {
 
 		$new_settings['length'] = absint( $new_settings['length'] ?? $settings['length'] );
 
+		// Sanitize target field settings.
+		$new_settings['target_field_type'] = sanitize_text_field( $new_settings['target_field_type'] ?? 'post_excerpt' );
+		
+		if ( 'custom_meta' === $new_settings['target_field_type'] ) {
+			$new_settings['target_custom_field'] = sanitize_text_field( $new_settings['target_custom_field'] ?? '' );
+		} elseif ( 'acf_field' === $new_settings['target_field_type'] ) {
+			$new_settings['target_acf_field'] = sanitize_text_field( $new_settings['target_acf_field'] ?? '' );
+		}
+
 		foreach ( $post_types as $post_type ) {
 			if ( ! post_type_supports( $post_type->name, 'excerpt' ) ) {
 				continue;
--- a/includes/Classifai/Features/ExcerptGeneration.php
+++ b/includes/Classifai/Features/ExcerptGeneration.php
@@ -718,4 +718,46 @@ class ExcerptGeneration extends Feature {
 				return apply_filters( 'classifai_excerpt_save_to_custom_target', new WP_Error( 'invalid_field_type', __( 'Invalid target field type.', 'classifai' ) ), $excerpt, $post_id, $field_type );
 		}
 	}
+
+	/**
+	 * Get the current value from the target field.
+	 *
+	 * @param int $post_id The post ID.
+	 * @return string The current value.
+	 */
+	public function get_current_target_field_value( $post_id ) {
+		$settings = $this->get_settings();
+		$field_type = $settings['target_field_type'] ?? 'post_excerpt';
+
+		switch ( $field_type ) {
+			case 'post_excerpt':
+				return get_the_excerpt( $post_id );
+
+			case 'custom_meta':
+				$meta_key = $settings['target_custom_field'] ?? '';
+				return get_post_meta( $post_id, $meta_key, true );
+
+			case 'acf_field':
+				$acf_field_key = $settings['target_acf_field'] ?? '';
+				if ( function_exists( 'get_field' ) ) {
+					return get_field( $acf_field_key, $post_id );
+				}
+				return '';
+
+			default:
+				/**
+				 * Filter for custom target field types when getting current value.
+				 *
+				 * @since 3.x.x
+				 * @hook classifai_excerpt_get_custom_target_value
+				 *
+				 * @param {string} $value      The current value.
+				 * @param {int}    $post_id    The post ID.
+				 * @param {string} $field_type The target field type.
+				 *
+				 * @return {string} The current value.
+				 */
+				return apply_filters( 'classifai_excerpt_get_custom_target_value', '', $post_id, $field_type );
+		}
+	}
 }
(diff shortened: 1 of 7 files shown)


src/js/features/excerpt-generation/classic/index.js
```diff
@@ -87,10 +94,322 @@ const classifaiExcerptData = window.classifaiGenerateExcerpt || {};
...
+											'Edit the custom field directly in the post editor or custom fields panel',
+											'classifai'
+									  ) +
+									  '</li>' ) +
+								'<li>' +
+								__(
+									'Change the target field in',
+									'classifai'
+								) +
+								'<a href="' +
```
iamdharmesh, line 230:
```suggestion
								' <a href="' +
```

Add space before settings link.

Image

includes/Classifai/Features/ExcerptGeneration.php
```diff
@@ -365,6 +450,108 @@ public function add_custom_settings_fields() {
...
+			[ $this, 'render_target_field_settings' ],
+			$this->get_option_name(),
+			$this->get_option_name() . '_section',
+			[
+				'label_for'     => 'target_field',
+				'default_value' => $settings['target_field_type'] ?? 'post_excerpt',
+				'description'   => __( 'Choose where to save the generated excerpt. You can target the default excerpt field, custom meta fields, or ACF fields.', 'classifai' ),
+			]
+		);
+	}
```
iamdharmesh, line 467: We are not supporting legacy settings panel now. So, this is no longer needed now.

includes/Classifai/Features/ExcerptGeneration.php
```diff
@@ -386,6 +573,9 @@ public function get_feature_default_settings(): array {
 			],
 			'length'                  => absint( apply_filters( 'excerpt_length', 55 ) ),
 			'provider'                => ChatGPT::ID,
+			'target_field_type'       => 'post_excerpt',
+			'target_custom_field'     => '',
+			'target_acf_field'        => '',
```
iamdharmesh, line 578: Nitpick: As we are already storing `target_field_type`, can we merge this two into single as a `target_field`?

includes/Classifai/Features/ExcerptGeneration.php
```diff
@@ -484,6 +683,202 @@ public function migrate_settings() {
...
+				return apply_filters( 'classifai_excerpt_get_custom_target_value', '', $post_id, $field_type );
+		}
+	}
+
+	/**
+	 * Get available ACF fields for the target field selector.
+	 *
+	 * @return array Array of ACF fields.
+	 */
+	public function get_available_acf_fields() {
```
iamdharmesh, line 812: Noticed similar `get_acf_fields()` function in Admin/Settings.php. Maybe better to move this to helper functions to prevent code duplication here?

iamdharmesh (changes requested):
Thanks for the working on this and raising the PR @faisal-alvi. Overall it looks good. However, I have added few comments. 

Additionally, I noticed that we are marking the default excerpt field as read-only/disabled when a custom meta field or ACF field is selected as the target field. I think we should keep the excerpt field as it is. We might instead show a notice near the excerpt generation button, indicating that the generated content will be saved into the selected custom field/ACF field. Otherwise, I believe we should leave the excerpt field unchanged, as there is no direct connection between them. I just wanted to know if there is a specific reasoning behind this. (cc: @dkotter any thoughts on this?)

Image

Thanks!

dkotter:
>  might instead show a notice near the excerpt generation button, indicating that the generated content will be saved into the selected custom field/ACF field. Otherwise, I believe we should leave the excerpt field unchanged, as there is no direct connection between them

Yeah, I agree feels a little weird to have that field disabled.

I also think we may need further discussion on how exactly this should work. For instance, feels like we may want to allow multiple selections here, so someone can store this in the excerpt field and a custom field (similar to our Descriptive Text Generator where you can choose one or multiple options for storage).

In addition, if I choose ACF or custom meta, I don't love the idea of having the `Generate Excerpt` button show in the Excerpt panel, as you're not storing the excerpt there anymore. There's probably a better location for that button in those scenarios, though that would require some more discussion.

And finally, if this Feature is on, we remove the core WordPress excerpt panel and replace it with our own, which allows us to add that generate button. Since that was introduced, WordPress has changed how the excerpt panel works and it's been on my list to bring those changes over to ClassifAI. Doesn't have to be done in this PR but might be a good opportunity to align our approach with core (or even investigate if there's a better option now to add that button).

includes/Classifai/Features/ExcerptGeneration.php
faisal-alvi: The reason for keeping these separate is to ensure that if the admin switches back to the previous type, they will still get the same field they were originally using.

faisal-alvi:
@iamdharmesh Thanks for the feedback!

> I noticed that we are marking the default excerpt field as read-only/disabled when a custom meta field or ACF field is selected as the target field.

1. Just to clarify, we’re not disabling the core excerpt field; we’re hiding it.
Instead, we show a read-only field that reflects the targeted Custom/ACF field value, along with a note for users explaining what to do if they want to edit it.

Reference: https://github.com/10up/classifai/blob/0397277b699369632a0e9593faa06ccb26ec58a0/src/js/features/excerpt-generation/classic/index.js#L104-L111

image

---

@dkotter Thanks for adding details. A few follow-ups from my side:

2. Multiple Selections

> feels like we may want to allow multiple selections here, so someone can store this in the excerpt field and a custom field 

My assumption is that the AI-generated excerpt would be inserted into both the custom/ACF field and the Core field. If that’s the case, should we consider adding:

- A new sub-setting when a custom/ACF target is chosen (checkbox: “Also insert into core excerpt”), or
- New dropdown options in the existing “Target field” selector, like “Both: Custom & Default” and “Both: ACF & Default”.

3. Location of the Generate Excerpt Button

> if I choose ACF or custom meta, I don’t love the idea of having the Generate Excerpt button show in the Excerpt panel

Personally, I feel the current location is still reasonable for now since we provide a clarifying note (screenshot above), but in the future, we could revisit placement. However, do you have a preferred spot in mind? Happy to make those changes. @iamdharmesh would also love your input here or on any other points.

4. Bringing Core Changes to ClassifAI

> WordPress has changed how the excerpt panel works, and it’s been on my list to bring those changes over to ClassifAI

Could you elaborate on what specific changes have been made in core? That would help us plan how best to align.

dkotter:
Thanks for all the work here @faisal-alvi. I think at this point we're better off closing this out and potentially rethinking the approach here. If you'd like to continue on this PR though, happy to re-open

PR #971, opened 2025-08-12
Fix display of Smart Crop thumbnails on WordPress UI

### LINKED ISSUE #768: Smart crop thumbnails not showing in WordPress UI
### Describe the bug

The smart cropped thumbnails are being generated but are not shown in the WordPress UI.  The smart crop feature appears to be properly generating thumbnails as viewed in the uploads directory, but for some reason are not displayed in the WP UI.  Expectation is that viewing the Media Library in either list or grid views would show the correct smart-cropped thumbnail for the image.

From @dkotter:
> So the problem seems to be the Grid view doesn't use the thumbnail image. For this particular image, the List view uses the 150x150 thumbnail image which is smart cropped. But the Grid view is using the medium image size (136x300 in this case), which isn't smart cropped (it doesn't get cropped at all, just scaled down).

So we may need to check WP core to see if there's a way to filter the grid view to show the correct cropped thumbnail size or perhaps work on a fix upstream in core to allow the thumbnails to be shown in grid view (or at least filterable).

### Steps to Reproduce

1. Ensure ClassifAI is active and the Smart Crop feature (for older versions the Image Processing / Azure setting area) settings is enabled
2. Upload an image to the Media Library
3. View the image and its thumbnail in the uploads directory, see properly smart-cropped image
4. View within the WP UI and see that the default center-crop is still shown

Sample images to test with:
th-2843024898

test

Screenshot 2024-05-14 143311

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

mendezcode:
### Description of the Change

Introduces a filter on  [wp_prepare_attachment_for_js](https://developer.wordpress.org/reference/functions/wp_prepare_attachment_for_js/) and replaces the `medium` thumbnail size (150x300) with the `thumbnail` thumbnail size (150x150), which effectively displays the cropped image thumbnail on the WordPress Media Library. 

> :bulb: Since the `wp_prepare_attachment_for_js` filter works within `wp-ajax.php`, there is no way to access the `$pagenow` global, to possibly filter the current page to the WordPress Media Library and potentially make the thumbnail override on that page only, and not globally (within the JavaScript data, that is).

This change replaces the `medium` thumbnail with the `thumbnail` size only for the JavaScript data used with the frontend Backbone models. It is a global change, but only within JavaScript.
&nbsp;
Screenshot 2025-08-12 at 4 52 15 PM

Closes #768

### How to test the Change

1. Ensure ClassifAI is active and the Smart Crop feature
2. Upload an image to the Media Library
3. View the image and its thumbnail in the uploads directory, see properly smart-cropped image
4. Confirm within the WP UI that the cropped image is shown instead of the default center-crop one

### Changelog Entry

> Fixed - Smart crop thumbnails not showing in WordPress UI

### Credits

Props @kmgalanakis , @dkotter 

### Checklist:

### DIFF
--- a/includes/Classifai/Features/ImageCropping.php
+++ b/includes/Classifai/Features/ImageCropping.php
@@ -170,8 +170,8 @@ class ImageCropping extends Feature {
 	 * @param array $response Array of prepared attachment data.
 	 * @return array
 	 */
-	public function set_media_library_attachment_size( array $response ): array {
-		if ( 'image' !== $response['type'] ) {
+	public function set_media_library_attachment_size( $response ) {
+		if ( ! isset( $response['type'] ) || 'image' !== $response['type'] ) {
 			return $response;
 		}
 
--- a/includes/Classifai/Features/ImageCropping.php
+++ b/includes/Classifai/Features/ImageCropping.php
@@ -63,6 +63,7 @@ class ImageCropping extends Feature {
 		add_action( 'edit_attachment', [ $this, 'maybe_crop_image' ] );
 
 		add_filter( 'attachment_fields_to_edit', [ $this, 'add_rescan_button_to_media_modal' ], 10, 2 );
+		add_filter( 'wp_prepare_attachment_for_js', [ $this, 'set_media_library_attachment_size' ], 10, 1 );
 		add_filter( 'wp_generate_attachment_metadata', [ $this, 'generate_smart_crops' ], 7, 2 );
 	}
 
@@ -161,6 +162,28 @@ class ImageCropping extends Feature {
 		}
 
 		return $metadata;
+	}
+
+	/**
+	 * Adjust the media library attachment size to account for cropped thumbnails.
+	 *
+	 * @param array $response Array of prepared attachment data.
+	 * @return array
+	 */
+	public function set_media_library_attachment_size( array $response ): array {
+		if ( 'image' !== $response['type'] ) {
+			return $response;
+		}
+
+		/*
+		 * Replace the `medium` thumbnail with the original `thumbnail` size, which has been
+		 * smart cropped to show the actual cropped image on the media library thumbnail.
+		 */
+		if ( isset( $response['sizes']['thumbnail'] ) && isset( $response['sizes']['medium'] ) ) {
+			$response['sizes']['medium'] = $response['sizes']['thumbnail'];
+		}
+
+		return $response;
 	}
 
 	/**


iamdharmesh (dismissed):
Thanks a lot for the PR @mendezcode. LGTM.

@dkotter, since we are updating the medium size with the thumbnail one, I’m tagging you here to check if you have any concerns with it. I did a quick check on this and couldn’t find a better solution.

dkotter (commented):
So in testing this out, I'm not seeing this working. I get the exact same images used both on this branch and on `develop`, unless there's something I'm missing here?

iamdharmesh:
> I'm not seeing this working. I get the exact same images used both on this branch and on develop, unless there's something I'm missing here?

Here is how it looks to me. Maybe you should try with regenerate thumbnails?

| on `develop` | on this branch |
| --- | --- |
|Screenshot 2025-08-20 at 4 11 48 PM|Screenshot 2025-08-20 at 4 22 19 PM|

includes/Classifai/Features/ImageCropping.php
```diff
@@ -163,6 +164,28 @@ public function generate_smart_crops( array $metadata, int $attachment_id ): arr
 		return $metadata;
 	}
 
+	/**
+	 * Adjust the media library attachment size to account for cropped thumbnails.
+	 *
+	 * @param array $response Array of prepared attachment data.
+	 * @return array
+	 */
+	public function set_media_library_attachment_size( array $response ): array {
```
dkotter: Because this is hooked to a core WordPress filter, someone else may filter this and return invalid data, which would then cause these type checks to fail. My general approach is if we have a function hooked to a core WordPress filter, we don't add these type hints
```suggestion
	public function set_media_library_attachment_size( $response ) {
```

includes/Classifai/Features/ImageCropping.php
```diff
@@ -163,6 +164,28 @@ public function generate_smart_crops( array $metadata, int $attachment_id ): arr
...
+		if ( 'image' !== $response['type'] ) {
```
dkotter: Similar to the above, if someone else uses this same filter and returns invalid data, `$response` may no longer be an array. We should check if this is an array and if the `type` key is set

dkotter (changes requested):

dkotter:
> > 
> 
> Here is how it looks to me. Maybe you should try with regenerate thumbnails?

Working now, I thought I had the Feature enabled but did not, so the code wasn't loading 😅 

This is working though I left a few comments. I don't necessarily love that this is a global change, does make me a bit worried this could impact other areas unintentionally but not seeing anything doing a quick review of the admin

includes/Classifai/Features/ImageCropping.php
iamdharmesh: Handled in https://github.com/10up/classifai/pull/971/commits/a07975d00db40d02a3106dc3ec7e3a9e00fbd4aa

includes/Classifai/Features/ImageCropping.php
iamdharmesh: Handled in https://github.com/10up/classifai/pull/971/commits/a07975d00db40d02a3106dc3ec7e3a9e00fbd4aa

dkotter (approved):

PR #974, opened 2025-08-13
Add support for generating images locally using Stable Diffusion

dkotter:
### Description of the Change

We have a decent amount of Features that can currently be run locally using Ollama. Ollama only currently supports models that can generate text, not things like images, so our Image Generation is one Feature that can't be used locally.

This PR addresses that by introducing support for Stable Diffusion, run locally using the [Stable Diffusion Web UI tool](https://github.com/AUTOMATIC1111/stable-diffusion-webui). This is fairly easy to set up, though does require some technical knowledge as it requires installing a handful of dependencies and actually running the tool from the command line.

With this PR checked out, you'll see a new Provider for the Image Generation Feature, Stable Diffusion. As long as you have this running locally with an appropriate model, you can select that and then use it to generate images, 100% offline on your own machine. Worth noting here that performance and quality will depend on the model you use and the machine you are running it on. It is fairly slow (at least for me) with a time of about 90 seconds to generate an image.

While I wouldn't expect anyone to use this in production, due to the speed, it is a fun demo and I can see some use cases where the generation time isn't a factor, especially considering cost and privacy tradeoffs.

### How to test the Change

1. First, you'll need to install the Stable Diffusion Web UI tool. They have pretty good instructions to follow depending on your device. If on an Apple Silicon device, [these instructions](https://github.com/AUTOMATIC1111/stable-diffusion-webui/wiki/Installation-on-Apple-Silic) worked for me. Main thing to note is you'll need the right version of Python installed _prior_ to installing and running the tool. This is the main thing that caught me and took a bit to figure out as I had multiple versions of python on my machine
2. Once installed, you'll need to decide which model to download and use. The instructions linked above have a [nice guide](https://github.com/AUTOMATIC1111/stable-diffusion-webui/wiki/Installation-on-Apple-Silicon#downloading-stable-diffusion-models) on how to do this. I downloaded the first model mentioned there, Stable Diffusion 1.4, as it was fairly small (~4GB)
3. Next you'll need to run the tool in API mode: `./webui.sh --api`. This is the second piece that caught me and took me a bit to figure out. If you don't include the `--api` part, the API we need to interact with won't be available. When you run that command, you'll get some output on the command line and it will open a browser window with the web tool. You can test generating images here but this isn't needed for the interaction in ClassifAI
4. Once the tool is running properly, go to the Image Generation Feature in ClassifAI, choose Stable Diffusion as the Provider and ensure you have the right API URL. It defaults to the URL that comes out of the box but if you change that, you'll need to change that here
5. Save the settings to trigger an API request to get the available models. It will default to the first model in the list but you can change to a different model if you've installed multiple
6. Try generating an image and ensure it works. It does take a bit so depending on your server settings, you may encounter timeouts, though I have raised the timeout on the request itself pretty high

### Changelog Entry

> Added - Support for Stable Diffusion as a Provider for Image Generation, allowing the generation of images locally.

### Credits

Props @dkotter

### Checklist:


src/js/settings/components/provider-settings/stable-diffusion.js
```diff
@@ -0,0 +1,167 @@
...
+			<a
+				title={ __( 'Install Stable Diffusion', 'classifai' ) }
+				href="https://github.com/AUTOMATIC1111/stable-diffusion-webui/"
+				target="_blank"
+				rel="noopener noreferrer"
+			>
+				{ ' ' }
+				{ __( 'Download Stable Diffusion', 'classifai' ) }
+			</a>
+		</>
```
iamdharmesh: Maybe we should also include a link to docs for the installation and setup?

iamdharmesh (dismissed):
Looks good to me and I am able to generate images locally.

src/js/settings/components/provider-settings/stable-diffusion.js
dkotter: So this does link to the GitHub repo, which has installation instructions in the README. I'll update this link to go directly to that section but I think that's probably good enough

src/js/settings/components/provider-settings/stable-diffusion.js
```diff
@@ -0,0 +1,167 @@
...
+
+	const Description = () => (
+		<>
+			{ __(
+				"URL of the locally hosted Stable Diffusion instance. Defaults to http://127.0.0.1:7860/. Don't have Stable Diffusion installed yet?",
+				'classifai'
+			) }
+			<a
+				title={ __( 'Install Stable Diffusion', 'classifai' ) }
+				href="https://github.com/AUTOMATIC1111/stable-diffusion-webui/"
```
dkotter:
```suggestion
				href="https://github.com/AUTOMATIC1111/stable-diffusion-webui/#installation-and-running"
```

iamdharmesh (dismissed):

PR #976, opened 2025-08-15
Add support for generating images using models hosted on Together AI

dkotter:
### Description of the Change

This PR brings support for the Together AI platform within the Image Generation Feature. [Together AI](https://www.together.ai/) allows you to easily make API requests to lots of different models that they host, with many of these models being open source.

For image generation, they currently have 13 different models you can choose from, with one of those ([FLUX.1 [schnell] Free](https://www.together.ai/models/flux-1-schnell)) being completely free, though limited to I think 3 requests per minute and only a single image generated at a time. Other models will provide better quality results with fairly reasonable pricing (typically around 5 cents per image).

I envision this being the first step to integrating Together AI across all of our other Features, as they have multiple models for  text generation, transcription, text to speech, moderation and embeddings. By adding this as a Provider for all of those, someone can more easily use various open source models and have better control over where their data is sent. I see this as a middle ground between someone going the completely self-hosted route (which we support using Ollama) and someone using a non-open-source solution like OpenAI.

### How to test the Change

1. You'll need to create a Together AI account first. This was pretty easy and did not require any payment information to be entered. In fact, I was prompted after to add some additional business details and in return, was given a $1 credit to try things
2. Once your account is created, generate an API key
3. Configure the Image Generation Feature to use Together AI, pasting in your API key
4. Save settings and if valid, the model dropdown will be populated with all available image models
5. Recommend choosing the `FLUX.1 [schnell] Free` model to avoid any costs
6. Try generating an image and ensure that works

### Changelog Entry

> Added - Support for Together AI as a Provider for Image Generation, allowing the generation of images using open source models hosted by Together AI

### Credits

Props @dkotter

### Checklist:


src/js/settings/components/provider-settings/together-ai-images.js
```diff
@@ -0,0 +1,140 @@
...
+					'Choose the model you want to use.',
+					'classifai'
+				) }
+			>
+				<SelectControl
+					id={ `${ providerName }_model` }
+					onChange={ ( value ) => onChange( { model: value } ) }
+					value={ providerSettings?.model || '' }
+					options={ models }
+					disabled={ ! isConfigured }
```
iamdharmesh:
```suggestion
					disabled={ models.length <= 1 }
```

`isConfigured` get updated to false when we click on edit button for the configured provider. Definitely need to update variable name to something else to make it clear, but for now, in this PR would suggest to update it to something like this. So, model select dropdown get enabled once it has models to show.

iamdharmesh (dismissed):
Code looks good and I have tested it using `FLUX.1 [schnell] Free` model and it works well. Add minor note to discuss otherwise we are good to go here.

src/js/settings/components/provider-settings/together-ai-images.js
dkotter: I'm fine with this change, noting my thought here was if someone wanted to change their Provider settings (in this case, the API key) a different account may have different models it supports, so having this dropdown be disabled while those settings are being modified was my thought. That said, this basically accomplishes the same thing

iamdharmesh (approved):

PR #977, opened 2025-08-15
Allow setting the number of responses when using Ollama as the Provider in Title Generation and Content Resizing Features

dkotter:
### Description of the Change

If using OpenAI as the Provider for the Title Generation or Content Resizing Features, we have a setting that allows you to choose how many responses are returned. This is something OpenAI supports natively in their API. If using Ollama, this setting is removed as Ollama itself does not support that API parameter.

This PR brings that setting over to these Features when using Ollama as a Provider and behind the scenes, we end up making multiple requests to support this. For instance, if I want 3 title suggestions, that will end up making 3 separate requests to Ollama to generate those. This isn't perfect, as you can get the same results for each request but in my testing, these results to tend to be different often enough to be of value.

Also changed those two API requests to use JSON structured output, making it easier for us to be sure the response we get is what we want. For instance, prior to that, Ollama would often return something like `Here's the generated title: TITLE`. That is hard for us to properly parse and so we end up just showing that to the user. You can usually address this with a more specific system prompt but seems that using structured output resolves this.

### How to test the Change

1. Install Ollama if you don't have it already and install at least one model
2. Configure both Title Generation and Content Resizing to use Ollama
3. Test that getting a single response from both still works
4. Change the `Number of suggestions` setting to something higher than 1
5. Ensure you get the right number of responses

### Changelog Entry

> Added - Number of suggestions setting added to the Title Generation and Content Resizing Features when using Ollama
> Changed - Request JSON structured output when using Ollama with the Title Generation and Content Resizing Features

### Credits

Props @dkotter, @jeffpaul 

### Checklist:

### DIFF
--- a/includes/Classifai/Providers/Localhost/Ollama.php
+++ b/includes/Classifai/Providers/Localhost/Ollama.php
@@ -478,30 +478,51 @@ class Ollama extends Provider {
 					],
 				],
 				'stream'   => false,
+				'format'   => [
+					'type'       => 'object',
+					'properties' => [
+						'content' => [
+							'type' => 'string',
+						],
+					],
+					'required'   => [ 'content' ],
+				],
 			],
 			$post_id
 		);
 
-		// Make our API request.
-		$request  = new APIRequest( 'test' );
-		$response = $request->post(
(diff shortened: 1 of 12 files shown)


iamdharmesh (approved):
Code looks good and it tests well.

PR #972, opened 2025-08-13
Make all external request VIP-compliant and add dedicated VIPCS workflow

MiguelAxcar:
### Description of the Change

This PR intends to complete VIP remote I/O compliance started by PR #901 and add a dedicated VIPCS check.

#### What changed

- Added a generic wrapper `safe_wp_remote_request( $method, $url, $args )` preserving caller args with wrappers thin `safe_wp_remote_get()` and `safe_wp_remote_post()`
- Explicit timeouts are kept (Azure Speech 20s, RSS parser 60s)
- Use `vip_safe_wp_remote_request()` when available, falling back to `wp_remote_request()`
- Add a default UA `ClassifAI/<version>` if none provided respecting if there is an override coming on the call for all requests coming through `safe_wp_remote_request()` to make it easier to trace requests / WAF to help support / debugging, allowing ClassifAI to be allow listed or handled separately from generic WP traffic, 
- Replaced remaining direct HTTP calls with the safe wrappers for all providers / services
- License validation now uses `safe_wp_remote_post()`
- Removed remote `file_get_contents()` usage so now Ollama Multimodal now fetch remote images via HTTP API `safe_file_get_contents()` helper 
- CLI ClassifAI text supports remote input via safe HTTP, while local files still use native read.
- AudioTranscriptsGeneration::remote_url_to_path() using `safe_wp_remote_get()`
- Updated hook docs snippet to use the safe GET helper.
- Added VIPCS specific workflow on `.github/workflows/vipcs.yml` plus README badge.

#### Notes / behavior

Dev only RSS Importer CLI that uses `simplexml_load_file()` (it doesn't have compliance to VIP) is not handled here, we created a separate issue to it: #973 

Closes #791 
Refs #901

#### Code check

- Confirm no direct calls remain in code: `wp_remote_(get|post|request)` and `file_get_contents` (should only appear in the helper fallback)

#### Providers

Verify success and errors.

- OpenAI: generate titles / excerpts / content, verify success and errors.
- GoogleAI/xAI: run a basic text / gen call
- Azure: 
   - Speech: connect and synthesize; confirm 20s timeout honored.
   - Read: submit document and poll status.
   - Personalizer: rank + reward endpoints.
   - ComputerVision/SmartCropping: run a scan/crop request.
   - Watson: GET/POST flows, verify other verbs route through the generic wrapper.
   - License check in settings, verify request succeeds.

#### Ollama Multimodal

Run alt-text / OCR / tags using a remote image URL, and verify it works

#### CLI

[...]

PR #987, opened 2025-08-25
Fix VIP compatibility

### LINKED ISSUE #973: Remove legacy RSS Importer WP-CLI command

While working on #901 we found the dev only RSS importer CLI command uses `simplexml_load_file()` which is not allowed on VIP. This command is old and for devs only, and not a supported feature, so no reason for making it VIP compliant as it doesn't add value to the user.

### Designs

_No response_

- Delete RSS Importer command
- Remove any related code, such as registration / wiring
- Search and remove any mentions for the command on the documentation

### Code of Conduct

av3nger:
### Description of the Change

Removes the RSS Importer WP-CLI command, as this command was there for legacy reasons and is no longer used.

Closes #973

### How to test the Change
- Verify file changes look accurate

### Credits
Props: @av3nger 

### Checklist:


dkotter:
So this file was purposely not updated coming out of #972 because it is a legacy thing that is better off just being removed. A separate issue was opened for this (#973). I'd suggest updating this PR to remove this file instead of updating the use of `wp_remote_get`

av3nger:
@dkotter , thank you for the insight. I've double checked and the class is not used anywhere in the plugin. I have removed it from the plugin.

dkotter (approved):

PR #988, opened 2025-08-25
Consolidate API request logic with HTTPClient abstraction

MiguelAxcar:
### Description of the Change

This PR intends to extract common HTTP functionality into a new `HTTPClient` base class.  

#### Reasons
- Reduce duplicated logic across providers and improve architecture  
- Open room for integrating https://github.com/WordPress/php-ai-client once it has a release
- Minor: honestly, I was _itching_ to do this after seeing some repeated code while working on [#972](https://github.com/10up/classifai/pull/972)  

Integrating https://github.com/WordPress/php-ai-client may be a good idea for the future as it would give ClassifAI a standardized way to communicate with AI providers, reduce the amount of custom request code maintained, and align with a broader WP ecosys standard.

My idea / proposal is to land this refactor first to centralize the shared HTTP code in `HTTPClient`, handling potential adjustments, then wait for php-ai-client to have a release then add a small bridge with minimal changes to integrate it.

### How to test the Change

- Test the whole Critical Flow ([instructions](https://github.com/10up/classifai/wiki/Testing-Instructions-for-Critical-Flow))
- Verify success and errors for all providers and features
- Run a full feature and regression test
- Click through all admin UI flows, verify that no new notices or fatal errors will show.

### Changelog

> Developer - Consolidate API request logic with HTTPClient abstraction
> Developer - Add HTTP client for local providers
> Fixed - Fatal error thrown by `is_embeddings_generation_in_progress()` when provider key is invalid

### Credits
Props @MiguelAxcar 

### Checklist:


includes/Classifai/Providers/HTTPClient.php
```diff
@@ -0,0 +1,375 @@
...
+	protected function parse_response( array $response ) {
+		$code = wp_remote_retrieve_response_code( $response );
+		$body = wp_remote_retrieve_body( $response );
+
+		// Error responses
+		if ( $code >= 400 ) {
+			$response_message = wp_remote_retrieve_response_message( $response );
+			$status_text      = $response_message ? $response_message : __( 'Unknown error', 'classifai' );
+
+			// Try to extract specific error message from the server response
```
MiguelAxcar, line 205: Former code had chance for improvement:
```php
$message = $json['error']['message'] ?? esc_html__( 'An error occurred', 'classifai' );
```

It should be checking on `$json['error']['message']` only if `$json['error']` isn't a string itself.

It's expected now for users to see much more details from providers on errors instead of former `An error ocurred` message only.

includes/Classifai/Providers/Localhost/Ollama.php
```diff
@@ -6,7 +6,7 @@
 namespace Classifai\Providers\Localhost;
 
 use Classifai\Providers\Provider;
-use Classifai\Providers\OpenAI\APIRequest;
+use Classifai\Providers\Localhost\APIRequest;
```
MiguelAxcar, line 9: No longer using OpenAI for local providers, now they have its own clients

includes/Classifai/Providers/OpenAI/Embeddings.php
```diff
@@ -1728,6 +1728,10 @@ public function get_debug_information(): array {
 	 * @return bool
 	 */
 	public function is_embeddings_generation_in_progress(): bool {
+		if ( null === self::$scheduler_instance ) {
+			return false;
+		}
```
MiguelAxcar, line 1733: This was throwing a fatal error when key was empty / with error.

includes/Classifai/Providers/Watson/APIRequest.php
```diff
@@ -38,79 +36,24 @@ class APIRequest {
...
-	 * @param array  $options Additional query params
-	 * @return array|\WP_Error
-	 */
-	public function get( string $url, array $options = [] ) {
-		$this->add_headers( $options );
-		return $this->get_result( safe_wp_remote_get( $url, $options ) );
+	public function __construct( string $api_key = '', string $feature = '' ) {
+		parent::__construct( $api_key, $feature );
+		$this->username = get_username();
+		$this->password = get_password();
```
MiguelAxcar, line 47: Kept specificities from Watson auth

iamdharmesh:
Thanks for the PR @MiguelAxcar.

I noticed we already have a similar ongoing PR (#994), which is doing the same thing along with integrating the php-ai-client. It’s currently implemented only for OpenAI, with plans to extend it to other providers as well. @dkotter, since you’re the author of that PR, I’d like your feedback on the potential next steps for this one. Whether we should update #994 to extend the base class for all providers (I’d prefer to keep it small initially and focus on php-ai-client or wp-ai-client integration, then handle other providers in a follow-up), tweak this PR to make it useful for #994, or keep it on hold until the ongoing discussions in #994 are finalized. Thanks!

dkotter:
> Thanks for the PR @MiguelAxcar.
> 
> I noticed we already have a similar ongoing PR (#994), which is doing the same thing along with integrating the php-ai-client. It’s currently implemented only for OpenAI, with plans to extend it to other providers as well. @dkotter, since you’re the author of that PR, I’d like your feedback on the potential next steps for this one. Whether we should update #994 to extend the base class for all providers (I’d prefer to keep it small initially and focus on php-ai-client or wp-ai-client integration, then handle other providers in a follow-up), tweak this PR to make it useful for #994, or keep it on hold until the ongoing discussions in #994 are finalized. Thanks!

Yes, thanks for the work here @MiguelAxcar and I did take a look at this to see if there was anything I should be implementing with the work I was doing in #994. That work is still early stage, as the PHP AI Client (and now the WP AI Client) are still in early development, but we're likely to take at the very least a similar approach that is in that PR.

For now, I'd like to leave this PR open but we're likely to consolidate our approaches here once the AI Client is more stable. I'll come back to this PR at that point and decide next best steps.

MiguelAxcar:
Thanks @dkotter @iamdharmesh for the feedback.

My intent here is to first land a provider-agnostic HTTP layer (auth/headers/errors/timeouts/specificities) and migrate providers onto it, while `php-ai-client` gets more matured and reach a release.

As mentioned by Dharmesh https://github.com/10up/classifai/pull/994 experiment bridges only OpenAI; great exploration of course, awesome to see this in action first hand btw, but potentially never shipped or will need changes to work. Given that, I would suggest keeping this (or another) PR focused on consolidating the HTTP client across providers, then introduce the PHP AI Client bridge as a small, targeted follow-up, using results from https://github.com/10up/classifai/pull/994 experiment / exploration, 

That order keeps the blast radius small, preserves parity, and makes the `php-ai-client` bridge mostly mechanical per provider.

Happy to help split/cherry-pick pieces to get this in smoothly, or any help I could provide for this.

jeffpaul:
Punting to 3.9.0 when we're likely near WP 7.0 Beta 1 and know whether the WP AI Client made it into core and how we really need to integrate/extend there.

dkotter:
Thanks for the work on this one @MiguelAxcar. While I agree with the overall idea here, we're looking to make some significant changes to ClassifAI, in particular using the new AI Client that shipped in WordPress 7.0 for all of our API requests. As such, going to close this one out

PR #994, opened 2025-09-04
WIP: Start work on integrating the PHP AI Client package

dkotter:
### Description of the Change

> [!NOTE]  
> This is an early test of the new [PHP AI Client package](https://github.com/WordPress/php-ai-client) and things will likely change as that matures. Opening this PR both for discussion and as an example of how that package can be implemented, but this PR may never ship or may change to use the [WordPress AI Client](https://github.com/WordPress/wp-ai-client) package.

This PR brings in the PHP AI Client package as a new composer dependency and starts to use that where possible. At the moment, only implemented for OpenAI and only for these Features:

- Title Generation
- Excerpt Generation
- Content Resizing
- Key Takeaways
- Descriptive Text Generator
- Image Tags Generator
- Image Text Extraction
- Image Generation

We introduce a new base `APIRequest` class that Providers can extend for authentication purposes. This class has handling for existing API requests (GET, POST, POST FORM) so any existing functionality still works as expected but then introduces new handling for the PHP AI Client package, which allows Providers to opt-in to using that where it makes sense.

Potential next steps:

- [ ] Move the Google Provider over to the new approach, where that Provider is supported in the above Feature list
- [ ] Once v0.2.0 of the PHP AI Client package is released, move the Content Generation handling over, as right now there's not a great way to handle the back and forth messaging of that Feature but this is planned for v0.2.0
- [ ] Potentially look at extending the PHP AI Client package to support either other AI Providers (like Ollama) or other API features (like moderation, text to speech, speech to text). These may be things worth pushing upstream instead of leaving in ClassifAI

> [!NOTE]  
> E2E tests are failing here because our test plugin that hooks in and captures HTTP requests and mocks those no longer works with this PHP AI Client package, as it doesn't use the WordPress HTTP API and thus there aren't any hooks firing there. If we continue with this approach, we'll need to figure out a different way to capture and mock those requests. I think if we end up using the WordPress AI Client package though, that will be solved for us.

### How to test the Change

Coming soon...

### Changelog Entry

> Added - New feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

### Credits

Props @dkotter

### Checklist:


jeffpaul:
> Potentially look at extending the PHP AI Client package to support either other AI Providers (like Ollama) or other API features (like moderation, text to speech, speech to text). These may be things worth pushing upstream instead of leaving in ClassifAI

My hope is we can push those upstream, though if the core team decides not to include additional providers or features then we might consider releasing those as their own extensions of the core PHP/WP clients versus keeping hardcoded within ClassifAI.

dkotter:
This was some early exploratory work that may be useful at a future point but for now, closing this out. We do want to fully utilize the AI Client in the 4.0 release of ClassifAI but will be easier to start from scratch than continue this PR

PR #1001, opened 2025-09-08
Update Azure OpenAI to the latest GA API version

dkotter:
### Description of the Change

We had a report that if using Azure OpenAI with the Key Takeaways Feature, the response would always fail because we set the `reponse_type` to be `json_schema` and it seems that isn't supported in the API version we were using.

This PR fixes that by updating the API version from `2023-05-15` to the latest GA release, `2024-10-21`. This release does [support structured outputs](https://learn.microsoft.com/en-us/azure/ai-foundry/openai/how-to/structured-outputs?tabs=python-secure%2Cdotnet-entra-id&pivots=programming-language-csharp#api-support) but worth noting that only works if using a [compatible model](https://learn.microsoft.com/en-us/azure/ai-foundry/openai/how-to/structured-outputs?tabs=python-secure%2Cdotnet-entra-id&pivots=programming-language-csharp#supported-models).

In addition, this PR adds a new helper method to get the API version and wraps the response of that in a new filter, `classifai_azure_openai_api_version`, making it easier for others to change the API version that is used (if, for instance, you want to use a preview version).

And finally, in testing to ensure this new API version works for all the Features we support, discovered that the Content Generation Feature wasn't working due to an existing bug so that is fixed here as well.

### How to test the Change

Azure OpenAI supports the following Language Processing Features:

- Title Generation
- Excerpt Generation
- Content Generation
- Content Resizing
- Key Takeaways

Configure each of these to use Azure OpenAI and ensure they all still work. Note that for the Key Takeaways Feature, you need to test with a compatible model (not GPT-3.5 Turbo for instance).

### Credits

Props @gsarig, @dkotter 

### Checklist:

### DIFF
--- a/includes/Classifai/Providers/Azure/OpenAI.php
+++ b/includes/Classifai/Providers/Azure/OpenAI.php
@@ -273,7 +273,8 @@ class OpenAI extends Provider {
 			( $feature instanceof ContentResizing ||
 			$feature instanceof ExcerptGeneration ||
 			$feature instanceof TitleGeneration ||
-			$feature instanceof KeyTakeaways ) &&
+			$feature instanceof KeyTakeaways ||
+			$feature instanceof ContentGeneration ) &&
 			$deployment
 		) {
 			$endpoint = trailingslashit( $endpoint ) . str_replace( '{deployment-id}', $deployment, $this->chat_completion_url );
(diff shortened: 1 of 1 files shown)


iamdharmesh (approved):
@dkotter Thanks for the fixing this. Code looks good and it tests well. (I skipped testing of key takeaway block as it's not compatible with GPT-3.5 Turbo)

dkotter:
@gsarig Let me know if you have time to test this and ensure it fixes the issues you were running into, prior to us merging this in. Thanks!

gsarig:
> @gsarig Let me know if you have time to test this and ensure it fixes the issues you were running into, prior to us merging this in. Thanks!

Hi @dkotter Yes, it seems to work as expected, thank you!

PR #1002, opened 2025-09-09
Add ElevenLabs as a Provider for Text to Speech

### LINKED ISSUE #993: Add ElevenLabs as a Provider for Text to Speech

Following up on #986 and PR #991, we should add ElevenLabs as a supported Text-to-Speech (TTS) provider.  This enhancement leverages the new provider architecture introduced in #991 (hopefully making the integration effort here a bit easier ;).

Noting @dkotter's comment in that PR:

> Note: probably worth a followup to add ElevenLabs as a Text to Speech Provider once this is merged in, as that will be pretty easy at that point.

- [ ] Add ElevenLabs as an available Text-to-Speech provider in settings with respective ElevenLabs API credentials.
- [ ] Support selecting from available ElevenLabs voices within the provider settings UI.
- [ ] Integrate with the existing TTS generation flows, ensuring parity with other providers in terms of:
  - [ ] Generating audio from a post
  - [ ] Downloading the generated file
  - [ ] Displaying the audio file on posts/pages via the frontend "read to me" feature
- [ ] Include helpful inline documentation and a link to ElevenLabs’ API docs.

### Designs

_No response_

_No response_

### Code of Conduct

iamdharmesh:
### Description of the Change
PR adds ElevenLabs as a provider for the Text to Speech feature.

Closes #993 

### How to test the Change
1. If you don't have an ElevenLabs account, you'll need to sign up for one (it is free)
2. In your account, create an API key and ensure you have the Text to Speech endpoint enabled and Read access to the Models information (which we use to validate the connection) and Voices
3. Turn on Text to Speech, select ElevenLabs and enter your API key
4. Save settings and ensure no error happens
5. If desired, change the model and voices as per your need
6. Publish post and ensure audio generated properly.
7. Change content of post, enable audio generation from sidebar panel and save post, ensure audio generated properly.

### Credits

Props @dkotter @jeffpaul @iamdharmesh 

### Checklist:


includes/Classifai/Providers/ElevenLabs/TextToSpeech.php
```diff
@@ -0,0 +1,287 @@
...
+		$new_settings[ static::ID ]['voice'] = sanitize_text_field( $new_settings[ static::ID ]['voice'] ?? $settings[ static::ID ]['voice'] );
+
+		// Ensure the model being saved is valid. If not valid or we don't have one, use the first model.
+		if ( ! in_array( $new_settings[ static::ID ]['model'], array_column( $new_settings[ static::ID ]['models'], 'id' ), true ) ) {
+			$new_settings[ static::ID ]['model'] = array_column( $new_settings[ static::ID ]['models'], 'id' )[0];
+		}
+
+		// Ensure the voice being saved is valid. If not valid or we don't have one, use the first voice.
+		if ( ! in_array( $new_settings[ static::ID ]['voice'], array_column( $new_settings[ static::ID ]['voices'], 'id' ), true ) ) {
+			$new_settings[ static::ID ]['voice'] = array_column( $new_settings[ static::ID ]['voices'], 'id' )[0];
```
dkotter, line 104: I ran into an issue when testing where this was throwing an error because I hadn't turned on access to the Voices endpoint in my account and thus we set voices to an empty array, causing this line to fail.

I fixed this by adding `! empty( $new_settings[ static::ID ]['voices'] )` to this `if` statement, which we probably want to do here. But in addition, would be nice if in that scenario, we output an error message letting the user know things aren't configured properly. Because at the moment, we have no voices to choose from but otherwise it seems the Feature is completely set up and ready to work. If I'm following their API correctly, having a voice is crucial for it to work so if we have no voice set, I think we need to show an error message and not allow the Feature to work

dkotter (changes requested):

includes/Classifai/Providers/ElevenLabs/ElevenLabs.php
```diff
@@ -174,6 +182,16 @@ public function sanitize_api_key_settings( array $new_settings = [], array $sett
...
 			$new_settings[ static::ID ]['authenticated'] = true;
 			$new_settings[ static::ID ]['models']        = $models;
+
+			if ( $this->feature_instance instanceof TextToSpeech ) {
+				// Get the available voices.
+				$voices = $this->get_voices( $new_settings[ static::ID ]['api_key'] ?? '' );
+				if ( ! is_wp_error( $voices ) && ! empty( $voices ) ) {
+					$new_settings[ static::ID ]['voices'] = $voices;
+				} else {
+					$new_settings[ static::ID ]['voices'] = [];
```
dkotter: Going along with my other comment, I think if we return the WP_Error object here instead of an empty array, we could then output that error message, giving better feedback to the user that something is wrong

includes/Classifai/Providers/ElevenLabs/TextToSpeech.php
iamdharmesh: Thanks for catching this. handled in https://github.com/10up/classifai/pull/1002/commits/4015d5565165a934868056d1e806a3c2bf111201

includes/Classifai/Providers/ElevenLabs/ElevenLabs.php
iamdharmesh: handled in https://github.com/10up/classifai/commit/4015d5565165a934868056d1e806a3c2bf111201

dkotter (approved):

PR #1003, opened 2025-09-10
WIP: Start work on integrating the new WordPress Abilities API

dkotter:
### Description of the Change

> [!NOTE]  
> Similar to #994, this is an early test of the new [WordPress Abilities API](https://github.com/WordPress/abilities-api) and things will likely change as that matures. Opening this PR both for discussion and as an example of how we may want to implement this.

This PR brings in the WordPress Abilities API as a new composer dependency and integrates that into the Title Generation and Image Generation Features. Within the existing base Feature class, if a Feature is enabled and configured, we hook into the new `abilities_api_init` hook to register any needed abilities. The thought is that each individual Feature will have their own `register_ability` method that will be called here.

I set this up in the Title Generation and Image Generation Features, registering the ability, adding a custom permission callback and adding a custom execution callback. All of this is structured very similarly to our existing custom REST endpoint that this Feature uses. I've then updated the front-end handling to use the new Abilities API endpoint instead of our custom endpoints, which required some slight changes to how the data is structured.

Points open to discussion:

- Should we stick with our custom REST endpoints or use the new Abilities endpoints? My opinion (and what I did here) was to use the new Abilities endpoints. Seems like if we're going to register something as an ability, might as well go all in
- If we use the new Abilities endpoints, should we remove the existing custom endpoints? I have not done that in this PR though I would suggest we do that (to avoid duplicate code), noting that could be seen as a breaking change
- Once this gets merged into Core (slated for WP 6.9) should we continue to bundle the package with ClassifAI or set our minimum WordPress version to 6.9?

Next steps:

- [ ] Determine if the approach in this PR is the best way to integrate the Abilities API
- [ ] Integrate the Abilities API into all other Features
- [ ] Remove existing REST endpoints

### How to test the Change

1. Checkout this branch and run `composer install` and `npm run build`
2. Configure the Title Generation Feature
3. Confirm that this Feature still works as expected
4. Configure the Image Generation Feature
5. Confirm that this Feature still works as expected

### Changelog Entry

> Added - New feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

### Credits

Props @dkotter

### Checklist:


includes/Classifai/Features/Feature.php
```diff
@@ -84,6 +88,25 @@ public function setup() {
...
 	}
 
+	/**
+	 * Register an ability after the abilities API is initialized.
+	 *
+	 * Only fires if the Feature is enabled and configured.
+	 */
+	public function abilities_api_init() {
+		if ( function_exists( 'wp_register_ability' ) ) {
+			$this->register_ability();
```
dkotter, line 98: I've set this up to match what we already do in this class but wondering if it would be better to have a `method_exists` check here and remove the empty method below?

includes/Classifai/Features/TitleGeneration.php
```diff
@@ -95,6 +95,61 @@ public function feature_setup() {
...
+							'sanitize_callback' => 'absint',
+							'description'       => esc_html__( 'Number of titles to generate', 'classifai' ),
+						],
+					],
+					'required'   => [
+						'id',
+					],
+				],
+				'output_schema'       => [
+					'type'       => 'object',
```
dkotter, line 133: In my testing, seems we need to return an object in order to support any `WP_Error`'s that may return. This does mean we have to change the structure of what we return to be more than just a simple array unfortunately

dkotter:
This was some early exploratory work that may be useful at a future point but for now, closing this out. Will likely look to utilize the Abilities API for the 4.0 release but a new PR will be easier at that point

PR #1005, opened 2025-09-10
When requesting multiple results, increase the `temperature` parameter

### LINKED ISSUE #996: Consider modifying the "temperature" param when requesting multiple results

In the Title Generation Feature, we have an option to request multiple results. Behind the scenes, this sets a param, like `n`, and the LLM will respond with multiple results. But it uses the exact same prompt for all those results. This can end up producing very similar, if not the exact same set of results.

Right now we hardcode the `temperature` param to `0.9`. According to OpenAI, this value can be between 0 and 2 with higher values making the output more random. Would be interesting to test setting this to a higher value and seeing if that helps get less similar results.

My thought is use a default of `0.9` or so and for each extra item someone is requesting, modify that slightly. So if they want 2 results, use `1.0`, if they want 3 results use `1.1`, etc. Not sure if the match works exactly but something along those lines. And then see if that helps get more random results.

### Designs

_No response_

_No response_

### Code of Conduct

dkotter:
### Description of the Change

Our Title Generation and Resize Content Features allow you to choose how many results you want returned (if using OpenAI, Azure OpenAI or xAI). It was reported though that often the results returned are pretty similar, if not exactly the same. While we don't have control over what results the LLMs return, one parameter we currently set is `temperature`. This is supposed to determine how random the results are, from 0 to 2, 0 meaning the results should be the same each time and 2 meaning you should see significant randomness.

We default that to `0.9` and in this PR, a new helper function has been added that will take that default value and increase it proportionally to the number of results you want. So if you want 2 results, it will go from `0.9` to `1.1`. If you want 5 it goes from `0.9` to `1.4`. If you want 10 it goes from `0.9` to `1.9`. In theory, this should mean you get more random results as you request more results, hopefully making each result more unique.

In addition, noticed we didn't have any minimum or maximum attributes on these selectors, meaning someone could enter a negative number or a super high number. Have added that in this PR to limit these from 1 to 10. There may be someone that is currently requesting more than 10 results but I'm assuming that's rare and we can adjust if someone complains :)

Closes #996

### How to test the Change

1. Enable the Title Generation and Content Resizing Features, with either OpenAI, Azure OpenAI or xAI
2. Try out these Features with different combinations of results
3. Ensure things work and you seem to be getting unique results (kind of hard to tell this, mostly a judgement call)

### Changelog Entry

> Added - Increase the temperature value when requesting more results, hopefully leading to more unique results
> Changed - Limit our number of suggestions input from 1 to 10

### Credits

Props @dkotter

### Checklist:

### DIFF
--- a/includes/Classifai/Helpers.php
+++ b/includes/Classifai/Helpers.php
@@ -956,3 +956,23 @@ function safe_wp_remote_post( string $url, array $args = [] ) {
 	return safe_wp_remote_request( 'POST', $url, $args );
 }
 
+/**
+ * Get the temperature for the request.
+ *
+ * We increase the base temperature proportionally
+ * to the number of results, ensuring it never exceeds 2.
+ *
+ * The goal here is to get more diverse results when
+ * we are requesting more results.
+ *
+ * @param float $temperature The temperature.
+ * @param int   $results The number of results.
+ * @return float The temperature.
+ */
+function get_temperature( float $temperature, int $results = 1 ): float {
+	if ( 1 === $results ) {
+		return $temperature;
+	}
+
+	return (float) min( 2.0, $temperature + ( $results / 10 ) );
+}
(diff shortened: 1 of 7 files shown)


iamdharmesh (approved):
Thanks @dkotter. LGTM.

PR #1010, opened 2025-09-15
WIP: Start work on integrating the new WordPress MCP Adapter

dkotter:
### Description of the Change

> [!NOTE]  
> This is branched from #1003 as the [WordPress MCP Adapter](https://github.com/WordPress/mcp-adapter) requires the Abilities API. And similar to that PR, this is an early test of adding MCP support to ClassifAI; opening this PR both for discussion and as an example of how we may want to implement this.

This PR brings in the WordPress MCP Adapter as a new composer dependency and creates a new MCP server with the Title Generation Feature as it's only tool for now. With this code in place, a new REST endpoint is registered (`local.test/wp-json/classifai/mcp`), that allows you to see all available tools and run individual tools.

Worth noting by default, the MCP server requires you to be logged in (this can be overridden but I've left that for now). And the Generate Titles Feature requires you to be logged in and have access, so there are two separate permission checks that you must pass here.

### How to test the Change

1. Ensure the Title Generation Feature is on and configured properly.
2. Make a curl request to the MCP endpoint to list out tools, passing in proper credentials:
```bash
curl -X POST "https://local.test/wp-json/classifai/mcp/" \
--user "admin:password" \
-H "Content-Type: application/json" \
-d '{"method": "tools/list"}'
```
3. Make a curl request to the MCP endpoint to run the generate titles tool, passing in proper credentials:
```bash
curl -X POST "https://local.test/wp-json/classifai/mcp/" \
--user "admin:password" \
-H "Content-Type: application/json" \
-d '{
    "method": "tools/call",
    "params": {
        "name": "classifai-generate-title",
        "arguments": {
            "id": 1
        }
    }
}'
```

### Changelog Entry

> Added - New feature
> Removed - Feature
> Fixed - Bug fix
> Security - Vulnerability

### Credits

Props @dkotter

### Checklist:


includes/Classifai/Plugin.php
```diff
@@ -42,6 +52,10 @@ public function enable() {
 		add_filter( 'plugin_action_links_' . CLASSIFAI_PLUGIN_BASENAME, [ $this, 'filter_plugin_action_links' ] );
 		add_filter( 'robots_txt', [ $this, 'maybe_block_ai_crawlers' ] );
 		add_action( 'after_classifai_init', [ $this, 'load_action_scheduler' ] );
+
+		// Initialize the MCP adapter so the hooks are loaded.
+		$this->mcp_adapter = McpAdapter::instance();
```
dkotter: In my testing, this was needed to ensure the MCP code was loaded. Not ideal but I'm assuming this will get better as this package matures

dkotter: So looked at this a bit closer and the issue here is when loading the plugin via composer, composer will add the autoloaded files from the plugin to our autoloader but that doesn't include the main plugin file (as WordPress takes care of loading that). In our case, we aren't loading this as a normal plugin and so that main file never loads, which is what initializes everything.

If you look at that file, it does this:
```php
// Load the plugin.
if ( class_exists( WP\MCP\Plugin::class ) ) {
	WP\MCP\Plugin::instance();
}
```

That ends up calling `McpAdapter::instance();` which is what we are doing here. So we can either leave this as-is or we can swap this out for the above

dkotter:
This was some early exploratory work that may be useful at a future point but for now, closing this out

PR #1022, opened 2026-01-06
Fix timeouts that can happen for long running tasks on a WP VIP environment

dkotter:
### Description of the Change

We had a report of timeouts happening when using the Text to Speech Feature. After getting more information, they were running into timeouts after 5 seconds while the timeout we set within that Feature is 90 seconds.

Investigating further we found out the problem for them was they are on a WP VIP site, which coming out of #972, we try and filter all remote requests for sites hosted there through the `vip_safe_wp_remote_request` function instead of `wp_remote_request`.

This function limits the HTTP timeout value to 5 seconds though, which is not enough for any longer running requests (like generating speech or images). This was missed as part of #972 so addressing that in this PR.

I've decided that most requests we make don't benefit from caching and could often take longer than that 5 second limit. Because of this, I've modified our helper functions to default to _not_ using the VIP functions and only use those for requests we know are fast and should return the same result each time (like authentication requests or requests to get available models).

All other requests (generating text, generating images, generating speech, etc) will never use the VIP request functions even if those functions exist.

### How to test the Change

This is a fairly broad change so will need to test existing Features to ensure they still work as expected, in particular:

- Title generation
- Text to Speech
- Speech to Text
- Image generation
- Descriptive text generation

### Changelog Entry

> Changed - For any HTTP requests that need extra time to run, ensure those don't utilize the WP VIP functions

### Credits

Props @jhudson-ent, @dkotter

### Checklist:


peterwilsoncc:
Reviewing the remote requests documentation on VIP, [they note](https://docs.wpvip.com/databases/optimize-queries/retrieving-remote-data/#h-uncached-remote-requests):

> If for some reason an uncached remote request is required (e.g., to ping an external service during post publish) then the [WordPress HTTP API](https://developer.wordpress.org/plugins/http-api/) should be used rather than directly using cURL or another method.
>
> Caution
> For speed and performance reasons, uncached remote requests should never run on the front end of a site.

@dkotter Should Classifai requests ever be cached, they run in the admin and vary depending on the prompt so I am wondering if we should always use the WP Core API to avoid caching requests that are likely to be made once and once only. Thoughts?

dkotter:
> Reviewing the remote requests documentation on VIP, [they note](https://docs.wpvip.com/databases/optimize-queries/retrieving-remote-data/#h-uncached-remote-requests):
> 
> > If for some reason an uncached remote request is required (e.g., to ping an external service during post publish) then the [WordPress HTTP API](https://developer.wordpress.org/plugins/http-api/) should be used rather than directly using cURL or another method.
> > Caution
> > For speed and performance reasons, uncached remote requests should never run on the front end of a site.
> 
> @dkotter Should Classifai requests ever be cached, they run in the admin and vary depending on the prompt so I am wondering if we should always use the WP Core API to avoid caching requests that are likely to be made once and once only. Thoughts?

Yeah, I went back and forth on this. Most external requests we make don't need to be cached (and probably shouldn't be cached) so there is really no benefit in using these VIP functions (other than making VIP happy). There are a few requests (like authentication checks) that are probably fine to keep but otherwise I think I'd prefer to default to not using the VIP helper functions. I can look to make that change here

peterwilsoncc:
Generally the changes look good but I wasn't able to get the environment running fully to check the text to speech items. I kept getting errors. Possibly PEBKAC.

I'll pick this up again for further testing tomorrow. 

Testing note: I'm using this as an mu-plugin to test https://gist.github.com/peterwilsoncc/413c915362963e46a5389ad836ddf493

dkotter:
@peterwilsoncc I've changed things up in this PR so we now default to _not_ using the VIP request functions unless we explicitly want them. And the only areas I've set to use them are authentication requests or requests to get information from the AI tool, like available models or voice options.

I think this makes a lot more sense as all other requests won't benefit from caching (and honestly shouldn't be cached at all) and very high likelihood of those requests taking longer than 5 seconds

includes/Classifai/Helpers.php
```diff
@@ -832,7 +832,7 @@ function is_local_path( string $resource_ref ): bool {
 /**
  * Safe generic HTTP wrapper compatible with WP VIP.
  *
- * - Use vip_safe_wp_remote_request() when available.
+ * - Use vip_safe_wp_remote_request() when available and wanted.
```
peterwilsoncc: I think this needs an update now that the default is to use the WP API.

```suggestion
 * By default this function uses the WordPress HTTP API for requests,
 * to use vip_safe_wp_remote_request() (if available) specify 
 * `$args['use_vip'] = true`.
 *
 * As a general guide, prompt requests should use the WordPress HTTP API;
 * administrative requests (API authentication, scope requests) should use the
 * WordPress VIP API if available.
```



🔢 Applies below too.

peterwilsoncc (dismissed):
This looks good to me with one suggestion inline (that applies in two places), feel free to adjust the copy as desired. 

Pre-approving as I am happy with the functional changes.

Testing notes:

* [Helper plugin](https://gist.github.com/peterwilsoncc/413c915362963e46a5389ad836ddf493) to include the VIP function.
* Used OpenAI for each test
* Image generation: Able to reproduce with a relatively complicated prompt.
* Image tag and image description generation: Unable to reproduce as they were quick requests.
* Speech related APIS: unable to get this going via ngrok locally sorry.

PR #1042, opened 2026-02-02
Move text-to-speech audio generation to Action Scheduler

rahulsprajapati:
### Description of the Change
Text-to-speech audio generation is no longer run synchronously during the REST “generate audio” request or when saving the post. It is offloaded to [Action Scheduler](https://actionscheduler.org/) so that:

- **REST and save don’t block**: The API and post save return immediately; synthesis runs in the background.
- **Timeouts are avoided**: Long or slow TTS requests no longer risk PHP/HTTP timeouts.
- **Duplicate work is avoided**: Before enqueueing, the code checks for an existing scheduled action for the same post via as_has_scheduled_action() and skips if one exists.
- **Fallback when Action Scheduler is missing**: If as_enqueue_async_action is not available, behavior falls back to the previous synchronous flow via a new generate_text_to_speech_audio( $post_id ) method.

### How to test the Change

1. Edit a new post (or one without tts audio already generated) 
2. Check the "Enable audio generation" toggle in ClassifAI panel
3. Save the post.
4. Confirm the ClassifAI panel/meta box shows “Audio generation is in progress.” and the request/save finishes without waiting for synthesis.
5. Wait for the scheduled action to run (or run the queue manually if your environment supports it).
    - You can go to "Tools" -> Action Scheduler page to look for this job.
    - Meanwhile Trigger save post again for the same post and confirm no duplicate job is created (e.g. check Action Scheduler list for that hook + post ID).
7. Reload the post and confirm the audio is present and the panel/meta box shows the normal controls and audio player.
8. Optionally run any existing ClassifAI tests and confirm they still pass.

### Credits
Props @rahulsprajapati, @dkotter 

### Checklist:


dkotter (commented):
Note I've pushed a couple commits to ensure [Action Scheduler is properly loaded](https://github.com/10up/classifai/pull/1042/changes/920d1fce08bd8f5b87a981f78efe2075cb46f4c1) when the TTS Feature is turned on and to ensure we [set the proper current user](https://github.com/10up/classifai/pull/1042/changes/5644150b370665186e4c19deb9ef9ab6bce4ddf9) so permission checks pass.

But in my testing, I never saw the message that lets you know processing is happening. I see that's supposed to be happening: `Audio generation is in progress`, but that never showed for me.

In addition, would be great if things updated dynamically with no need for a page refresh. Right now you have to refresh a handful of times and eventually the processing is done and you see the audio player but would be great if we could continually listen for processing to be done and update the UI as needed

dkotter:
> But in my testing, I never saw the message that lets you know processing is happening. I see that's supposed to be happening: `Audio generation is in progress`, but that never showed for me.

Quick follow up. Tested a bit more and found out it was a bug in how we register meta, in particular if used on post types other than post (I was testing on a page). I've pushed a [fix](https://github.com/10up/classifai/pull/1042/changes/8ba57bb3c0918c30bdc5e2ccba0a3a6d6fe4be18) for this and I am seeing the processing message now. That said, you still have to refresh to see the final result and it would be great if it could auto update the UI once that is done.

Also could use the same fix for the error message. I've pushed a small fix for that now as it was previously only showing during the save event, so if an error happens now, it will never render. This renders now but again, only if you refresh the page.

includes/Classifai/Features/TextToSpeech.php
```diff
@@ -497,25 +528,69 @@ public function save_post_metadata( int $post_id ) {
...
+			! isset( $_POST['classifai_synthesize_speech'] )
+			|| (
+				function_exists( 'as_has_scheduled_action' )
+				&& \as_has_scheduled_action( 'classifai_schedule_text_to_speech_job', $job_args, 'classifai' )
+			)
+		) {
+			return;
+		}
+
+		// We enqueue the async action to generate the audio, if available.
```
peterwilsoncc, line 554: Before scheduling the job, I think it would be best to do a cap check: `if ( ! current_user_can( 'edit_post', $post_id ) && ( ! defined( 'WP_CLI' ) || ! WP_CLI ) )`

By the time this part of the function is reached, I think it's been checked but it wold be good to have it here as a little protection for our future selves if it gets moved.

src/js/features/text-to-speech/index.js
```diff
@@ -141,6 +147,46 @@ const TextToSpeechPlugin = () => {
...
+			try {
+				const postId = wp.data
+					.select( 'core/editor' )
+					.getCurrentPostId();
+				const postType = wp.data
+					.select( 'core/editor' )
+					.getCurrentPostType();
+				wp.data
+					.dispatch( 'core' )
+					.invalidateResolution( 'getEntityRecord', [
```
peterwilsoncc, line 175: I couldn't figure it out but...

Is it possible to use the `_fields` parameter in the resulting rest request to only return the fields we need here: meta, custom endpoint data and post ID? I tried adding a `query` parameter without success but that may well have been my error.

src/js/features/text-to-speech/index.js
```diff
@@ -35,6 +36,11 @@ const TextToSpeechPlugin = () => {
 		)
 	);
 
+	const isTextToSpeechScheduled = useSelect( ( select ) => {
+		const meta = select( 'core/editor' ).getCurrentPostAttribute( 'meta' );
+		return ( meta && meta._classifai_text_to_speech_scheduled ) ?? false;
```
peterwilsoncc: Nit: I think optional chaining is slightly more readable but ignore if you disagree.

```suggestion
		return meta?._classifai_text_to_speech_scheduled ?? false;
```

includes/Classifai/Features/TextToSpeech.php
```diff
@@ -238,39 +248,42 @@ public function rest_handle_audio( \WP_Post $post, WP_REST_Request $request ) {
...
-					'_classifai_text_to_speech_error',
-					wp_json_encode(
-						[
-							'code'    => $results->get_error_code(),
-							'message' => $results->get_error_message(),
-						]
-					)
-				);
+			if ( function_exists( 'as_enqueue_async_action' ) ) {
+				\as_enqueue_async_action( 'classifai_schedule_text_to_speech_job', $job_args, 'classifai' );
```
peterwilsoncc, line 259: cap check per other comment.

peterwilsoncc (commented):
Just a couple of notes inline.

The main one is about including a cap check before scheduling the task given the new changes to the generate function.

When testing with the classic editor, I noticed that the status doesn't update once the audio is generated, it sticks with the in progress message.

Image

It's the classic editor so happy if that's intentional.

includes/Classifai/Features/TextToSpeech.php
dkotter: So we do already have a `current_user_can` check in this function a few lines above:

```
if ( ( defined( 'DOING_AUTOSAVE' ) && DOING_AUTOSAVE ) || ! current_user_can( 'edit_post', $post_id ) || 'revision' === get_post_type( $post_id ) ) {
	return;
}
```

Are you saying we should add an additional check here? Or move that check down to here?

dkotter:
> When testing with the classic editor, I noticed that the status doesn't update once the audio is generated, it sticks with the in progress message.
> 
> It's the classic editor so happy if that's intentional.

I didn't change how things work in the classic editor and I think that's fine for now. If we get reports that people are still using the classic editor here we can make this a nicer experience

includes/Classifai/Features/TextToSpeech.php
dkotter: Added: https://github.com/10up/classifai/pull/1042/changes/debce43a7ed4c46217e6a343bffd6a6a7b8775c2

src/js/features/text-to-speech/index.js
dkotter: I couldn't figure this out either (and had both Cursor and Claude give it a shot). The final response I got from AI is the block editor pulls data from the main entity so refreshing the cache based on the `_fields` parameter doesn't work (not sure if that's true 🤷🏻)

includes/Classifai/Features/TextToSpeech.php
peterwilsoncc: > So we do already have a `current_user_can` check in this function a few lines above:

Sorry Darin, we really do. ["In my defense, I have none."](https://www.youtube.com/watch?v=RNIqfTcFDYI)

peterwilsoncc (approved):
> I didn't change how things work in the classic editor and I think that's fine for now. If we get reports that people are still using the classic editor here we can make this a nicer experience

Cool, it's been about eight years.

Thanks for following up, these changes look good to me.

PR #1020, opened 2025-12-09
Feature: Add Support for Programatic Credentials

fabiankaegy:
### Description of the Change

This pull request introduces a centralized credentials management system for all provider integrations, enabling programmatic overrides and easier integration with external secret management solutions. The changes ensure that all provider classes consistently retrieve credentials through a new helper class and apply filtering hooks, improving security and flexibility.

**Centralized Credential Management:**

* Introduced a new `Credentials` helper class in `Classifai\Helpers` that centralizes credential retrieval and allows credentials to be filtered or overridden via the `classifai_provider_credentials` filter, supporting integration with external secret managers.

**Provider Refactoring to Use Centralized Credentials:**

[...]

PR #1043, opened 2026-02-03
Allow filtering of Provider credentials

### LINKED ISSUE #1019: Programmatic Credentials Support for External Secret Management
## Problem Statement

Currently, ClassifAI stores API credentials (API keys, endpoint URLs, etc.) in the WordPress database via options. This prevents integration with external secret management services like [Azure Key Vault](https://azure.microsoft.com/en-us/products/key-vault), AWS Secrets Manager, or HashiCorp Vault, where credentials should be fetched at runtime rather than stored in the database.

## Proposed Solution

Create a **unified credential filtering system** that works across all providers, following WordPress conventions. This involves:

1. A single generalized filter hook that all providers use
2. A centralized helper class for credential management
3. Provider-specific secondary filters for fine-grained control

### 1. Create a Centralized Credentials Helper Class

Create a new helper class at [`includes/Classifai/Helpers/Credentials.php`](includes/Classifai/Helpers/Credentials.php):

```php
<?php
namespace Classifai\Helpers;

/**
 * Centralized credential management for all providers.
 *
 * @since 3.x.0
 */
class Credentials {

    /**
     * Get credentials for a provider.
     *
     * @param string $provider_id The provider ID (e.g., 'azure_openai', 'openai_chatgpt').
     * @param string $feature_id  The feature ID (e.g., 'feature_title_generation').
     * @param array  $settings    The provider settings from the database.
     * @return array Filtered credentials array.
     */
    public static function get_credentials( string $provider_id, string $feature_id, array $settings ): array {
        /**
         * Filter provider credentials before making an API request.
         *
         * This is the primary hook for integrating external secret management
         * services like Azure Key Vault, AWS Secrets Manager, or HashiCorp Vault.
         *
         * @since 3.x.0
         * @hook classifai_provider_credentials
         *
         * @param array  $credentials The credentials array from settings.
         * @param string $provider_id The provider ID (e.g., 'azure_openai', 'openai_chatgpt').
         * @param string $feature_id  The feature ID making the request.
         *
         * @return array Filtered credentials array.
         */
        return apply_filters(
            'classifai_provider_credentials',
            $settings,
            $provider_id,
            $feature_id
        );
    }

    /**
     * Get a specific credential value.
     *
     * @param string $provider_id    The provider ID.
     * @param string $feature_id     The feature ID.
     * @param array  $settings       The provider settings.
     * @param string $credential_key The specific credential key (e.g., 'api_key').
     * @return mixed The credential value.
     */
    public static function get_credential( 
        string $provider_id, 
        string $feature_id, 
        array $settings, 
        string $credential_key 
    ) {
        $credentials = self::get_credentials( $provider_id, $feature_id, $settings );
        return $credentials[ $credential_key ] ?? '';
    }
}
```

### 2. Update the Base Provider Class

Modify [`includes/Classifai/Providers/Provider.php`](includes/Classifai/Providers/Provider.php) to add a method that providers can use:

```php
/**
 * Get filtered credentials for this provider.
 *
 * @param string $feature_id The feature ID making the request.
 * @return array The filtered credentials.
 */
protected function get_provider_credentials( string $feature_id ): array {
    $settings = $this->feature_instance->get_settings( static::ID );
    
    return \Classifai\Helpers\Credentials::get_credentials(
        static::ID,
        $feature_id,
        $settings
    );
}
```

### 3. Update All Providers to Use Centralized Credentials

Each provider will be updated to use the centralized credential retrieval. The credential keys vary by provider:

| Provider | Credential Keys |
|----------|-----------------|
| Azure OpenAI | `api_key`, `endpoint_url`, `deployment` |
| Azure Embeddings | `api_key`, `endpoint_url`, `deployment` |
| Azure Computer Vision | `api_key`, `endpoint_url` |
| Azure Speech | `api_key`, `endpoint_url` |
| OpenAI (all) | `api_key` |
| Google AI | `api_key` |
| IBM Watson | `api_key`, `endpoint_url` |
| AWS Polly | `access_key_id`, `secret_access_key` |
| ElevenLabs | `api_key` |
| xAI Grok | `api_key` |

### 4. Files to Modify

**New file:**

- `includes/Classifai/Helpers/Credentials.php` - New centralized helper

**Provider files to update:**

- `includes/Classifai/Providers/Provider.php` - Add base method
- `includes/Classifai/Providers/Azure/OpenAI.php`
- `includes/Classifai/Providers/Azure/Embeddings.php`
- `includes/Classifai/Providers/Azure/ComputerVision.php`
- `includes/Classifai/Providers/Azure/Speech.php`
- `includes/Classifai/Providers/Azure/Read.php`
- `includes/Classifai/Providers/Azure/SmartCropping.php`
- `includes/Classifai/Providers/OpenAI/APIRequest.php`
- `includes/Classifai/Providers/GoogleAI/APIRequest.php`
- `includes/Classifai/Providers/Watson/NLU.php`
- `includes/Classifai/Providers/AWS/AmazonPolly.php`
- `includes/Classifai/Providers/ElevenLabs/ElevenLabs.php`
- `includes/Classifai/Providers/XAI/APIRequest.php`

### Example 1: Azure Key Vault Integration (All Providers)

```php
add_filter( 'classifai_provider_credentials', function( $credentials, $provider_id, $feature_id ) {
    // Only override for Azure OpenAI providers
    if ( ! str_starts_with( $provider_id, 'azure_' ) ) {
        return $credentials;
    }
    
    $vault = new AzureKeyVaultClient( AZURE_VAULT_URL );
    
    return array_merge( $credentials, [
        'api_key' => $vault->get_secret( "classifai-{$provider_id}-api-key" ),
    ]);
}, 10, 3 );
```

### Example 2: Environment Variable Credentials

```php
add_filter( 'classifai_provider_credentials', function( $credentials, $provider_id, $feature_id ) {
    $env_key = strtoupper( "CLASSIFAI_{$provider_id}_API_KEY" );
    
    if ( defined( $env_key ) ) {
        $credentials['api_key'] = constant( $env_key );
    }
    
    return $credentials;
}, 10, 3 );
```

### Example 3: Feature-Specific Credentials

```php
add_filter( 'classifai_provider_credentials', function( $credentials, $provider_id, $feature_id ) {
    // Use different API key for content generation vs title generation
    if ( 'feature_content_generation' === $feature_id ) {
        $credentials['api_key'] = get_premium_api_key();
    }
    
    return $credentials;
}, 10, 3 );
```

## Considerations

- **Backward Compatibility**: Existing installations continue to work unchanged since the filter passes original values by default.
- **Caching**: Developers should implement caching when using external secret services to avoid excessive API calls.
- **Settings Page**: The authentication check on settings pages will also use filtered credentials, so developers can test their integration.
- **Error Handling**: Empty credentials will be caught by existing validation and display appropriate error messages.
- **Security**: Filtered credentials are never stored back to the database; they're only used for the current request.

dkotter:
### Description of the Change

This PR takes just the Provider credential filtering out of #1038 in order to make both of those PRs easier to review and test and hopefully get this piece merged in sooner.

The work here was heavily inspired by the work done in #1020 (and supersedes that PR) and was originally done in #1038 to ensure any changes we make around credentials management also work when we move Providers to be global.

Here's a high-level overview of changes here:

- Introduce a new CredentialResolver class that will pull just credentials out of our settings
- Add new methods on the Provider class that get all credentials for that Provider or a specific credential field. These all pass through new filters (`classifai_provider_credentials_{$provider_id}` and `classifai_provider_credentials`) making it easy for other to hook in and change those credentials
- More standardization on how API requests are made, making it easier to ensure all credentials pass through our new filter and setting us up for the future where we will integrate the new WP AI Client for these requests

Closes #1019 

### How to test the Change

Practically all Feature/Provider combinations are touched by this update though in general, this is all fairly minor changes. Testing to ensure Features still work as expected (that API requests work) is what is needed here.

### Changelog Entry

> Added - New methods on the Provider class that allow you to get all credentials or a specific credential, pulling the credentials from the larger Feature settings
> Added - New filters, `classifai_provider_credentials_{$provider_id}` and `classifai_provider_credentials`, that allows developers the ability to override AI credentials prior to those being used
> Changed - For filters that run when API requests are made, the Feature name, not the Feature option name is now passed as an argument. If you have code that relies on that argument, updates will be needed

### Credits

Props @fabiankaegy, @dkotter 

### Checklist:

### DIFF
--- a/includes/Classifai/Providers/GoogleAI/APIRequest.php
+++ b/includes/Classifai/Providers/GoogleAI/APIRequest.php
@@ -2,7 +2,6 @@
 
 namespace Classifai\Providers\GoogleAI;
 
-use Classifai\Features\Feature;
 use Classifai\Providers\Provider;
 use WP_Error;
 
@@ -24,14 +23,21 @@ use function Classifai\safe_wp_remote_post;
 class APIRequest {
 
 	/**
-	 * The Provider instance.
+	 * The Google AI API key.
 	 *
-	 * @var Provider
+	 * @var string
 	 */
-	public Provider $provider;
+	public $api_key;
 
 	/**
-	 * The Feature ID.
+	 * The Provider instance.
+	 *
+	 * @var Provider|null
+	 */
+	public ?Provider $provider = null;
+
+	/**
+	 * The Feature name.
 	 *
 	 * @var string
 	 */
@@ -47,13 +53,15 @@ class APIRequest {
 	/**
 	 * Google AI APIRequest constructor.
 	 *
-	 * @param Provider     $provider Provider instance.
-	 * @param Feature|null $feature Feature instance.
-	 * @param array        $settings Feature settings. Optional, useful when settings aren't saved yet.
+	 * @param string        $api_key Google AI API key.
+	 * @param string        $feature Feature name.
+	 * @param Provider|null $provider Provider instance.
+	 * @param array         $settings Feature settings. Optional, useful when settings aren't saved yet.
 	 */
-	public function __construct( Provider $provider, ?Feature $feature, array $settings = [] ) {
+	public function __construct( string $api_key = '', string $feature = '', ?Provider $provider = null, array $settings = [] ) {
+		$this->api_key  = $api_key;
+		$this->feature  = $feature;
 		$this->provider = $provider;
-		$this->feature  = $feature ? $feature::ID : '';
 		$this->settings = $settings;
 	}
 
@@ -247,6 +255,10 @@ class APIRequest {
 	 * @return string
 	 */
 	public function get_api_key() {
-		return $this->provider->get_credential( 'api_key', $this->settings ) ?? '';
+		if ( $this->provider ) {
+			return $this->provider->get_credential( 'api_key', $this->settings ) ?? '';
+		}
+
+		return $this->api_key ?? '';
 	}
 }
--- a/includes/Classifai/Providers/GoogleAI/APIRequest.php
+++ b/includes/Classifai/Providers/GoogleAI/APIRequest.php
@@ -2,7 +2,10 @@
 
 namespace Classifai\Providers\GoogleAI;
 
+use Classifai\Features\Feature;
+use Classifai\Providers\Provider;
 use WP_Error;
+
 use function Classifai\safe_wp_remote_get;
 use function Classifai\safe_wp_remote_post;
 
@@ -21,28 +24,37 @@ use function Classifai\safe_wp_remote_post;
 class APIRequest {
 
 	/**
-	 * The Google AI API key.
+	 * The Provider instance.
 	 *
-	 * @var string
+	 * @var Provider
 	 */
-	public $api_key;
+	public Provider $provider;
 
 	/**
-	 * The feature name.
+	 * The Feature instance.
 	 *
-	 * @var string
+	 * @var Feature
 	 */
 	public $feature;
 
 	/**
+	 * The Feature settings.
+	 *
+	 * @var array
+	 */
+	public $settings;
+
+	/**
 	 * Google AI APIRequest constructor.
 	 *
-	 * @param string $api_key Google AI API key.
-	 * @param string $feature Feature name.
+	 * @param Provider     $provider Provider instance.
+	 * @param Feature|null $feature Feature instance.
+	 * @param array        $settings Feature settings. Optional, useful when settings aren't saved yet.
 	 */
-	public function __construct( string $api_key = '', string $feature = '' ) {
-		$this->api_key = $api_key;
-		$this->feature = $feature;
+	public function __construct( Provider $provider, ?Feature $feature, array $settings = [] ) {
+		$this->provider = $provider;
+		$this->feature  = $feature ? $feature::ID : '';
+		$this->settings = $settings;
 	}
 
 	/**
@@ -235,6 +247,6 @@ class APIRequest {
 	 * @return string
 	 */
 	public function get_api_key() {
-		return $this->api_key;
+		return $this->provider->get_credential( 'api_key', $this->settings ) ?? '';
 	}
 }
--- a/includes/Classifai/Providers/GoogleAI/APIRequest.php
+++ b/includes/Classifai/Providers/GoogleAI/APIRequest.php
@@ -31,9 +31,9 @@ class APIRequest {
 	public Provider $provider;
 
 	/**
-	 * The Feature instance.
+	 * The Feature ID.
 	 *
-	 * @var Feature
+	 * @var string
 	 */
 	public $feature;
 
--- a/includes/Classifai/Providers/Provider.php
+++ b/includes/Classifai/Providers/Provider.php
@@ -189,7 +189,7 @@ abstract class Provider {
 		) {
 			$feature_provider_settings = $settings[ $provider_id ];
 		} else {
-			$feature_provider_settings = $this->feature_instance->get_settings( $provider_id ) ?? [];
+			$feature_provider_settings = $this->feature_instance ? $this->feature_instance->get_settings( $provider_id ) : [];
 		}
 
 		// Get our credentials.
--- a/includes/Classifai/Providers/Provider.php
+++ b/includes/Classifai/Providers/Provider.php
@@ -223,9 +223,9 @@ abstract class Provider {
 		 * @since x.x.x
 		 * @hook classifai_provider_credentials_{$provider_id}
 		 *
-		 * @param array $credentials The credentials for the Provider.
-		 * @param string|null $feature_id The ID of the Feature, or null if not set.
-		 * @param array $feature_provider_settings The Provider specific Feature settings.
+		 * @param array       $credentials               The credentials for the Provider.
+		 * @param string|null $feature_id                The ID of the Feature, or null if not set.
+		 * @param array       $feature_provider_settings The Provider specific Feature settings.
 		 * @return array The filtered credentials.
 		 */
 		$credentials = apply_filters(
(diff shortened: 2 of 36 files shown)


includes/Classifai/Providers/Provider.php
```diff
@@ -167,4 +169,60 @@ public function add_api_key_field( array $args = [] ) {
...
+		 * @hook classifai_provider_credentials
+		 *
+		 * @param array $credentials The credentials for the Provider.
+		 * @param string $provider_id The ID of the Provider.
+		 * @param string|null $feature_id The ID of the Feature, or null if not set.
+		 * @param array $feature_provider_settings The Provider specific Feature settings.
+		 * @return array The filtered credentials.
+		 */
+		return apply_filters(
+			'classifai_provider_credentials',
```
peterwilsoncc, line 251:
```suggestion
			"classifai_provider_credentials_{$provider_id}",
```

This will allow developers to use `classifai_provider_credentials_openai_chatgpt` etc rather than requiring they specify two arguments be passed each time they use the filter and include an `if...elseif`/`switch` block each time.

In the filter docs, it would be helpful to note each of the possible values which is done for most filters in core, see [`update_{$meta_type}_metadata`](https://developer.wordpress.org/reference/hooks/update_meta_type_metadata/) for example.

If you think a global filter would be helpful then we could retain this and add the dynamic version.

includes/Classifai/Providers/Provider.php
```diff
@@ -167,4 +169,60 @@ public function add_api_key_field( array $args = [] ) {
...
+		/**
+		 * Filter the credentials for the Provider.
+		 *
+		 * @since x.x.x
+		 * @hook classifai_provider_credentials
+		 *
+		 * @param array $credentials The credentials for the Provider.
+		 * @param string $provider_id The ID of the Provider.
+		 * @param string|null $feature_id The ID of the Feature, or null if not set.
+		 * @param array $feature_provider_settings The Provider specific Feature settings.
```
peterwilsoncc: Nitpick: align name and descriptions with spaces.

It's a little mixed but for the most part the docblocks are aligned in this manner.

includes/Classifai/Providers/GoogleAI/APIRequest.php
```diff
@@ -21,28 +24,37 @@
...
-	 * @param string $api_key Google AI API key.
-	 * @param string $feature Feature name.
+	 * @param Provider     $provider Provider instance.
+	 * @param Feature|null $feature Feature instance.
+	 * @param array        $settings Feature settings. Optional, useful when settings aren't saved yet.
 	 */
-	public function __construct( string $api_key = '', string $feature = '' ) {
-		$this->api_key = $api_key;
-		$this->feature = $feature;
+	public function __construct( Provider $provider, ?Feature $feature, array $settings = [] ) {
```
peterwilsoncc: Back-compat: The type changes here will affect any third party developers making use of the class, resulting in a fatal error.

* Do we need to handle this situation?
* Does this affect developers registering custom features/providers?

If it's possible to handle this without breaking back-compat it would be nice to do so.

:1234: This same change is made in a few of places.

peterwilsoncc (commented):
I've added a few notes inline.

Primarily, I think we could make the filter easier to use by including the provider name in the filter. Secondly, I'm wondering about back-compat with the type change.

includes/Classifai/Providers/Provider.php
dkotter: Makes sense, I've updated in https://github.com/10up/classifai/pull/1043/changes/b505db2fe4c2e7d85c4cb7e67c234622d3e75fcb. I've left the global filter as well as I could see someone wanting a single filter to hook into to change credentials for multiple Providers instead of having to hook into each specific filter. But I could be convinced to remove that if we don't think it provides value

includes/Classifai/Providers/Provider.php
dkotter: Updated in https://github.com/10up/classifai/pull/1043/changes/593bb5d95b30b725c61c7823a521ae09940abefc

includes/Classifai/Providers/GoogleAI/APIRequest.php
dkotter: So the plan here is to get rid of these individual APIRequest classes and instead use a single APIRequest class for everything, that uses the PHP AI Client when possible. This is likely coming in the next release so I wasn't overly worried about back-compat here as that change will also break things.

That said, not hard to maintain back-compat for now so I've updated the approach to do that: https://github.com/10up/classifai/pull/1043/changes/97d126c6e57bd18ce65d0badb8acbf33cc5ea92c and https://github.com/10up/classifai/pull/1043/changes/04d2bca34b6b5adf4a552d4cd7e7f84daeb58b0d

peterwilsoncc:
This looks good to me code wise and is generally working when testing.

As discussed in Slack, I was having trouble with `classifai_provider_credentials_ms_computer_vision` filtering. Running this code and enabling "Descriptive Text Generator" was allowing me to save the settings without a warning but the features list was showing the warning "This Feature is enabled but needs a Provider configured. Please click the Settings button to complete the setup."

```php
add_filter(
	'classifai_provider_credentials_ms_computer_vision',
	function( $creds, $feature_id, $feature_provider_settings ) {

		$creds = array(
			'endpoint_url' => 'https://eastus.api.cognitive.microsoft.com/',
			'api_key' => '', // Cognitive Services | Computer Vision from the internal docs
		);


		return $creds;
	},
	10,
	3
);
```

I added an xdebug break point within the filter and it wasn't hit when I saved the features settings.

After our discussion, I put some junk values in the API settings and the filter began working as expected

Screenshot 2026-02-06 at 11 03 48 am

dkotter:
> After our discussion, I put some junk values in the API settings and the filter began working as expected

Yeah, the issue here is a couple checks we run when most Provider settings are saved (varies slightly on a Provider by Provider basis). The first is a check like this:

```
if ( ! empty( $new_settings[ static::ID ]['endpoint_url'] ) && ! empty( $new_settings[ static::ID ]['api_key'] ) ) {
```

So if the credentials we want (in this case the endpoint URL and API key) aren't passed in, we don't try and authenticate. This used to make sense as there's no reason to authenticate if we don't have the needed credentials. But now that credentials don't need to be set in the UI and can instead be set by a filter, this check can cause problems.

Second, we have checks like this:

```
if ( ! ( $is_authenticated && $is_endpoint_same && $is_api_key_same ) ) {
```

In this case if the credentials we have saved aren't different from the credentials being submitted, we also don't try to authenticate again. This is a less likely issue to run into but could cause problems.

Both of these should be addressed but unless you think differently @peterwilsoncc, I'm inclined to merge this in and address these problems in a follow-up PR, allowing us to get this part of it released sooner (and hopefully get some real user testing on it to see if there are additional follow-ups needed).

peterwilsoncc (approved):
> I'm inclined to merge this in and address these problems in a follow-up PR, allowing us to get this part of it released sooner (and hopefully get some real user testing on it to see if there are additional follow-ups needed).

@dkotter I'm happy to take this approach. That will give us an opportunity to figure out the approach.

Otherwise the code looks good and is testing well in manual testing with various providers in the internal documentation.

PR #1047, opened 2026-02-09
Obfuscate credentials before rendering to the front-end

dkotter:
### Description of the Change

At the moment, we grab all settings when rendering our settings page. This includes any Provider credentials that have previously been saved. This allows us to populate those credential fields with saved values so if any additional saves happen, we don't lose those credentials.

The issue here is that means these credentials are rendered in plain text, both in the API response that gets our settings and in the text inputs themselves. This isn't a huge deal as only administrators have access to these setting pages but it still isn't ideal.

This PR fixes that by adding in an `CredentialObfuscator` class that will take all the settings we have saved and will then obfuscate just the private credential fields (like API keys). This only happens when we get our settings to display, not when we get the settings to use in API requests. We then check that we don't have obfuscated credentials when saving and if we do, we ignore those credentials (so we don't save obfuscated credentials).

### How to test the Change

1. Add credentials for one or more Providers and save those
2. Refresh the page and then inspect the source, ensuring the value of the text input is obfuscated
3. Can also refresh the page and inspect the network requests to ensure the response from our settings endpoint returns obfuscated credentials
4. Ensure that if you re-save a previously configured Provider that the obfuscated credentials aren't saved to the database and the Provider continues to work

### Changelog Entry

> Security - Add a `CredentialObfuscator` class that will obfuscate any private credentials before rendering those.

### Credits

Props @dkotter

### Checklist:

### DIFF
--- a/includes/Classifai/Providers/CredentialObfuscator.php
+++ b/includes/Classifai/Providers/CredentialObfuscator.php
@@ -56,7 +56,9 @@ class CredentialObfuscator {
 		$length = strlen( $value );
 
 		// If the value is too short, just return asterisks.
-		if ( $length <= self::VISIBLE_PREFIX_LENGTH ) {
+		if ( $length <= self::VISIBLE_PREFIX_LENGTH && $length <= self::MIN_ASTERISKS_TO_DETECT ) {
+			return str_repeat( '*', self::MIN_ASTERISKS_TO_DETECT );
+		} elseif ( $length <= self::VISIBLE_PREFIX_LENGTH ) {
 			return str_repeat( '*', $length );
 		}
 
--- a/includes/Classifai/Providers/CredentialObfuscator.php
+++ b/includes/Classifai/Providers/CredentialObfuscator.php
@@ -77,6 +77,11 @@ class CredentialObfuscator {
 		$prefix    = substr( $value, 0, self::VISIBLE_PREFIX_LENGTH );
 		$asterisks = str_repeat( '*', $length - self::VISIBLE_PREFIX_LENGTH );
 
+		// If we don't have enough asterisks, add more.
+		if ( strlen( $asterisks ) < self::MIN_ASTERISKS_TO_DETECT ) {
+			$asterisks = str_repeat( '*', self::MIN_ASTERISKS_TO_DETECT );
+		}
+
 		return $prefix . $asterisks;
 	}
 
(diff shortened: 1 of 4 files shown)


includes/Classifai/Providers/CredentialObfuscator.php
```diff
@@ -0,0 +1,256 @@
...
+
+		$length = strlen( $value );
+
+		// If the value is too short, just return asterisks.
+		if ( $length <= self::VISIBLE_PREFIX_LENGTH ) {
+			return str_repeat( '*', $length );
+		}
+
+		$prefix    = substr( $value, 0, self::VISIBLE_PREFIX_LENGTH );
+		$asterisks = str_repeat( '*', $length - self::VISIBLE_PREFIX_LENGTH );
```
peterwilsoncc, line 66: Paranoid nit pick. 

```suggestion
		$asterisks = str_repeat( '*', $length - self::VISIBLE_PREFIX_LENGTH );
		if ( strlen( $asterisks ) < self::MIN_ASTERISKS_TO_DETECT ) {
			$asterisks = str_repeat( '*', self::MIN_ASTERISKS_TO_DETECT );
		}
```

includes/Classifai/Providers/CredentialObfuscator.php
```diff
@@ -0,0 +1,256 @@
...
+	 * @var int
+	 */
+	const MIN_ASTERISKS_TO_DETECT = 3;
+
+	/**
+	 * Fields that should NOT be obfuscated (non-sensitive).
+	 *
+	 * @var array
+	 */
+	private static $non_sensitive_fields = [
```
peterwilsoncc: It could, maybe, possibly, be worth defining the sensitive fields in the `ProviderProfiles` class to allow for the circumstance where the same field ID is sensitive in one provider but not another, for example the endpoint_url contains the API key `example.com/{$api_key}/request`.

peterwilsoncc (commented):
A couple of suggestions inline.

They're both fairly paranoid so I'm happy to hear any push back.

Otherwise the code looks good and is testing well.

includes/Classifai/Providers/CredentialObfuscator.php
dkotter: Makes sense, updated in https://github.com/10up/classifai/pull/1047/changes/a2aa70d7f7d9a6e833d9f47073c7a46ed7c871ef

includes/Classifai/Providers/CredentialObfuscator.php
dkotter: Yeah, this is a better approach, updated in https://github.com/10up/classifai/pull/1047/changes/d0a84c5962defb3362a5d34fa72370677d0067f2

peterwilsoncc (approved):
I pushed one last paranoid length check in f16a21d80a0e80bbc3cbd3cf7b72accaf932370d.

This is testing well and the code looks good to merge.

PR #1048, opened 2026-02-09
Allow empty credentials when saving settings

dkotter:
### Description of the Change

We added the ability to easily filter Provider credentials in #1043. One [comment](https://github.com/10up/classifai/pull/1043#issuecomment-3857065765) on that PR flagged an issue with some Providers, where if you want to filter credentials you first need to add in dummy credentials in order for our pre-save checks to pass.

As a quick explanation, we had some checks in place where we wouldn't authenticate credentials if those credential fields were empty (no need to authenticate if we don't have an API key set, as an example). But in situations where you're overriding credentials, you shouldn't be forced to enter any credentials (dummy or not) in the admin first.

In investigating, found out that about half of our Providers worked fine but the others did not. In comparing those, the ones that work didn't check if credentials exist first before trying to authenticate and instead rely on the authentication check failing to alert the user that something is wrong. You could argue that's not the best user experience but it does more or less accomplish the same thing.

In order to keep this PR as simple as possible, I decided to follow that same approach on all the Providers that weren't working:

- AWS Polly
- Azure OpenAI
- ElevenLabs
- IBM Watson
- Azure TTS
- Azure AI Vision

All Providers should now have an `authenticate_credentials` method (most already did but not all) and we use that method as one of the first checks when sanitizing settings. If that check fails, we show an error message but we still save all settings (including any credential fields, even if blank).

### How to test the Change

In particular, test the Providers listed above:

1. Create a new plugin and hook into the new `classifai_provider_credentials`, overriding the credentials for the Providers you want to test
2. Install that plugin and ClassifAI into a new environment
3. Choose a Feature one of these Providers uses, configure any settings you may want (leaving credential fields blank) and then hit save. Ensure it authenticates correctly
4. Optional: test that the Feature works as expected
5. Disable the custom plugin and re-save a Feature, again leaving credential fields blank
6. Ensure an error message shows

### Changelog Entry

> Added - Ensure all Providers have an `authenticate_credentials` method that we can use to validate the credentials are valid
> Changed - Allow all settings to be saved, including credential fields, even if they are blank. This is mostly to support those using the new `classifai_provider_credentials` filter to override credentials

### Credits

Props @peterwilsoncc, @dkotter 

### Checklist:

### DIFF
--- a/includes/Classifai/Providers/AWS/AmazonPolly.php
+++ b/includes/Classifai/Providers/AWS/AmazonPolly.php
@@ -198,66 +198,99 @@ class AmazonPolly extends Provider {
 	 * @return array
 	 */
 	public function sanitize_settings( array $new_settings ): array {
-		$settings               = $this->feature_instance->get_settings();
-		$is_credentials_changed = false;
+		$settings      = $this->feature_instance->get_settings();
+		$authenticated = $this->authenticate_credentials( $new_settings );
 
 		$new_settings[ static::ID ]['authenticated'] = $settings[ static::ID ]['authenticated'];
 		$new_settings[ static::ID ]['voices']        = $settings[ static::ID ]['voices'];
 
-		if (
-			! empty( $new_settings[ static::ID ]['access_key_id'] ) &&
-			! empty( $new_settings[ static::ID ]['secret_access_key'] ) &&
-			! empty( $new_settings[ static::ID ]['aws_region'] )
-		) {
-			$new_access_key_id     = sanitize_text_field( $new_settings[ static::ID ]['access_key_id'] );
-			$new_secret_access_key = sanitize_text_field( $new_settings[ static::ID ]['secret_access_key'] );
-			$new_aws_region        = sanitize_text_field( $new_settings[ static::ID ]['aws_region'] );
-
-			if (
-				$new_access_key_id !== $settings[ static::ID ]['access_key_id'] ||
-				$new_secret_access_key !== $settings[ static::ID ]['secret_access_key'] ||
-				$new_aws_region !== $settings[ static::ID ]['aws_region']
-			) {
-				$is_credentials_changed = true;
-			}
-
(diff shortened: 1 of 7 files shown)


includes/Classifai/Providers/AWS/AmazonPolly.php
```diff
@@ -198,68 +198,98 @@ public function get_default_provider_settings(): array {
...
+			 * This filter is useful for E2E tests.
+			 *
+			 * @since 3.1.0
+			 * @hook classifai_aws_polly_pre_connect_to_service
+			 *
+			 * @param bool $pre The value of pre connect to service. Default false. A non-false value will short-circuit the describe voices request.
+			 *
+			 * @return bool|mixed The filtered value of connect to service.
+			 */
+			$pre = apply_filters( 'classifai_' . self::ID . '_pre_connect_to_service', false );
```
peterwilsoncc, line 270: Should we add this filter to the other providers?

Unrelated to this ticket/PR, not a merge consideration.

peterwilsoncc (approved):
This looks good to me.

For testing I used two approaches:

* created filters for various credentials
* Added a break point at the code point below to ensure the provider settings were saved https://github.com/10up/classifai/blob/119271a14af8a501357fcd6b5d70ab76f5fff66d/includes/Classifai/Features/Feature.php#L508

I've dropped a question inline as to whether we need a follow up ticket.

includes/Classifai/Providers/AWS/AmazonPolly.php
dkotter: So this was added specifically to help mock API responses for AWS Polly in our E2E tests. All other requests we mock using the `pre_http_request` filter but because we're using the AWS SDK here, these requests don't filter through the WordPress HTTP API and thus that filter never fires.

That said, I'm not opposed to more standard hooks here though I think worth thinking through once we look to standardize on using the PHP AI Client to make our requests (where possible)

PR #1053, opened 2026-02-12
Update how we run PHPUnit

dkotter:
### Description of the Change

Our [PHPUnit workflow has started to fail](https://github.com/10up/classifai/actions/runs/21961547268?pr=1052) and in investigating that, seems GitHub has [updated their default version of Docker](https://github.com/actions/runner-images/issues/13474) that is installed in their runners and that is conflicting with the tool we use to run our tests. That tool hasn't been updated since June 2024 so I'm not confident they'll push out a fix.

Also, there's a better way to run PHPUnit tests now that didn't exist/we didn't know about when we set this up (using `wp-env`) so this is a good opportunity to clean up that setup.

The following updates are made:

- We update our GitHub Action workflow to use the `wp-env` environment to run our PHPUnit tests (and run those on a wider matrix)
- Update our various config and bootstrap files as needed
- Fixed a few tests that started failing now that we're running tests (both unit and E2E) with debug turned on
- Rename our test directory to `Integration` (previously was `Classifai`)

Note most of this was copied from the approach we take on the [AI Experiments plugin](https://github.com/WordPress/ai).

### How to test the Change

Ensure the PHPUnit GitHub Action runs and passes on this PR

### Changelog Entry

> Developer - Update our PHPUnit workflow

### Credits

Props @dkotter

### Checklist:


.github/workflows/phpunit.yml
```diff
@@ -15,45 +15,77 @@ on:
...
-        os: [ ubuntu-latest ]
+        fail-fast: false
+        matrix:
+            php: [ '8.4', '8.3', '8.2', '8.1', '8.0', '7.4' ]
+            wp: [ latest, trunk ]
+            coverage: [ false ]
+            include:
+                - php: '8.4'
+                  wp: latest
+                  coverage: false
```
peterwilsoncc: This appears to be covered by the matrix, why is it needed?

tests/test-plugin/e2e-test-plugin.php
```diff
@@ -102,7 +102,7 @@ function classifai_test_mock_http_requests( $preempt, $parsed_args, $url ) {
 	} elseif ( strpos( $url, 'http://e2e-test-image-processing.test/computervision/imageanalysis:analyze?api-version=2024-02-01' ) !== false ) {
 		$response = file_get_contents( __DIR__ . '/image_analyze.json' );
 	} elseif ( strpos( $url, 'http://e2e-test-image-processing.test/vision/v3.2/generateThumbnail' ) !== false ) {
-		$response = file_get_contents( __DIR__ . '../classifai/assets/img/icon256x256.png' );
+		$response = file_get_contents( '/var/www/html/wp-content/plugins/classifai/assets/img/icon-256x256.png' );
```
peterwilsoncc: As an open source project, I think it would be good to keep this relative to `__DIR__` to account for people using different environments to run the tests on.

It's not a hill I will die on though.

composer.json
```diff
@@ -31,9 +31,11 @@
     "phpcompatibility/php-compatibility": "10.0.0-alpha2 as 9.99.99",
     "phpstan/extension-installer": "1.4.x-dev",
     "phpstan/phpstan-deprecation-rules": "2.0.3",
+    "phpunit/phpunit": "^8.5|^9.6",
```
peterwilsoncc: This is required by the Yoast polyfills, is there are reason to include it?

peterwilsoncc (commented):
A couple of questions and a minor quibble inline.

PHP Unit is generating the files and folders below that will need to be added the ignore file.

```
tests/_output/html/
tests/_output/php-coverage.xml
```

I'm able to get the tests running using both Chassis and WP-Env. On WP-Env I'm getting the message `Warning:       No code coverage driver available` so you may need to add something to the config file.

.github/workflows/phpunit.yml
dkotter: Originally I had `coverage: true` here, so the coverage report would be generated only on PHP 8.4 and WP latest. But our test coverage is currently so low that the report isn't super useful so I changed this to `false` and yeah, no need to have this now

tests/test-plugin/e2e-test-plugin.php
dkotter: The issue here is it seems this code never worked and we just never realized it. Now that our test environment runs with debug on, this started causing failures and I realized it was broken (and thus fixed it here). It seems the path traversal doesn't work (or I couldn't get it to work) so easy answer was this. But I can also just copy this image into our test directory and load it from there

composer.json
dkotter: Ah, didn't check the dependency graph but yeah, no need for this

.github/workflows/phpunit.yml
dkotter: Removed in https://github.com/10up/classifai/pull/1053/changes/f18f9a67ac297156b8a6cbe779ff20dc6d012fdf

tests/test-plugin/e2e-test-plugin.php
dkotter: Updated in https://github.com/10up/classifai/pull/1053/changes/2ddd7cb5abf6adb7959077d5d65a27a931d09000

composer.json
dkotter: Removed in https://github.com/10up/classifai/pull/1053/changes/bb61e3af2148471710ff91cacf303b813efc123a

dkotter:
> A couple of questions and a minor quibble inline.
> 
> PHP Unit is generating the files and folders below that will need to be added the ignore file.

Added in https://github.com/10up/classifai/pull/1053/changes/45bd2ce283dd7c0fe15633ceb2253aa5904003c5

> I'm able to get the tests running using both Chassis and WP-Env. On WP-Env I'm getting the message `Warning: No code coverage driver available` so you may need to add something to the config file.

So this can be fixed by running `npm run env start -- --xdebug=coverage` when we start our test environment. I'm currently not doing that as we're not using the coverage reports for anything (and this causes things to be slightly slower). But if you think having those coverage reports is useful, I can change this.

tests/test-plugin/e2e-test-plugin.php
peterwilsoncc: I didn't notice it yesterday but I now see the typo in the original code, it was `__DIR__' . '../'` rather than `__DIR__ ' . '/../'`

dkotter: So there were two typos, a missing slash (`__DIR__ . '../classifai` vs `__DIR__ . '/../classifai`) and a missing dash in the filename (`icon256x256.png` vs `icon-256x256.png`). I swear I fixed both of those yesterday and tests were still failing locally but trying again today things work so 🤷🏻 

In any case, this has been updated now to use the correct image path: https://github.com/10up/classifai/pull/1053/changes/20c50a17dc0281a387afb8422cc8aae0a937ecb2

(p.s. thanks for sticking with this one)

peterwilsoncc (approved):
> So this can be fixed by running npm run env start -- --xdebug=coverage when we start our test environment. I'm currently not doing that as we're not using the coverage reports for anything (and this causes things to be slightly slower). But if you think having those coverage reports is useful, I can change this.

That makes sense to me.

This looks good to merge.

PR #1054, opened 2026-02-13
Add OpenAI pricing/costs usage feature.

rahulsprajapati:
### Description of the Change

**Features**

- Site owners can monitor OpenAI spend and set soft (alert-only) and hard (disable-features) limits to avoid surprise bills.
- Optional email alerts and clear admin notices keep usage visible.
- Hard limit prevents further API calls until the admin re-enables or raises the limit.

This PR adds **OpenAI usage monitoring and spending controls** to ClassifAI:

- **OpenAI Pricing Controller** (`OpenAIPricingController.php`): Central admin controller that:
  - Fetches usage/cost data from the OpenAI Costs API (organization-level Admin API key) via a new `UsageCosts` class.
  - Caches usage (this month, year-to-date, all-time) in the `classifai_openai_pricing` option and refreshes it on a configurable interval (default 15 minutes) using Action Scheduler.
  - Implements **soft threshold**: when usage exceeds a configured amount for a scope (current month / year to date / all time), sends an email alert to configured addresses (once per period) and shows a dismissible warning notice in the admin.
  - Implements **hard threshold**: when usage exceeds a higher configured amount, sets the `classifai_openai_hard_limit_reached` option, which gates all OpenAI API requests in `APIRequest` (get/post/post_form) and shows an error notice; optionally sends an email. Users can re-enable by increasing the limit or disabling the hard threshold on the Pricing page.
  - Registers a **dashboard widget** (Tools → Dashboard) showing this month, YTD, and all-time usage with a “Configure alerts” link when pricing is enabled.

- **UsageCosts** (`UsageCosts.php`): New provider class that calls the OpenAI organization Costs API (`/v1/organization/costs`), with optional project ID filter, supports pagination, and provides `fetch_costs(start_ts, end_ts)` and `fetch_all_time_costs()` (from 2020).

- **Admin notices** (`Notifications.php`): New `render_openai_threshold_notice()` — shows a dismissible notice when soft (warning) or hard (error) threshold is exceeded, with link to **Tools → ClassifAI → Pricing**.

- **REST API** (`Settings.php`): New routes:
  - `GET classifai/v1/openai-usage` — returns cached usage for the dashboard/settings UI.
  - `GET classifai/v1/openai-pricing-settings` — returns pricing settings with API key obfuscated and `hard_limit_reached` flag.
  - `POST classifai/v1/openai-pricing-settings` — updates pricing settings; supports `force_refresh: true` to trigger an immediate usage refresh.

[...]

PR #1060, opened 2026-02-25
Add Usage Tracking feature with OpenAI support

### LINKED ISSUE #122: Display available credits/quotas from service providers
In chatting with @dingman, one item that would be nice to expose if we can get it via service provider APIs is the current usage / credits / quotas a user has with their account.

- [ ] on the Language and Image Processing settings pages, display the available credits/quotas from enabled services
- [ ] Azure Computer Vision would ideally show `Pricing Tier`, `Subscription`, `Quota Info` (`Total Calls`, `Remaining Calls`, `Rest Period`), and `Total Cost`
- [ ] Watson NLU would ideally show `NLU Items Available`, `NLU Items Used`, `API Rate Limit`, and `Total Cost`

**Designs**
- [ ] designs needed

n/a

**Additional context**
NOTE: A NLU item is based on the number of data units enriched and the number of enrichment features applied. A data unit is 10,000 characters or less. For example: extracting Entities and Sentiment from 15,000 characters of text is (2 Data Units * 2 Enrichment Features) = 4 NLU Items.

rahulsprajapati:
### Description of the Change

This PR adds **OpenAI Usage Tracking** to ClassifAI: the ability to monitor usage/costs from the OpenAI API and enforce soft (alert-only) and hard (disable features) spending limits.

**Design overview:**

- **New service & feature:** A new `UsageTracking` service and `OpenAIUsage` feature are introduced, with provider `OpenAI\UsageTracking` that calls the [OpenAI organization costs API](https://developers.openai.com/api/reference/resources/organization/subresources/audit_logs/methods/get_costs).
- **Usage data:** Usage (MTD, YTD, all time) is fetched and cached with a configurable refresh interval (default 15 minutes) via Action Scheduler. A "Force refresh" action is available via REST and in the UI.
- **Thresholds:** 
    **Soft threshold** — when usage exceeds the configured amount for the chosen scope, a dismissible admin notice is shown and an optional email can be sent (once per period). 
    **Hard threshold** — when exceeded, an option is set, all OpenAI API requests are blocked in `APIRequest::get/post/post_form` (returning a `WP_Error`), and a non-dismissible error notice is shown with a link to re-enable from the pricing/usage page. Email can be sent once per period.
- **Scopes:** Both thresholds support scope: current month, year to date, or all time.
- **UI:** 
     - Settings under Tools → ClassifAI → Usage tracking: Admin API key, optional project ID, refresh interval, force refresh button, and soft/hard threshold toggles with amount, scope, and comma-separated emails. 
     - A dashboard widget shows MTD/YTD/all-time usage and links to configure alerts and force refresh.
- **Integration:** 
     - `Plugin.php` registers the Usage Tracking service and OpenAI Usage feature and adds `openai_usage_tracking` support to load the Action Scheduler. 
     - `APIRequest` checks `is_request_allowed()` (which respects the hard-limit option and the `classifai_can_make_request` filter) before any `get`/`post`/`post_form` call for OpenAI providers.
     - `Notifications.php` gains `render_openai_threshold_notice()` so admins see soft/hard threshold warnings in the admin.

**Verification:** 
- Enable the feature, set an OpenAI org Admin API key (and optional project ID), save. 
- Confirm usage appears in the dashboard widget and in the usage-tracking settings. 
- Set a soft threshold (e.g. $1), trigger a refresh so it's exceeded: dismissible notice and (if emails set) one email per period. 
- Set a hard threshold and exceed it: OpenAI features are blocked, non-dismissible notice appears, and (if emails set) one email per period. 
- Re-enable from the linked pricing/usage page and confirm requests work again. 
- Confirm "Force refresh" in the UI and on the dashboard triggers a refresh and that the REST endpoint `/classifai/v1/openai-usage/force-refresh` is only allowed when the feature is enabled.

**Benefits:** Visibility into OpenAI spend (MTD/YTD/all time), alerts and optional automatic disabling to avoid cost overruns, and one email per period per threshold to avoid spam.

**Possible drawback / note:** Uses an organization-level Admin API key; docs should clarify this is required and separate from project keys.

Closes https://github.com/10up/classifai/issues/122#issuecomment-3761541080

### How to test the Change

1. Install/activate ClassifAI and ensure the new settings UI is used (legacy settings off).
2. Go to **Tools → ClassifAI** and open the **Usage tracking** section for the OpenAI Usage feature. 
3. Verify you see "OpenAI usage tracking" section. Click on toggle button for this feature to enable it. 
4. Click on "Settings" button in "OpenAI usage tracking" section itself, and verify you see "OpenAI usage tracking Settings" section with following fields:
     - Enable Feature
     - "Select a provider" with "OpenAI Usage Tracking" provider 
     - Admin API Key
     - OpenAI Project ID
     - Refresh interval (minutes)
     - Force refresh data
     - "Soft threshold (alert)" and "Hard threshold (alert)" section with toggle to enable alert and scope, email fields. 
5. Enter an OpenAI **organization Admin API key** (sk-admin-…) to "Admin API Key". Optionally set **OpenAI Project ID** to restrict to one project.
6.  Save Settings
7. Wait for 15min cron to run (or whatever Refresh interval you set) to update the data. OR click on "Force refresh data" button and wait for it to finish refreshing data. (You will see "Force refresh data" button disabled and processing UI until it complete the data updates).
8. Verify **Usage display:**: Open the **Dashboard** and confirm the "OpenAI usage (ClassifAI)" widget shows This month / Year to date / All time (and "Last updated" or "Updating…"). 
9. **Soft threshold:** 
      - Enable **Soft threshold (alert)** and set a low amount (e.g. $1), scope (e.g. Current month), and at least one email. Save. 
      - Force refresh so current usage exceeds the amount. 
      - Confirm a dismissible **warning** notice appears and that the configured email(s) receive one alert for that period. 
      - Dismiss the notice and confirm it does not reappear until the next period (or after clearing the dismissed meta).
11. **Hard threshold:** 
      - Enable **Hard threshold (disable features)** and set an amount already exceeded (or very low), with scope and emails. Save
      -  Force refresh so the hard limit is exceeded. 
      - Confirm a red, non-dismissible **error** notice appears and that OpenAI-powered features (e.g. ChatGPT, image generation) are blocked (e.g. requests return an error about the hard limit). 
      - Use the link in the notice to open the pricing/usage page and re-enable; confirm the notice can be cleared and OpenAI features work again. 
      - Confirm one hard-limit email per period when emails are set.
13. **Regression:** With usage tracking disabled, confirm existing OpenAI features still work and no usage notices appear. Confirm the new REST route and dashboard script only load when the feature is enabled.

### Changelog Entry

> Added - OpenAI usage tracking: monitor usage/costs via OpenAI org API, optional soft (notice and email alert) and hard (notice, email and disable OpenAI features) spending thresholds with configurable scope (current month, YTD, all time), dashboard widget, and force refresh.

### Credits

Props @rahulsprajapati 

### Checklist:

### DIFF
--- a/includes/Classifai/Admin/Notifications.php
+++ b/includes/Classifai/Admin/Notifications.php
@@ -449,16 +449,16 @@ EOD;
 			$classes[] = 'notice-error';
 			$message   = sprintf(
 				/* translators: 1: amount with currency, 2: link to settings */
-				__( 'AI Features are currently disabled due to exceeding your hard limit of %1$s for this period. <a href="%2$s">Re-enable it from the pricing page</a>.', 'classifai' ),
-				esc_html( number_format_i18n( $threshold, 2 ) . ' ' . ( $usage_data['currency'] ?? 'USD' ) ),
+				__( 'AI Features are currently disabled due to exceeding your hard threshold of $%1$s for this period. <a href="%2$s">Re-enable it from the pricing page</a>.', 'classifai' ),
+				esc_html( number_format_i18n( $threshold, 2 ) . ' ' . strtoupper( $usage_data['currency'] ?? 'USD' ) ),
 				esc_url( $settings_url )
 			);
 		} else {
 			$classes[] = 'notice-warning';
 			$message   = sprintf(
 				/* translators: 1: amount with currency, 2: link to settings */
-				__( 'AI usage has exceeded your soft limit of %1$s for this period. <a href="%2$s">Configure alerts</a>.', 'classifai' ),
-				esc_html( number_format_i18n( $threshold, 2 ) . ' ' . ( $usage_data['currency'] ?? 'USD' ) ),
+				__( 'AI usage has exceeded your soft threshold of $%1$s for this period. <a href="%2$s">Configure alerts</a>.', 'classifai' ),
+				esc_html( number_format_i18n( $threshold, 2 ) . ' ' . strtoupper( $usage_data['currency'] ?? 'USD' ) ),
 				esc_url( $settings_url )
 			);
 		}
--- a/includes/Classifai/Admin/Notifications.php
+++ b/includes/Classifai/Admin/Notifications.php
@@ -412,7 +412,8 @@ EOD;
 		}
 
 		// Don't show the notice if the provider is not OpenAI Usage Tracking.
-		$provider = $feature_instance->get_settings( 'provider' );
+		$provider = $feature_instance->get_feature_provider_instance();
+
 		if ( ! $provider instanceof UsageTrackingProvider ) {
 			return;
 		}
--- a/includes/Classifai/Admin/Notifications.php
+++ b/includes/Classifai/Admin/Notifications.php
@@ -5,7 +5,8 @@ namespace Classifai\Admin;
 
 use Classifai\Features\DescriptiveTextGenerator;
 use Classifai\Features\Classification;
-use Classifai\Features\OpenAIUsage;
+use Classifai\Features\APIUsageTracking;
+use Classifai\Providers\UsageTrackingProvider;
 use Classifai\Providers\OpenAI\UsageTracking as OpenAIUsageTracking;
 use function Classifai\should_use_legacy_settings_panel;
 
@@ -53,7 +54,7 @@ class Notifications {
 		$this->v3_migration_completed_notice();
 		$this->render_embeddings_notice();
 		$this->render_legacy_settings_deprecation_notice();
-		$this->render_openai_threshold_notice();
+		$this->render_api_threshold_notice();
 		$this->render_notices();
 	}
 
@@ -394,16 +395,16 @@ EOD;
 	}
 
 	/**
-	 * Render a notice when OpenAI usage exceeds the soft or hard threshold limit.
+	 * Render a notice when AI usage exceeds the soft or hard threshold limit.
 	 */
-	public function render_openai_threshold_notice() {
+	public function render_api_threshold_notice() {
 
 		// Ensure the feature exists.
-		if ( ! class_exists( 'Classifai\Features\OpenAIUsage' ) ) {
+		if ( ! class_exists( 'Classifai\Features\APIUsageTracking' ) ) {
 			return;
 		}
 
-		$feature_instance = new OpenAIUsage();
+		$feature_instance = new APIUsageTracking();
 
 		// Don't show the notice if the feature is not enabled.
 		if ( ! $feature_instance->is_feature_enabled() ) {
@@ -412,11 +413,11 @@ EOD;
 
 		// Don't show the notice if the provider is not OpenAI Usage Tracking.
 		$provider = $feature_instance->get_settings( 'provider' );
-		if ( OpenAIUsageTracking::ID !== $provider ) {
+		if ( ! $provider instanceof UsageTrackingProvider ) {
 			return;
 		}
 
-		$settings = $feature_instance->get_settings( OpenAIUsageTracking::ID );
+		$settings = $feature_instance->get_settings( $provider::ID );
 
 		if ( empty( $settings['soft_threshold_enabled'] ) && empty( $settings['hard_threshold_enabled'] ) ) {
 			return;
@@ -438,14 +439,14 @@ EOD;
 			return;
 		}
 
-		$key = 'openai_threshold_reached';
+		$key = 'api_threshold_reached';
 
 		// Don't show the notice if the user has already dismissed it for soft threshold. Always show for hard threshold.
 		if ( ! $hard_threshold_reached && get_user_meta( get_current_user_id(), "classifai_dismissed_{$key}", true ) ) {
 			return;
 		}
 
-		$settings_url = admin_url( 'tools.php?page=classifai#/usage_tracking/feature_openai_usage' );
+		$settings_url = admin_url( 'tools.php?page=classifai#/usage_tracking/api_usage_tracking' );
 
 		$classes = [
 			'notice',
@@ -455,7 +456,7 @@ EOD;
 			$classes[] = 'notice-error';
 			$message   = sprintf(
 				/* translators: 1: amount with currency, 2: link to settings */
-				__( 'OpenAI Features are currently disabled due to exceeded your hard limit of %1$s for this period. <a href="%2$s">Re-enable it from the pricing page</a>.', 'classifai' ),
+				__( 'AI Features are currently disabled due to exceeded your hard limit of %1$s for this period. <a href="%2$s">Re-enable it from the pricing page</a>.', 'classifai' ),
 				esc_html( number_format_i18n( $threshold, 2 ) . ' ' . ( $usage_data['currency'] ?? 'USD' ) ),
 				esc_url( $settings_url )
 			);
@@ -465,7 +466,7 @@ EOD;
 			$classes[] = 'notice-warning';
 			$message   = sprintf(
 				/* translators: 1: amount with currency, 2: link to settings */
-				__( 'OpenAI usage has exceeded your soft limit of %1$s for this period. <a href="%2$s">Configure alerts</a>.', 'classifai' ),
+				__( 'AI usage has exceeded your soft limit of %1$s for this period. <a href="%2$s">Configure alerts</a>.', 'classifai' ),
 				esc_html( number_format_i18n( $threshold, 2 ) . ' ' . ( $usage_data['currency'] ?? 'USD' ) ),
 				esc_url( $settings_url )
 			);
--- a/includes/Classifai/Admin/Notifications.php
+++ b/includes/Classifai/Admin/Notifications.php
@@ -456,7 +456,7 @@ EOD;
 			$classes[] = 'notice-error';
 			$message   = sprintf(
 				/* translators: 1: amount with currency, 2: link to settings */
-				__( 'AI Features are currently disabled due to exceeded your hard limit of %1$s for this period. <a href="%2$s">Re-enable it from the pricing page</a>.', 'classifai' ),
+				__( 'AI Features are currently disabled due to exceeding your hard limit of %1$s for this period. <a href="%2$s">Re-enable it from the pricing page</a>.', 'classifai' ),
 				esc_html( number_format_i18n( $threshold, 2 ) . ' ' . ( $usage_data['currency'] ?? 'USD' ) ),
 				esc_url( $settings_url )
 			);
--- a/includes/Classifai/Admin/Notifications.php
+++ b/includes/Classifai/Admin/Notifications.php
@@ -410,7 +410,7 @@ EOD;
 			return;
 		}
 
-		// Don't show the notice if the provider is not OpenAI Usage Tracking.
+		// Don't show the notice if the provider is not UsageTrackingProvider.
 		$provider = $feature_instance->get_feature_provider_instance();
 
 		if ( ! $provider instanceof UsageTrackingProvider ) {
--- a/includes/Classifai/Admin/Notifications.php
+++ b/includes/Classifai/Admin/Notifications.php
@@ -5,6 +5,8 @@ namespace Classifai\Admin;
 
 use Classifai\Features\DescriptiveTextGenerator;
 use Classifai\Features\Classification;
+use Classifai\Features\OpenAIUsage;
+use Classifai\Providers\OpenAI\UsageTracking as OpenAIUsageTracking;
 use function Classifai\should_use_legacy_settings_panel;
 
 if ( ! defined( 'ABSPATH' ) ) {
@@ -51,6 +53,7 @@ class Notifications {
 		$this->v3_migration_completed_notice();
 		$this->render_embeddings_notice();
 		$this->render_legacy_settings_deprecation_notice();
+		$this->render_openai_threshold_notice();
 		$this->render_notices();
 	}
 
@@ -388,6 +391,98 @@ EOD;
 		$notice_id = sanitize_text_field( wp_unslash( $_POST['notice_id'] ) );
 
 		update_user_meta( get_current_user_id(), "classifai_dismissed_{$notice_id}", true );
+	}
+
+	/**
+	 * Renders a dismissible warning when OpenAI usage exceeds the hard threshold limit.
+	 */
+	public function render_openai_threshold_notice() {
+
+		// Ensure the feature exists.
+		if ( ! class_exists( 'Classifai\Features\OpenAIUsage' ) ) {
+			return;
+		}
+
+		$feature_instance = new OpenAIUsage();
+
+		// Don't show the notice if the feature is not enabled.
+		if ( ! $feature_instance->is_feature_enabled() ) {
+			return;
+		}
+
+		// Don't show the notice if the provider is not OpenAI Usage Tracking.
+		$provider = $feature_instance->get_settings( 'provider' );
+		if ( OpenAIUsageTracking::ID !== $provider ) {
+			return;
+		}
+
+		$key = 'openai_threshold_reached';
+
+		// Don't show the notice if the user has already dismissed it.
+		if ( get_user_meta( get_current_user_id(), "classifai_dismissed_{$key}", true ) ) {
+			return;
+		}
+
+		$settings = $feature_instance->get_settings( OpenAIUsageTracking::ID );
+
+		if ( empty( $settings['soft_threshold_enabled'] ) && empty( $settings['hard_threshold_enabled'] ) ) {
+			return;
+		}
+
+		$hard_threshold_reached = get_option( $feature_instance::HARD_LIMIT_REACHED_KEY, false );
+
+		if ( $hard_threshold_reached ) {
+			$scope = isset( $settings['hard_threshold_scope'] ) ? $settings['hard_threshold_scope'] : 'current_month';
+		} else {
+			$scope = isset( $settings['soft_threshold_scope'] ) ? $settings['soft_threshold_scope'] : 'current_month';
+		}
+
+		$usage_data = $feature_instance->get_usage_data();
+		$amount     = $feature_instance->get_amount_for_scope( $usage_data, $scope );
+		$threshold  = $hard_threshold_reached ? (float) $settings['hard_threshold_amount'] : (float) $settings['soft_threshold_amount'];
+
+		if ( $amount < $threshold ) {
+			return;
+		}
+
+		$key = 'openai_threshold_reached';
+
+		if ( get_user_meta( get_current_user_id(), "classifai_dismissed_{$key}", true ) ) {
+			return;
+		}
+
+		$settings_url = admin_url( 'tools.php?page=classifai#/usage_tracking/feature_openai_usage' );
+
+		$classes = [
+			'notice',
+			'is-dismissible',
+			'classifai-dismissible-notice',
+		];
+
+		if ( $hard_threshold_reached ) {
+			$classes[] = 'notice-error';
+			$message   = sprintf(
+				/* translators: 1: amount with currency, 2: link to settings */
+				__( 'OpenAI features are currently disabled due to exceeded your hard limit of %1$s for this period. <a href="%2$s">Re-enable it from the pricing page</a>.', 'classifai' ),
+				esc_html( number_format_i18n( $threshold, 2 ) . ' ' . ( $usage['currency'] ?? 'USD' ) ),
+				esc_url( $settings_url )
+			);
+		} else {
+			$classes[] = 'notice-warning';
+			$message   = sprintf(
+				/* translators: 1: amount with currency, 2: link to settings */
+				__( 'OpenAI usage has exceeded your soft limit of %1$s for this period. <a href="%2$s">Configure alerts</a>.', 'classifai' ),
+				esc_html( number_format_i18n( $threshold, 2 ) . ' ' . ( $usage['currency'] ?? 'USD' ) ),
+				esc_url( $settings_url )
+			);
+		}
+		?>
+		<div class="<?php echo esc_attr( implode( ' ', $classes ) ); ?>" data-notice="<?php echo esc_attr( $key ); ?>">
+			<p>
+				<?php echo wp_kses_post( $message ); ?>
+			</p>
+		</div>
+		<?php
 	}
 
 	/**
--- a/includes/Classifai/Admin/Notifications.php
+++ b/includes/Classifai/Admin/Notifications.php
@@ -447,7 +447,8 @@ EOD;
 
 		$key = 'openai_threshold_reached';
 
-		if ( get_user_meta( get_current_user_id(), "classifai_dismissed_{$key}", true ) ) {
+		// Don't show the notice if the user has already dismissed it for soft threshold. Always show for hard threshold.
+		if ( ! $hard_threshold_reached && get_user_meta( get_current_user_id(), "classifai_dismissed_{$key}", true ) ) {
 			return;
 		}
 
@@ -455,19 +456,19 @@ EOD;
 
 		$classes = [
 			'notice',
-			'is-dismissible',
-			'classifai-dismissible-notice',
 		];
 
 		if ( $hard_threshold_reached ) {
 			$classes[] = 'notice-error';
 			$message   = sprintf(
 				/* translators: 1: amount with currency, 2: link to settings */
-				__( 'OpenAI features are currently disabled due to exceeded your hard limit of %1$s for this period. <a href="%2$s">Re-enable it from the pricing page</a>.', 'classifai' ),
+				__( 'OpenAI Features are currently disabled due to exceeded your hard limit of %1$s for this period. <a href="%2$s">Re-enable it from the pricing page</a>.', 'classifai' ),
 				esc_html( number_format_i18n( $threshold, 2 ) . ' ' . ( $usage['currency'] ?? 'USD' ) ),
 				esc_url( $settings_url )
 			);
 		} else {
+			$classes[] = 'is-dismissible';
+			$classes[] = 'classifai-dismissible-notice';
 			$classes[] = 'notice-warning';
 			$message   = sprintf(
 				/* translators: 1: amount with currency, 2: link to settings */
--- a/includes/Classifai/Admin/Notifications.php
+++ b/includes/Classifai/Admin/Notifications.php
@@ -394,7 +394,7 @@ EOD;
 	}
 
 	/**
-	 * Renders a dismissible warning when OpenAI usage exceeds the hard threshold limit.
+	 * Render a notice when OpenAI usage exceeds the soft or hard threshold limit.
 	 */
 	public function render_openai_threshold_notice() {
 
@@ -413,13 +413,6 @@ EOD;
 		// Don't show the notice if the provider is not OpenAI Usage Tracking.
 		$provider = $feature_instance->get_settings( 'provider' );
 		if ( OpenAIUsageTracking::ID !== $provider ) {
-			return;
-		}
-
-		$key = 'openai_threshold_reached';
-
-		// Don't show the notice if the user has already dismissed it.
-		if ( get_user_meta( get_current_user_id(), "classifai_dismissed_{$key}", true ) ) {
 			return;
 		}
 
--- a/includes/Classifai/Admin/Notifications.php
+++ b/includes/Classifai/Admin/Notifications.php
@@ -463,7 +463,7 @@ EOD;
 			$message   = sprintf(
 				/* translators: 1: amount with currency, 2: link to settings */
 				__( 'OpenAI Features are currently disabled due to exceeded your hard limit of %1$s for this period. <a href="%2$s">Re-enable it from the pricing page</a>.', 'classifai' ),
-				esc_html( number_format_i18n( $threshold, 2 ) . ' ' . ( $usage['currency'] ?? 'USD' ) ),
+				esc_html( number_format_i18n( $threshold, 2 ) . ' ' . ( $usage_data['currency'] ?? 'USD' ) ),
 				esc_url( $settings_url )
 			);
 		} else {
@@ -473,7 +473,7 @@ EOD;
 			$message   = sprintf(
 				/* translators: 1: amount with currency, 2: link to settings */
 				__( 'OpenAI usage has exceeded your soft limit of %1$s for this period. <a href="%2$s">Configure alerts</a>.', 'classifai' ),
-				esc_html( number_format_i18n( $threshold, 2 ) . ' ' . ( $usage['currency'] ?? 'USD' ) ),
+				esc_html( number_format_i18n( $threshold, 2 ) . ' ' . ( $usage_data['currency'] ?? 'USD' ) ),
 				esc_url( $settings_url )
 			);
 		}
--- a/includes/Classifai/Admin/Notifications.php
+++ b/includes/Classifai/Admin/Notifications.php
@@ -439,16 +439,9 @@ EOD;
 			return;
 		}
 
-		$key = 'api_threshold_reached';
-
-		// Don't show the notice if the user has already dismissed it for soft threshold. Always show for hard threshold.
-		if ( ! $hard_threshold_reached && get_user_meta( get_current_user_id(), "classifai_dismissed_{$key}", true ) ) {
-			return;
-		}
-
+		$key          = 'api_threshold_reached';
 		$settings_url = admin_url( 'tools.php?page=classifai#/usage_tracking/api_usage_tracking' );
-
-		$classes = [
+		$classes      = [
 			'notice',
 		];
 
@@ -461,8 +454,6 @@ EOD;
 				esc_url( $settings_url )
 			);
 		} else {
-			$classes[] = 'is-dismissible';
-			$classes[] = 'classifai-dismissible-notice';
 			$classes[] = 'notice-warning';
 			$message   = sprintf(
 				/* translators: 1: amount with currency, 2: link to settings */
--- a/includes/Classifai/Admin/Notifications.php
+++ b/includes/Classifai/Admin/Notifications.php
@@ -7,7 +7,6 @@ use Classifai\Features\DescriptiveTextGenerator;
 use Classifai\Features\Classification;
 use Classifai\Features\APIUsageTracking;
 use Classifai\Providers\UsageTrackingProvider;
-use Classifai\Providers\OpenAI\UsageTracking as OpenAIUsageTracking;
 use function Classifai\should_use_legacy_settings_panel;
 
 if ( ! defined( 'ABSPATH' ) ) {
--- a/includes/Classifai/Features/APIUsageTracking.php
+++ b/includes/Classifai/Features/APIUsageTracking.php
@@ -156,8 +156,48 @@ class APIUsageTracking extends Feature {
 
 		add_action( self::FORCE_CRON_HOOK, [ $this, 'run_usage_force_refresh' ] );
 		add_action( self::CRON_HOOK, [ $this, 'run_usage_refresh' ] );
+
+		add_filter( 'classifai_pre_fetch_feature_response', [ $this, 'pre_fetch_feature_response' ], 10, 2 );
 	}
 
+	/**
+	 * Pre-fetch the feature response to check if the feature is allowed to run based on the usage tracking data.
+	 *
+	 * @param mixed                         $response Response to return.
+	 * @param \Classifai\Providers\Provider $provider_instance provider used.
+	 *
+	 * @return mixed Response to return.
+	 */
+	public function pre_fetch_feature_response( $response, $provider_instance ) {
+		$usage_tracking_provider = $this->get_feature_provider_instance();
+		$provider_ids            = [];
+
+		if ( ! empty( $usage_tracking_provider ) && $usage_tracking_provider instanceof UsageTrackingProvider ) {
+			$provider_ids = $usage_tracking_provider->get_provider_ids();
+		}
+
+		if (
+			empty( $provider_ids )
+			|| empty( $provider_instance::ID )
+			|| ! in_array( $provider_instance::ID, $provider_ids, true )
+		) {
+			return $response;
+		}
+
+		$limit_reached = get_option( self::HARD_LIMIT_REACHED_KEY, false );
+
+		if ( $limit_reached ) {
+			return new WP_Error(
+				'classifai_hard_limit_reached',
+				__( 'Usage has reached the configured hard limit. Re-enable in ClassifAI / Pricing / Usage settings.', 'classifai' ),
+				[
+					'status' => 403,
+				]
+			);
+		}
+
+		return $response;
+	}
 
 	/**
 	 * Register any needed endpoints.
--- a/includes/Classifai/Features/APIUsageTracking.php
+++ b/includes/Classifai/Features/APIUsageTracking.php
@@ -555,7 +555,7 @@ class APIUsageTracking extends Feature {
 		 * @param array                 $usage_data Usage data.
 		 * @param float                 $amount Amount of usage for the scope.
 		 */
-		do_action( 'classifai_api_soft_threshold_exceeded', $provider, $settings, $usage_data, $amount, $scope );
+		do_action( 'classifai_api_soft_threshold_exceeded', $provider, $settings, $usage_data, $amount );
 
 		$emails = $this->get_email_list( $settings['soft_threshold_emails'] ?? '' );
 
--- a/includes/Classifai/Features/APIUsageTracking.php
+++ b/includes/Classifai/Features/APIUsageTracking.php
@@ -103,6 +103,25 @@ class APIUsageTracking extends Feature {
 	}
 
 	/**
+	 * Returns the default settings for the feature.
+	 *
+	 * @return array
+	 */
+	protected function get_default_settings(): array {
+		$settings = parent::get_default_settings();
+
+		if ( empty( $settings['roles'] ) ) {
+			$settings['roles'] = [];
+		}
+
+		$settings['roles'] = [
+			'administrator' => 'administrator',
+		];
+
+		return $settings;
+	}
+
+	/**
 	 * Get the description for the enable field.
 	 *
 	 * @return string
--- a/includes/Classifai/Features/APIUsageTracking.php
+++ b/includes/Classifai/Features/APIUsageTracking.php
@@ -325,7 +325,7 @@ class APIUsageTracking extends Feature {
 		 */
 		$interval = apply_filters( 'classifai_api_usage_refresh_interval', $interval );
 
-		if ( ! \as_has_scheduled_action( self::CRON_HOOK ) ) {
+		if ( ! \as_has_scheduled_action( self::CRON_HOOK, [], 'classifai' ) ) {
 			\as_schedule_recurring_action( time(), $interval, self::CRON_HOOK, [], 'classifai' );
 		}
 	}
--- a/includes/Classifai/Features/APIUsageTracking.php
+++ b/includes/Classifai/Features/APIUsageTracking.php
@@ -560,26 +560,9 @@ class APIUsageTracking extends Feature {
 			return;
 		}
 
-		$period_key   = '';
-		$period_label = '';
-
-		// Period key is used to track the last sent alert for a period. Sent only once a month.
-		switch ( $scope ) {
-			case 'current_month':
-				$period_key   = gmdate( 'Y-m' );
-				$period_label = __( 'current month', 'classifai' );
-				break;
-			case 'year_to_date':
-				// i.e 2026-01-2026-02 for feb month.
-				$period_key   = gmdate( 'Y' ) . '-01-' . gmdate( 'Y-m' );
-				$period_label = __( 'year to date', 'classifai' );
-				break;
-			case 'all_time':
-				// i.e all-2026-02 for current year feb month.
-				$period_key   = 'all-' . gmdate( 'Y-m' );
-				$period_label = __( 'all time', 'classifai' );
-				break;
-		}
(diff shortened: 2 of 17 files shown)


includes/Classifai/Providers/OpenAI/APIRequest.php
```diff
@@ -73,6 +74,13 @@ public function __construct( string $api_key = '', string $feature = '', ?Provid
 	 * @return array|WP_Error
 	 */
 	public function get( string $url, array $options = [] ) {
+		if ( ! $this->is_request_allowed() ) {
```
dkotter: See my comment from the other PR: https://github.com/10up/classifai/pull/1054#discussion_r2824468322. I know we have a new check using `OpenAIUsage::get_openai_provider_ids` but I'd still suggest this sort of disabling shouldn't happen here as this class won't exist much longer

includes/Classifai/Services/UsageTracking.php
```diff
@@ -0,0 +1,44 @@
+<?php
+/**
+ * Service definition for Recommended Content
```
dkotter: Needs updated

includes/Classifai/Services/UsageTracking.php
```diff
@@ -0,0 +1,44 @@
...
+	public function __construct() {
+		parent::__construct(
+			__( 'Usage Tracking', 'classifai' ),
+			'usage_tracking',
+			self::get_service_providers()
+		);
+	}
+
+	/**
+	 * Get service providers for Content Recommendation service.
```
dkotter: Needs updating

includes/Classifai/Services/UsageTracking.php
```diff
@@ -0,0 +1,44 @@
...
+	 *
+	 * @return array
+	 */
+	public static function get_service_providers(): array {
+		/**
+		 * Filter the service providers for Recommendation service.
+		 *
+		 * @since 3.0.0
+		 * @hook classifai_recommendation_service_providers
```
dkotter: Needs updating

includes/Classifai/Services/ServicesManager.php
```diff
@@ -139,6 +140,24 @@ public function register_recommendation_service_features( array $features ): arr
 		return $features;
 	}
 
+	/**
+	 * Registers features under the Usage Tracking Service.
```
dkotter:
```suggestion
	 * Registers Features under the Usage Tracking Service.
```

includes/Classifai/Services/ServicesManager.php
```diff
@@ -139,6 +140,24 @@ public function register_recommendation_service_features( array $features ): arr
...
+	 *
+	 * @param array $features The list of features to be registered.
```
dkotter:
```suggestion
	 * @param array $features The list of Features to be registered.
```

includes/Classifai/Features/OpenAIUsage.php
```diff
@@ -0,0 +1,657 @@
...
+use WP_REST_Server;
+
+use function Classifai\get_asset_info;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Class TextToSpeech
```
dkotter: Needs updated

includes/Classifai/Features/OpenAIUsage.php
```diff
@@ -0,0 +1,657 @@
...
+		'start_year'   => 2020, // Start year of the usage.
+	];
+
+	/**
+	 * Constructor.
+	 */
+	public function __construct() {
+		$this->label = __( 'OpenAI usage tracking', 'classifai' );
+
+		// Contains all providers that are registered to the service.
```
dkotter:
```suggestion
		// Contains all Providers that are registered to the service.
```

includes/Classifai/Features/OpenAIUsage.php
```diff
@@ -0,0 +1,657 @@
...
+		$this->provider_instances = $this->get_provider_instances( UsageTrackingService::get_service_providers() );
+
+		// // Contains just the providers this feature supports.
```
dkotter:
```suggestion
		// Contains just the Providers this Feature supports.
```

includes/Classifai/Features/OpenAIUsage.php
```diff
@@ -0,0 +1,657 @@
...
+		];
+	}
+
+	/**
+	 * Get the description for the enable field.
+	 *
+	 * @return string
+	 */
+	public function get_enable_description(): string {
+		return esc_html__( 'Monitor OpenAI usage and set levels for alerting and deactivating features.', 'classifai' );
```
dkotter:
```suggestion
		return esc_html__( 'Monitor OpenAI usage and set levels for alerting and deactivating Features.', 'classifai' );
```

includes/Classifai/Features/OpenAIUsage.php
```diff
@@ -0,0 +1,657 @@
...
+	 * @param array $usage_data Usage data.
+	 * @return void
+	 */
+	public function set_usage_data( array $usage_data ): void {
+		$usage_data = wp_parse_args( $usage_data, $this->usage_data );
+		update_option( self::USAGE_DATA_KEY, $usage_data );
+	}
+
+	/**
+	 * Returns the default settings for the feature.
```
dkotter:
```suggestion
	 * Returns the default settings for the Feature.
```

includes/Classifai/Features/OpenAIUsage.php
```diff
@@ -0,0 +1,657 @@
...
+	public function get_feature_default_settings(): array {
+		return [
+			'provider'       => OpenAIUsageTracking::ID,
+			'api_start_year' => 2020,
+			'usage_data'     => [],
+		];
+	}
+
+	/**
+	 * Get the OpenAI provider IDs.
```
dkotter:
```suggestion
	 * Get the OpenAI Provider IDs.
```

includes/Classifai/Features/OpenAIUsage.php
```diff
@@ -0,0 +1,657 @@
...
+	 *
+	 * This is used to check if the API request is allowed for the OpenAI providers.
```
dkotter:
```suggestion
	 * This is used to check if the API request is allowed for the OpenAI Providers.
```

includes/Classifai/Features/OpenAIUsage.php
```diff
@@ -0,0 +1,657 @@
...
+	}
+
+	/**
+	 * Checks if the user has permission to access the endpoint.
+	 *
+	 * @return bool|WP_Error
+	 */
+	public function rest_endpoint_permissions_check() {
+
+		// Ensure the feature is enabled. Also runs a user check.
```
dkotter:
```suggestion
		// Ensure the Feature is enabled. Also runs a user check.
```

includes/Classifai/Features/OpenAIUsage.php
```diff
@@ -0,0 +1,657 @@
...
+
+		return true;
+	}
+
+	/**
+	 * Schedules the next cron run when the option exists and we have an admin key.
+	 */
+	public function maybe_schedule_cron(): void {
+
+		// We return early if feature is not enabled or action scheduler is not active.
```
dkotter:
```suggestion
		// We return early if Feature is not enabled or action scheduler is not active.
```

includes/Classifai/Features/OpenAIUsage.php
```diff
@@ -0,0 +1,657 @@
...
+
+		$settings = $this->get_settings( OpenAIUsageTracking::ID );
+
+		if ( ! empty( $settings['refresh_interval_minutes'] ) ) {
+			$interval = $settings['refresh_interval_minutes'] * MINUTE_IN_SECONDS;
+		} else {
+			$interval = self::DEFAULT_REFRESH_INTERVAL;
+		}
+
+		$interval = apply_filters( 'classifai_openai_usage_refresh_interval', $interval );
```
dkotter: Should add a filter docblock to this

includes/Classifai/Features/OpenAIUsage.php
```diff
@@ -0,0 +1,657 @@
...
+	 * Renders the dashboard widget content.
+	 */
+	public function render_dashboard_widget(): void {
+		$usage    = $this->get_usage_data();
+		$currency = $usage['currency'];
+		$fmt      = function ( $val ) use ( $currency ) {
+			return number_format_i18n( $val, 2 ) . ' ' . $currency;
+		};
+
+		echo '<p class="classifai-openai-usage-disclaimer">';
```
dkotter: This is hard to read with all the `echo` statements. I'd suggest we clean this up

includes/Classifai/Features/OpenAIUsage.php
```diff
@@ -0,0 +1,657 @@
...
+
+		$usage_data = wp_parse_args( $usage_data, $this->get_usage_data() );
+		$this->set_usage_data( $usage_data );
+
+		$settings = $this->get_settings( OpenAIUsageTracking::ID );
+
+		/**
+		 * Fires after OpenAI usage has been updated from the API.
+		 *
+		 * @since 3.8.0
```
dkotter:
```suggestion
		 * @since x.x.x
```

includes/Classifai/Features/OpenAIUsage.php
```diff
@@ -0,0 +1,657 @@
...
+			// Remove the sent key from the pricing option if already set.
+			unset( $settings[ $sent_key ] );
+			$this->update_settings( [ OpenAIUsageTracking::ID => $settings ] );
+			return;
+		}
+
+		/**
+		 * Fires when the soft threshold is exceeded.
+		 *
+		 * @since 3.8.0
```
dkotter:
```suggestion
		 * @since x.x.x
```

includes/Classifai/Features/OpenAIUsage.php
```diff
@@ -0,0 +1,657 @@
...
+
+		$bcc = 'Bcc: ' . implode( ', ', array_slice( $emails, 1 ) );
+		wp_mail( $emails[0], $subject, $message, [ $bcc ] );
+
+		$settings[ $sent_key ] = $period_key;
+		$this->update_settings( [ OpenAIUsageTracking::ID => $settings ] );
+	}
+
+	/**
+	 * Checks hard threshold, sets option to disable features, sends email.
```
dkotter:
```suggestion
	 * Checks hard threshold, sets option to disable Features, sends email.
```

includes/Classifai/Features/OpenAIUsage.php
```diff
@@ -0,0 +1,657 @@
...
+			$this->update_settings( [ OpenAIUsageTracking::ID => $settings ] );
+			return;
+		}
+
+		update_option( self::HARD_LIMIT_REACHED_KEY, true );
+
+		/**
+		 * Fires when the hard threshold is exceeded.
+		 *
+		 * @since 3.8.0
```
dkotter:
```suggestion
		 * @since x.x.x
```

includes/Classifai/Providers/OpenAI/APIRequest.php
```diff
@@ -368,4 +390,35 @@ public function get_api_key() {
...
+	private function is_request_allowed(): bool {
+		if ( ! in_array( $this->provider::ID, OpenAIUsage::get_openai_provider_ids(), true ) ) {
+			return true;
+		}
+
+		/**
+		 * Filter whether OpenAI API requests are allowed.
+		 * When the hard usage limit is reached and not overridden, this is set to false.
+		 *
+		 * @since 3.8.0
```
dkotter:
```suggestion
		 * @since x.x.x
```

includes/Classifai/Providers/OpenAI/UsageTracking.php
```diff
@@ -0,0 +1,453 @@
...
+use WP_Error;
+
+class UsageTracking extends Provider {
+	use OpenAI;
+
+	const ID = 'openai_usage_tracking';
+
+	/**
+	 * OpenAI organization admin API key endpoint.
+	 * Used for authenticate the organization admin API key.
```
dkotter:
```suggestion
	 * Used to authenticate the organization admin API key.
```

includes/Classifai/Providers/OpenAI/UsageTracking.php
```diff
@@ -0,0 +1,453 @@
...
+
+	/**
+	 * Base URL for the OpenAI API.
+	 *
+	 * @var string
+	 */
+	const COSTS_API_URL = 'https://api.openai.com/v1/organization/costs';
+
+	/**
+	 * OpenAI TextToSpeech constructor.
```
dkotter: Needs updated

includes/Classifai/Providers/OpenAI/UsageTracking.php
```diff
@@ -0,0 +1,453 @@
...
+			'all_time' => $past_years_costs + $current_year_months_costs + $current_mtd_costs,
+		];
+	}
+
+	/**
+	 * Fetch current month to date costs.
+	 *
+	 * @param bool $force Whether to force the fetch.
+	 *
+	 * @return int
```
dkotter: Think this needs updated to float

includes/Classifai/Providers/OpenAI/UsageTracking.php
```diff
@@ -0,0 +1,453 @@
...
+
+		return $cached_data['mtd'];
+	}
+
+	/**
+	 * Fetch past months costs.
+	 *
+	 * @param bool $force Whether to force the fetch.
+	 *
+	 * @return int
```
dkotter: Same here

includes/Classifai/Providers/OpenAI/UsageTracking.php
```diff
@@ -0,0 +1,453 @@
...
+
+		return $cached_data['years_total'] ?? 0;
+	}
+
+	/**
+	 * Fetch past years costs.
+	 *
+	 * @param bool $force Whether to force the fetch.
+	 *
+	 * @return int
```
dkotter: Same here

src/js/settings/components/provider-settings/openai-usage-tracking.js
```diff
@@ -0,0 +1,356 @@
...
+/**
+ * Internal dependencies
+ */
+import { STORE_NAME } from '../../data/store';
+import { SettingsRow } from '../settings-row';
+
+/**
+ * Component for OpenAI ChatGPT Provider settings.
+ *
+ * This component is used within the ProviderSettings component to allow users to configure the OpenAI ChatGPT Provider settings.
```
dkotter: Needs updated

src/js/settings/components/provider-settings/openai-usage-tracking.js
```diff
@@ -0,0 +1,356 @@
...
+							</SettingsRow>
+						</>
+					) }
+				</PanelBody>
+			</Panel>
+
+			<Panel
+				className="settings-panel"
+				header={ __(
+					'Hard threshold (disable features)',
```
dkotter:
```suggestion
					'Hard threshold (disable Features)',
```

src/js/settings/components/provider-settings/openai-usage-tracking.js
```diff
@@ -0,0 +1,356 @@
...
+					'classifai'
+				) }
+			>
+				<PanelBody>
+					<ToggleControl
+						label={ __( 'Enable hard threshold', 'classifai' ) }
+						help={ __(
+							'Disable all OpenAI features when hard limit is reached',
```
dkotter:
```suggestion
							'Disable all OpenAI Features when hard limit is reached',
```

src/js/settings/components/provider-settings/openai-usage-tracking.js
```diff
@@ -0,0 +1,356 @@
...
+							onChange( { hard_threshold_enabled: value } )
+						}
+					/>
+
+					{ providerSettings.hard_threshold_enabled && (
+						<>
+							<SettingsRow
+								label={ __( 'Amount', 'classifai' ) }
+								description={ __(
+									'When exceeded, optionally disable all OpenAI features.',
```
dkotter:
```suggestion
									'When exceeded, optionally disable all OpenAI Features.',
```

includes/Classifai/Providers/OpenAI/UsageTracking.php
```diff
@@ -0,0 +1,453 @@
...
+		}
+
+		$cached_data['months']       = $all_month_pricing;
+		$cached_data['months_total'] = array_sum( $all_month_pricing );
+		$cached_data['currency']     = $usage_currency;
+		$cached_data['last_updated'] = time();
+
+		$this->feature_instance->set_usage_data( $cached_data );
+
+		return $cached_data['years_total'] ?? 0;
```
dkotter: I think this should be `months_total`, not `years_total`

includes/Classifai/Admin/Notifications.php
```diff
@@ -390,6 +393,99 @@ public function ajax_maybe_dismiss_notice() {
...
+		// Don't show the notice if the provider is not OpenAI Usage Tracking.
+		$provider = $feature_instance->get_settings( 'provider' );
+		if ( OpenAIUsageTracking::ID !== $provider ) {
+			return;
+		}
+
+		$key = 'openai_threshold_reached';
+
+		// Don't show the notice if the user has already dismissed it.
+		if ( get_user_meta( get_current_user_id(), "classifai_dismissed_{$key}", true ) ) {
```
dkotter: I think there's a bug here that if someone dismisses the soft threshold, they'll never see the hard threshold notice because of this

includes/Classifai/Admin/Notifications.php
```diff
@@ -390,6 +393,99 @@ public function ajax_maybe_dismiss_notice() {
...
+
+		$usage_data = $feature_instance->get_usage_data();
+		$amount     = $feature_instance->get_amount_for_scope( $usage_data, $scope );
+		$threshold  = $hard_threshold_reached ? (float) $settings['hard_threshold_amount'] : (float) $settings['soft_threshold_amount'];
+
+		if ( $amount < $threshold ) {
+			return;
+		}
+
+		$key = 'openai_threshold_reached';
```
dkotter: This was already defined above so I think this is just an unnecessary duplicate

includes/Classifai/Providers/OpenAI/APIRequest.php
```diff
@@ -368,4 +390,35 @@ public function get_api_key() {
...
 		return $this->api_key ?? '';
 	}
+
+	/**
+	 * Whether API requests are allowed (not blocked by hard usage limit).
+	 *
+	 * @return bool
+	 */
+	private function is_request_allowed(): bool {
+		if ( ! in_array( $this->provider::ID, OpenAIUsage::get_openai_provider_ids(), true ) ) {
```
dkotter: `$this->provider::ID` may be `null` which would throw a fatal error here

includes/Classifai/Admin/Notifications.php
```diff
@@ -390,6 +393,99 @@ public function ajax_maybe_dismiss_notice() {
 		update_user_meta( get_current_user_id(), "classifai_dismissed_{$notice_id}", true );
 	}
 
+	/**
+	 * Renders a dismissible warning when OpenAI usage exceeds the hard threshold limit.
```
dkotter: I don't think this is accurate

src/js/settings/components/provider-settings/openai-usage-tracking.js
```diff
@@ -0,0 +1,356 @@
...
+		{
+			label: __( 'All time', 'classifai' ),
+			value: 'all_time',
+		},
+	];
+
+	// Refresh site data periodically while a background force refresh is scheduled.
+	// Once the force refresh is completed, the site data updates are picked up by existing
+	// useSelect hooks, automatically updating the UI.
+	useEffect( () => {
```
dkotter, line 84: Looks like this is the same code that runs in `src/js/features/openai-usage/index.js`. I'd suggest extracting that out into a reusable hook

includes/Classifai/Features/OpenAIUsage.php
```diff
@@ -0,0 +1,657 @@
...
+		$message = sprintf(
+			/* translators: 1: amount, 2: currency, 3: period */
+			__( 'OpenAI usage has exceeded your soft limit of %1$s %2$s for this period (%3$s).', 'classifai' ),
+			number_format_i18n( $threshold, 2 ),
+			$usage_data['currency'],
+			$period_label
+		);
+
+		$bcc = 'Bcc: ' . implode( ', ', array_slice( $emails, 1 ) );
+		wp_mail( $emails[0], $subject, $message, [ $bcc ] );
```
dkotter: If we only have a single email, `$bcc` will be empty but we still add that header. I'm worried this might cause delivery issues so we should only add this if needed

includes/Classifai/Providers/OpenAI/UsageTracking.php
```diff
@@ -0,0 +1,453 @@
...
+					$hard_threshold_emails = array_map( 'sanitize_email', $hard_threshold_emails );
+					$hard_threshold_emails = array_filter( $hard_threshold_emails, 'is_email' );
+					$hard_threshold_emails = array_unique( $hard_threshold_emails );
+					$hard_threshold_emails = implode( ',', $hard_threshold_emails );
+				}
+
+				$new_settings[ static::ID ]['hard_threshold_enabled'] = $hard_threshold_enabled;
+				$new_settings[ static::ID ]['hard_threshold_amount']  = filter_var(
+					$new_settings[ static::ID ]['hard_threshold_amount'] ?? false,
+					FILTER_VALIDATE_INT,
```
dkotter: This threshold uses a step of 0.01 but this validation will force integers only, no floats

includes/Classifai/Features/OpenAIUsage.php
```diff
@@ -0,0 +1,657 @@
...
+			'normal'
+		);
+	}
+
+	/**
+	 * Renders the dashboard widget content.
+	 */
+	public function render_dashboard_widget(): void {
+		$usage    = $this->get_usage_data();
+		$currency = $usage['currency'];
```
dkotter: The `currency` key may not exist yet which will throw a warning here

includes/Classifai/Features/APIUsageTracking.php
```diff
@@ -0,0 +1,657 @@
...
+
+	/**
+	 * Cron hook for the usage refresh.
+	 *
+	 * @var string
+	 */
+	const CRON_HOOK = 'classifai_openai_usage_refresh';
+
+	/**
+	 * Cron hook for the usage refresh.
```
dkotter, line 38: This is the same docblock as the one above. Should update these to be unique

includes/Classifai/Providers/OpenAI/UsageTracking.php
```diff
@@ -0,0 +1,453 @@
...
+
+		switch ( $this->feature_instance::ID ) {
+			case FeatureOpenAIUsage::ID:
+				return array_merge(
+					$common_settings,
+					[
+						'api_key' => '',
+					]
+				);
+		}
```
dkotter: This is redundant right now so I'd remove it

includes/Classifai/Providers/OpenAI/APIRequest.php
```diff
@@ -368,4 +390,35 @@ public function get_api_key() {
...
+		 *
+		 * @since 3.8.0
+		 * @hook classifai_openai_can_make_request
+		 * @param bool $allowed Whether the request is allowed. Default true.
+		 * @param string $feature The feature name.
+		 * @param Provider $provider The provider instance.
+		 *
+		 * @return bool
+		 */
+		$allowed = (bool) apply_filters( 'classifai_can_make_request', true, $this->feature, $this->provider );
```
dkotter:
```suggestion
		$allowed = (bool) apply_filters( 'classifai_openai_can_make_request', true, $this->feature, $this->provider );
```

includes/Classifai/Features/OpenAIUsage.php
```diff
@@ -0,0 +1,657 @@
...
+		if (
+			strpos( $route, '/classifai/v1/openai-usage/force-refresh' ) !== 0
+			|| ! function_exists( 'as_enqueue_async_action' )
+			|| (
+				function_exists( 'as_has_scheduled_action' )
+				&& \as_has_scheduled_action( self::FORCE_CRON_HOOK, [], 'classifai' )
+			)
+		) {
+			return parent::rest_endpoint_callback( $request );
+		}
```
dkotter: Not sure I understand what this logic is trying to do? Particularly in the case where the action is already scheduled, why do we need to call the parent method?

includes/Classifai/Providers/OpenAI/UsageTracking.php
```diff
@@ -0,0 +1,453 @@
...
+					$soft_threshold_emails = implode( ',', $soft_threshold_emails );
+				}
+
+				$new_settings[ static::ID ]['soft_threshold_enabled'] = $soft_threshold_enabled;
+				$new_settings[ static::ID ]['soft_threshold_amount']  = filter_var(
+					$new_settings[ static::ID ]['soft_threshold_amount'] ?? false,
+					FILTER_VALIDATE_INT,
+					[
+						'options' => [
+							'min_range' => 1,
```
dkotter: We set the minimum to 1 here but in our settings, we set that to 0. Should standardize that

dkotter (changes requested):
Left a decent number of comments here that need addressed.

That said, I still don't know if this PR matches my expectations in terms of extensibility ([same comment](https://github.com/10up/classifai/pull/1054#pullrequestreview-3822240263) I had on the first PR).

If I wanted to add usage tracking for another Provider, I'd have to duplicate basically all the code here. I still think there's a better approach where we can extract out all of the shared functionality, making it much easier and cleaner to add support in the future.

I'm also not sure on the use of our Feature and Provider classes here. I know that's how the rest of ClassifAI is set up but I don't think these things really fit neatly into that paradigm. Doesn't necessarily need changed but something to consider.

**Edit**: at the very least, I think if we want to follow this structure we need to make the Usage Tracking Feature generic, it should not be tied to OpenAI at all (this is how all other Features work). Any OpenAI specific things can then be in the Usage Provider. I'd also recommend we create a general Usage Provider that contains all shared functionality. Then the OpenAI specific Usage Provider can extend that (or similar)

**Second edit**: Thinking about this some more, the above wouldn't quite work as we only support a single Feature -> Provider setup, where here ideally we can turn on usage tracking for multiple Providers. So I go back to my original thought that not sure our standard Feature / Provider setup works for this, at least not without some changes to how we render the settings and allow things to be turned on and configured

rahulsprajapati:
> That said, I still don't know if this PR matches my expectations in terms of extensibility (https://github.com/10up/classifai/pull/1054#pullrequestreview-3822240263 I had on the first PR).

I tried using that approach, but with that it will be whole different structure compare to current ClassifAI structure. Because of that we don't get to reuse same UI/UX and need lots of changes to make UI/UX match current one. So I tried to use existing Provider + Feature classes, which help reuse existing UI/UX. Maybe we need to separate out UI/UX code from current structure and allow any new/extend feature with same UI/UX without having to reuse Provider+Feature structure. 

> If I wanted to add usage tracking for another Provider, I'd have to duplicate basically all the code here. I still think there's a better approach where we can extract out all of the shared functionality, making it much easier and cleaner to add support in the future.

I think since I'm not sure what other provide feature looks like (how we get usage data, do we need to calculate similar way or get it directly from one endpoint, or require diff admin api key only) because of that I'm unsure to keep which feature/code shared. Like we can have fetch_period method shared but again not sure if we will have same type of api calls. So we can have interface without defined method but not sure what else we can do here. 
Maybe once we have another provider as well, we can get more idea on which code/method can be shared. 

> I'm also not sure on the use of our Feature and Provider classes here. I know that's how the rest of ClassifAI is set up but I don't think these things really fit neatly into that paradigm. Doesn't necessarily need changed but something to consider.

For this, looks like we might need full refactor on plugin itself. Since without using this Feature and Provider we don't get to reuse existing UI/UX and require lots of changes in style and scripts.

----

About the blocking direct api calls:
- I didn't wanted to disable feature itself since that looks confusing when usage is exceeded. Since editor might see as an bug/issue that feature disabled automatically without any notice and try request admin to enable again. We do have notice but that doesn't show up on post edit screen or media uploads screen.
- When we block it from api itself, we are keeping all feature as it is but show usage exceed error on each feature usage, which is clear message for why we can't use feature. So to me, displaying api call error when using feature looks better. 
- Even though this will move out somewhere else, this api call will happen from some or the other file. That can still use same code we already have. This will be moved somewhere else but won't be removed. 

But let me know what you think. If this doesn't make sense I can move this block to disable feature itself or whichever other place we can do this. 

I'll update other changes and setup a sync to discuss more on this and get this feature ready asap.

includes/Classifai/Admin/Notifications.php
rahulsprajapati: yeah I think I must have missed this in some code conflict and that ended up making this duplicate. The above defined one was not suppose to be there.

includes/Classifai/Admin/Notifications.php
rahulsprajapati: Yeah this one was not suppose to be there 🤦 , I handled this below and updated check to dismiss only soft limit notice

includes/Classifai/Features/OpenAIUsage.php
rahulsprajapati: Good callout. I'll update this.

includes/Classifai/Providers/OpenAI/APIRequest.php
rahulsprajapati: Let's discuss this over sync. Not sure what other possible solution for this. Disabling feature all together seems confusing for end users.

includes/Classifai/Providers/OpenAI/UsageTracking.php
rahulsprajapati: ohh, not sure why I didn't catch this my test. This was obvious issue that should have been caught in my test. 🤦  
Somehow, I was able to see correct amount. (I might have cached data that was set earlier correctly)

includes/Classifai/Services/UsageTracking.php
```diff
@@ -0,0 +1,44 @@
...
+	}
+
+	/**
+	 * Get service providers for Usage Tracking service.
+	 *
+	 * @return array
+	 */
+	public static function get_service_providers(): array {
+		/**
+		 * Filter the service providers for Recommendation service.
```
dkotter: This isn't accurate, referencing the Recommendation service

includes/Classifai/Admin/Notifications.php
```diff
@@ -390,6 +393,93 @@ public function ajax_maybe_dismiss_notice() {
...
+		if ( $amount < $threshold ) {
+			return;
+		}
+
+		$key = 'api_threshold_reached';
+
+		// Don't show the notice if the user has already dismissed it for soft threshold. Always show for hard threshold.
+		if ( ! $hard_threshold_reached && get_user_meta( get_current_user_id(), "classifai_dismissed_{$key}", true ) ) {
+			return;
+		}
```
dkotter: In thinking about this again, since we track thresholds on a specific scope (like monthly), once a user dismisses a soft threshold alert in that scope, they'll never see another soft threshold alert, even in a new scope period. I think we either need the scope as part of the key here or make it so neither threshold notice is dismissible

includes/Classifai/Admin/Notifications.php
```diff
@@ -390,6 +393,93 @@ public function ajax_maybe_dismiss_notice() {
...
+
+		$classes = [
+			'notice',
+		];
+
+		if ( $hard_threshold_reached ) {
+			$classes[] = 'notice-error';
+			$message   = sprintf(
+				/* translators: 1: amount with currency, 2: link to settings */
+				__( 'AI Features are currently disabled due to exceeded your hard limit of %1$s for this period. <a href="%2$s">Re-enable it from the pricing page</a>.', 'classifai' ),
```
dkotter:
```suggestion
				__( 'AI Features are currently disabled due to exceeding your hard limit of %1$s for this period. <a href="%2$s">Re-enable it from the pricing page</a>.', 'classifai' ),
```

includes/Classifai/Features/APIUsageTracking.php
```diff
@@ -0,0 +1,709 @@
...
+
+	/**
+	 * Checks if the user has permission to access the endpoint.
+	 *
+	 * @return bool|WP_Error
+	 */
+	public function rest_endpoint_permissions_check() {
+
+		// Ensure the Feature is enabled. Also runs a user check.
+		if ( ! $this->is_feature_enabled() ) {
```
dkotter: I know this is the approach we use on other Features, checking if the user has access based on the settings in the Feature. In this case though, we're allowing those users to trigger a full refresh of all our data which seems like something we may want to force a lockdown to admins only. I'd suggest adding a `manage_options` check here in addition to the enabled check

includes/Classifai/Features/APIUsageTracking.php
```diff
@@ -0,0 +1,709 @@
...
+		 * @since x.x.x
+		 * @hook classifai_api_usage_tracking_refresh_interval
+		 *
+		 * @param int $interval The refresh interval in seconds.
+		 *
+		 * @return int The filtered refresh interval.
+		 */
+		$interval = apply_filters( 'classifai_api_usage_refresh_interval', $interval );
+
+		if ( ! \as_has_scheduled_action( self::CRON_HOOK, [], 'classifai' ) ) {
```
dkotter, line 347: If someone changes the interval (say from 15 minutes to 60 minutes) I don't know if things will actually be rescheduled here. We may need a check to see if the currently scheduled action has a different interval and if so, reschedule the action with the new interval

includes/Classifai/Features/APIUsageTracking.php
```diff
@@ -0,0 +1,709 @@
...
+			get_asset_info( 'classifai-plugin-api-usage-tracking', 'version' ),
+			true
+		);
+	}
+
+	/**
+	 * Registers the dashboard widget when configured.
+	 */
+	public function register_dashboard_widget(): void {
+		if ( ! $this->is_feature_enabled() ) {
```
dkotter: Technically we don't need this since the method is only called from the `feature_setup` method, which itself is only called if `is_feature_enabled` returns true

includes/Classifai/Features/APIUsageTracking.php
```diff
@@ -0,0 +1,709 @@
...
+			$interval = $settings['refresh_interval_minutes'] * MINUTE_IN_SECONDS;
+		} else {
+			$interval = self::DEFAULT_REFRESH_INTERVAL;
+		}
+
+		/**
+		 * Filter the refresh interval for the AI API usage tracking.
+		 *
+		 * @since x.x.x
+		 * @hook classifai_api_usage_tracking_refresh_interval
```
dkotter:
```suggestion
		 * @hook classifai_api_usage_refresh_interval
```

includes/Classifai/Features/APIUsageTracking.php
```diff
@@ -0,0 +1,709 @@
...
+
+		$headers    = [];
+		$bcc_emails = array_slice( $emails, 1 );
+		if ( ! empty( $bcc_emails ) ) {
+			$headers[] = 'Bcc: ' . implode( ', ', $bcc_emails );
+		}
+		wp_mail( $emails[0], $subject, $message, $headers );
+
+		$settings[ $sent_key ] = $period_key;
+		$this->update_settings( [ $provider::ID => $settings ] );
```
dkotter: I've not compared line-by-line but seems like most (if not all) of this code is duplicated in the `check_hard_threshold` method below. Ideally we extract that out to shared method

includes/Classifai/Providers/OpenAI/UsageTracking.php
```diff
@@ -0,0 +1,221 @@
...
+					$hard_threshold_emails = explode( ',', $hard_threshold_emails );
+					$hard_threshold_emails = array_map( 'trim', $hard_threshold_emails );
+					$hard_threshold_emails = array_map( 'sanitize_email', $hard_threshold_emails );
+					$hard_threshold_emails = array_filter( $hard_threshold_emails, 'is_email' );
+					$hard_threshold_emails = array_unique( $hard_threshold_emails );
+					$hard_threshold_emails = implode( ',', $hard_threshold_emails );
+				}
+
+				$new_settings[ static::ID ]['hard_threshold_enabled'] = $hard_threshold_enabled;
+				$new_settings[ static::ID ]['hard_threshold_amount']  = filter_var(
```
dkotter, line 139: Minor thing but we may want a check to ensure the hard threshold is higher than the soft threshold

includes/Classifai/Providers/UsageTrackingProvider.php
```diff
@@ -0,0 +1,371 @@
...
+		$project_id = $this->get_credential( 'project_id', $settings );
+
+		$total_amount = 0.0;
+		$currency     = 'USD';
+		$page         = 1;
+		$has_more     = true;
+		$is_null_data = false;
+
+		while ( $has_more ) {
+			$args = $this->usage_api_args( $start_ts, $end_ts, $page );
```
dkotter, line 142: We pass `$page` as an integer here but the `usage_api_args` method expects that to be a string. Should either update that here or update `usage_api_args` to expect an integer

includes/Classifai/Providers/UsageTrackingProvider.php
```diff
@@ -0,0 +1,371 @@
...
+	 * Returns the default settings for the provider.
+	 *
+	 * @return array
+	 */
+	public function get_default_provider_settings(): array {
+		return [
+			'api_key'                  => '',
+			'authenticated'            => false,
+			'project_id'               => '',
+			'refresh_interval_minutes' => 15,
```
dkotter, line 78: Unless I missed it, I don't think we have sanitization on either of these values when they are saved. Should probably use `sanitize_text_field` on the `project_id` and for `refresh_interval_minutes`, we should ensure this is an integer that is greater or equal to 1, so we don't allow zero or a negative number

dkotter (changes requested):
Overall this looks good, much happier with the structure here and how this sets us up for the future. Left a handful of comments, let me know if there's questions on any of that.

Beyond those, would be great to get some E2E tests here, at least to verify the settings render properly (would be most ideal to ensure threshold notices render as well, though would require us to mock the proper API responses).

Also, slightly concerned with performance here when fetching all historical data (either when run the first time or when a force refresh is done). Curious if you've seen any issues when running this to fetch historical data?

jeffpaul:
1. I _think_ the provider option here should just be "OpenAI"?

Screenshot 2026-03-17 at 1 21 50 PM

2. Let's update the default set of roles that can access to just Administrator and force an admin to grant lower roles access to this info.

Screenshot 2026-03-17 at 1 30 16 PM

3. Can probably update the widget title to `AI Usage Tracking`.  Please right align the three values, prepend with a currency sign (in this case `$`).  Probably worth updating buttons to Title Case as I don't think Sentence case is standard, perhaps updating the second to `Refresh Usage Data` (doing the same with the button text on the settings page).  What's the `Updating...` static text meant to reference?

Screenshot 2026-03-17 at 1 31 55 PM

Screenshot 2026-03-17 at 1 36 17 PM

There are some other small copy tweaks I'd make and I've not tested the soft or hard thresholds as I suspect we'll get better testing there as folks use the feature and can iterate based on that.

jeffpaul:
4. Somehow (fortunately?) the soft threshold seemed to have tripped and looks like we should update the copy to "soft threshold of $" and capitalize USD

Screenshot 2026-03-17 at 5 19 44 PM

includes/Classifai/Admin/Notifications.php
rahulsprajapati: Done: https://github.com/10up/classifai/pull/1060/commits/26cd80bd105f8be805a3a0fc16c108faa4872876

includes/Classifai/Features/APIUsageTracking.php
rahulsprajapati: done: https://github.com/10up/classifai/pull/1060/commits/016cf459ed0871cc1b246c1979e9bbd51b000e44

includes/Classifai/Features/APIUsageTracking.php
rahulsprajapati: Done: https://github.com/10up/classifai/pull/1060/commits/456fb714feb311a182acdcf93ca2330429cde114

I think we don't currently have any direct way to check interval for scheduled action. I tried using `as_get_scheduled_actions` but it has private/protected value in object. So it seems bit hacky solution. 

So for now, I just unschedule when interval changes in setting itself.

includes/Classifai/Features/APIUsageTracking.php
rahulsprajapati: Done: https://github.com/10up/classifai/pull/1060/commits/e492723d9a4fb7640ab1612da242a4aae9bd9214

includes/Classifai/Features/APIUsageTracking.php
rahulsprajapati: That's right, most of the code is same. But moving all of it require multiple (4-5+ param) in new function, or require multiple else if conditions to keep it in one place.

For now I moved common scope label/key code to reusable function. 

https://github.com/10up/classifai/pull/1060/commits/b06872cae796f01f4bb6be8d0fb6aa4b5d1137ba

includes/Classifai/Providers/OpenAI/UsageTracking.php
rahulsprajapati: Good catch, just updated this to always have hard limit value greater then soft one: https://github.com/10up/classifai/pull/1060/commits/32ff80390457921aba8f2f98eaacb7abc632ad34

includes/Classifai/Providers/UsageTrackingProvider.php
rahulsprajapati: Done: https://github.com/10up/classifai/pull/1060/commits/a31fc5c5c88f4a579f5acc41a3603eee1f12759a

includes/Classifai/Providers/UsageTrackingProvider.php
rahulsprajapati: Done: https://github.com/10up/classifai/pull/1060/commits/6773148c050a3db841d546f060b63093e8069cef

Kept minimum interval to 5min only, since we have api callback in cron and sometime it might take a min to complete full cron job. Specially force refresh or first time data refresh job. So keeping it minimum 5 should help.

rahulsprajapati:
> Beyond those, would be great to get some E2E tests here, at least to verify the settings render properly (would be most ideal to ensure threshold notices render as well, though would require us to mock the proper API responses).

@dkotter  Sure, I can start on E2E now. Since we had couple of back-forth and refactoring code itself. Wanted to get approval on codebase first, and then start on E2E. 

> Also, slightly concerned with performance here when fetching all historical data (either when run the first time or when a force refresh is done). Curious if you've seen any issues when running this to fetch historical data?

I didn't face issue yet, tried on two projects. It fetched data from 2020 to 2026 (Each request is targeting 6month data in one go. Since that's upper limit 180 from OpenAI side). So almost 11-12 api request on first time, but all executed in background only. 
- I'm also keeping starting year in DB, so we don't always fetch from 2020 itself but from the day we start getting usage on API. i.e if api key was created in 2023 then it won't have any data prior to that, so on force refresh we can directly start from 2023. 

For every other cron job requests, we only fetch current month to date data from API. All past data already cached. i.e Till last year 2025 data are stored as yearly values and for current year last month wise. So for new request we only fetch data from 1st March to current date and we will get Jan, Feb data from cache. 

Since this is all happening in background, and only on first or force refresh time. This should not cause any performance issues. 

But let me know if you see any issue on this or have alternate solution for this. :)

rahulsprajapati:
> What's the Updating... static text meant to reference?

@jeffpaul That's for referring to current state of all usage data. As you can see in the screenshot you share, we Year to Date and All Time value is $0. So in your case it's still in progress and updating all data. 

Maybe need a better way to do this, since looks like it's not that obvious. Maybe updating it to `Fetching..` or showing it directly in place of `$0` could help..

dkotter (dismissed):
Overall this looks good to me. Approving now though ideally we get some E2E tests here and I think there's still some UI feedback from @jeffpaul

jeffpaul:
> That's for referring to current state of all usage data. As you can see in the screenshot you share, we Year to Date and All Time value is $0. So in your case it's still in progress and updating all data.

> Maybe need a better way to do this, since looks like it's not that obvious. Maybe updating it to Fetching.. or showing it directly in place of $0 could help..

I think the fact that the refresh button is already in a loading state suffices, let's remove that `Updating...` text.

rahulsprajapati:
> I think the fact that the refresh button is already in a loading state suffices, let's remove that Updating... text.

@jeffpaul that's just when we do manual refresh. But we don't have loading state for initial/first time data fetch.

jeffpaul:
> that's just when we do manual refresh. But we don't have loading state for initial/first time data fetch.

in that case, put "Updating..." in place of the currency display?

rahulsprajapati:
> in that case, put "Updating..." in place of the currency display?
Sure, Here is how it will look now:

During initial data fetch:

image

Data will show if it's already fetched for even one of dataset:
image

Fetched all data:

image

dkotter (approved):

dkotter:
I've gone ahead and merged this in but note ideally we come back to adding E2E tests here to ensure this functionality continues to work in the future cc / @rahulsprajapati @jeffpaul

rahulsprajapati:
@dkotter FYI: There is one php warning in this, fixed in new PR https://github.com/10up/classifai/pull/1065

PR #1080, opened 2026-05-28
Fix a potential fatal error when building user roles

dkotter:
### Description of the Change

We had a fatal error reported:

> PHP Fatal error: Uncaught ValueError: array_combine(): Argument number 1 ($keys) and argument number 2 ($values) must have the same number of elements

This pointed to our code that builds the list of roles that each Feature uses to determine which roles have access to the Feature. In looking at that code, if a role somehow doesn't have the `name` attribute set, you end up with two arrays of differing lengths and thus running `array_combine` will throw an error.

In a standard WordPress install I couldn't reproduce this but it seems like a good place to harden our code a bit. This updates both our general `setup_roles` method and the `setup_roles` method we run in the Image Generation Feature.

### How to test the Change

1. Checkout this PR
2. Enable a Feature like Title Generation
3. Go to the settings for that Feature and toggle open the `User Permissions` panel
4. Ensure you see all expected roles in the `Allowed roles` list
5. Go to the settings for Image Generation and check there as well
6. Ensure if you disallow a user role, like Administrator, that users with that role can't use the Feature

### Changelog Entry

> Fixed - Better handling of user roles to avoid a potential fatal error

### Credits

Props @dkotter, @gthayer 

### Checklist:

### DIFF
--- a/.github/workflows/cypress.yml
+++ b/.github/workflows/cypress.yml
@@ -18,7 +18,7 @@ jobs:
       matrix:
         core:
           - {name: 'WP latest', version: 'latest'}
-          - {name: 'WP minimum', version: 'WordPress/WordPress#6.8'}
+          - {name: 'WP minimum', version: 'https://wordpress.org/wordpress-6.8.zip'}
           - {name: 'WP trunk', version: 'WordPress/WordPress#master'}
 
     steps:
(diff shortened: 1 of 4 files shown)


dkotter:
Note Plugin Check isn't working due to an upstream issue (see https://github.com/WordPress/plugin-check-action/issues/579).

And our E2E tests aren't running on our minimum environment (WP 6.8) as it seems WooCommerce (which we install the latest version for some tests) has set their minimum to 6.9 and so the install fails. We can install an older version of Woo but I think it's fine to bump our minimum WP version to 6.9 but I'd suggest that's a separate PR.

jeffpaul (approved):

PR #1102, opened 2026-06-08
Introduce a stand-alone table to store embedding data

### LINKED ISSUE #889: Refactor and Clean Up Duplicate Code for Embedding Providers
There is duplicated code across embedding providers. Refactor this to reduce redundancy by moving shared code into a central location, either within the feature or into a shared class. Ref #881

### LINKED ISSUE #975: Embedding data causing cache failures
### Describe the bug

We have a handful of Features that can be powered by embeddings generated by an LLM. These embeddings are currently stored in either post meta or term meta and then used to run comparisons.

It's a known issue that this doesn't scale very well, as running these comparisons within WordPress starts to slow down significantly once you have hundreds or thousands of items. We probably haven't done a good enough job of making that limitation known though.

But another issue that came up recently is that this embedding data can get quite large. The way this currently works is we take the content of an item (say a post) and we break that down into smaller chunks. Each chunk is then sent to the LLM to generate embeddings and each of those embeddings are then stored together under a single meta key.

For long content, this data can easily get over 1MB. WordPress has some built-in functionality that in certain situations (like when running `get_posts` or `get_post_meta`), it will run a database query to get all meta for that item and store that in the cache, with the idea that this will make any subsequent requests for this data faster.

The problem here is this means in certain situations, this embedding data gets pulled into the cache and it can easily be large enough to overwhelm the cache size limit, which then forces all cached data to be purged. For sites with lots of traffic, this can lead to performance issues as more requests need to make database queries to get the data they need.

### Approaches

I think there are two approaches we should look at implementing here:

1. For any Feature that uses embeddings that doesn't currently support storing those in elasticsearch, add that functionality (Classification and Recommended Content)
2. For sites that don't have access to elasticsearch, add a new database table to store embeddings instead of using the meta tables

### Elasticsearch

Right now, the Smart 404 and Term Cleanup Features can take advantage of elasticsearch (through ElasticPress) to store and query embeddings. This leads to significant performance improvements on the query side and does mean we don't need to store the data in the meta tables, fixing the issue described above.

We should look to bring this same functionality to all other Features that use embeddings, as well as adjust the current approach to only store in elasticsearch (right now, those two existing Features will store in both places).

### New DB table

In addition to the above, we should look at introducing a new database table, designed for this embedding data. This prevents the problem discussed above and also allows us to design this table specifically to handle embeddings, whereas right now the meta tables are set to handle lots of data types. This will likely lead to better performing queries but will take some experimentation on how best to structure this (I would start by looking at https://github.com/Jameswlepage/wpvdb and seeing if there's things there we can use/learn from). Will also need to consider backwards compat here, if we should look to migrate existing embedding data from meta tables to this new table.

I would recommend we tackle this part first and then the elasticsearch part second, as I think this has more applicable use cases.

### Steps to Reproduce

1. Enable a Feature that uses embeddings
2. Create a long post and trigger embedding generation for that
3. View in your database the size of the `classifai_openai_embeddings` post meta item
4. If desired, set up an environment that has caching enabled and see how the above impacts that

_No response_

### Environment information

_No response_

### WordPress information

_No response_

### Code of Conduct

dkotter:
### Description of the Change

We have a number of Features that when enabled will send content off to an embedding provider and then store the resulting vector data in post or term meta. While this works, the amount of potential data here is very large and doesn't fit well into that table. And more concretely, this can cause cache failures as WordPress will try and prime the cache in certain circumstances with all post meta, leading to a cache overflow.

This PR introduces a new DB table (`wp_classifai_embeddings`), that is used to store this embedding data. This table is automatically created when needed and a whole CRUD layer is included here to ensure data is consumed from this table instead of the meta tables.

Because any existing ClassifAI site may have data already in the meta tables, this PR also includes a migration routine that will automatically migrate all data from meta and into the new table, deleting the data from the meta tables once done. In theory this should mean this change is a complete backwards compat update, though I am currently tagging this for the v4.0 release (though I'd personally be fine shipping this in v3.9).

Closes #889, #975

### How to test the Change

Multiple paths to test here, both the migration routine and that creating and using embeddings continues to work:

1. Ensure you have an environment that has existing embedding data stored in a meta table (either an environment already set up or create a new one with the latest release of ClassifAI and generate some embedding data)
2. Pull down this PR
3. Go to the ClassifAI settings screen and refresh; ensure you see an admin notice letting you know migration is in process
4. Access your site database and ensure you see a new `wp_classifai_embeddings` table
5. Ensure you see data start to populate here

Once migration is done:

1. Enable a Feature that uses embeddings (Classification, Term Cleanup, Recommended Content)
2. Ensure this is setup with a valid embeddings Provider
3. Create a new term. Inspect your database and ensure you see a new entry in the `wp_classifai_embeddings` for that term
4. Update that term and ensure a new row isn't created but the existing row is updated
5. Delete that term and ensure that row is also deleted
6. Do the same for a post (assuming you have Recommended Content turned on). Ensure you see the data land in the new table, updated in that table and deleted from that table
7. Try out these Features and ensure you see results (easiest is probably the Classification Previewer)

### Changelog Entry

> Added - Custom table to store embedding data
> Added - New CRUD methods to interact with the new embeddings table
> Added - Migration script to automatically migrate embeddings data from meta tables to the new table
> Changed - Update all existing embedding integrations to use the new embedding CRUD methods, ensuring data is stored and pulled from our new table
> Changed - Refactor existing embeddings code to avoid duplication

### Credits

Props @zamanq, @dkotter

### Checklist:

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2F10up%2Fclassifai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-1102-98430e5f417491558f58bc1137073baea8a2f5ef.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


jeffpaul (approved):
Some merge conflicts to resolve, but otherwise good to move along

dkotter:
I'm going to hold off on this for the moment and wait to see where https://github.com/WordPress/ai/pull/976 and similar things land. Just don't want to duplicate effort and since it's likely we'll see more movement on the AI plugin side, going to see what feedback we get there to help inform this one

jeffpaul:
@dkotter besides #976 are there other issues/PRs blocking progress here?

PR #1109, opened 2026-06-11
Ensure returned content matches our format by using `response_format`

dkotter:
### Description of the Change

OpenAI and most other LLMs now support a `response_format` field that allows you to request the format in a certain JSON schema. This is more reliable than asking for a certain response type in the prompt as some LLMs will follow that and some don't.

Image Tags Generator is the main Feature being updated here as the prompt asked the LLM to return 2-4 tags in a list. Now we instead pass in the JSON schema we want in our request which makes it way more reliable to parse out those tags.

The other Feature updated here is Key Takeaways, which was already passing the `response_format` for OpenAI and Azure OpenAI but not Ollama, so updated that.

In addition, our list of Ollama vision models was out of date so updated that as well.

### How to test the Change

1. Enable Key Takeaways and Image Tags Generator
2. Set these up with OpenAI or Ollama (if Ollama, ensure you have a proper model selected)
3. Test those Features out and ensure they still work

### Changelog Entry

> Changed - Use the `response_format` parameter to ensure the response we get matches what we expect

### Credits

Props @dkotter

### Checklist:

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2F10up%2Fclassifai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-1109-90a38b8f65eaa63a2ee62115a8438e0ec2345b22.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Classifai/Features/Prompts/ImageTagsGenerator/default.php
+++ b/includes/Classifai/Features/Prompts/ImageTagsGenerator/default.php
@@ -18,8 +18,6 @@ Generate 3-5 tags that:
 - Do not infer identities, emotions, age, gender, race, or ethnicity unless these are visually unambiguous and central to the image
 - Avoid generic photography meta-tags: "stock photo", "high resolution", "professional photography", "image", "picture", "photo"
 
-Format the output as a Markdown bulleted list, one tag per line, prefixed with "- ".
-
-Output only the list. Do not add a heading, preamble ("Here are the tags:"), trailing commentary, ellipsis, or numbering.
+Return only tags that satisfy these rules.
 INSTRUCTION;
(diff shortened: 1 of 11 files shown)


jeffpaul (approved):
Works as expected, thanks!

PR #1110, opened 2026-06-12
Add more PHPUnit tests

dkotter:
### Description of the Change

We've had some PHPUnit tests in place from the very beginning but it's not something we've actively maintained. With the advancements in AI coding proficiency though, seemed like a no-brainer to throw AI at our codebase and add in additional tests where it sees fit (which is exactly what I did in this PR).

Likely not full test coverage (which is not something I think we need to strive for) and likely could write better tests myself but AI was able to do it in a fraction of the time and seems best to increase our test coverage.

### How to test the Change

Only adding tests so ensure those pass on this PR

### Changelog Entry

> Developer - Added additional PHPUnit tests to cover areas that were missing test coverage.

### Credits

Props @dkotter

### Checklist:

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2F10up%2Fclassifai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-1110-9ef6532a57db8f80eeb557d800ae77c7fd9c42fa.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


jeffpaul:
@dkotter plugin check is also failing, Copilot seems confident in recommending a change to resolve but curious if you're seeing this elsewhere or are otherwise not concerned that there's an issue needing to be resolved natively in our ClassifAI check here?

dkotter:
> @dkotter plugin check is also failing, Copilot seems confident in recommending a change to resolve but curious if you're seeing this elsewhere or are otherwise not concerned that there's an issue needing to be resolved natively in our ClassifAI check here?

Plugin Check has been failing for a few weeks now due to an upstream issue (see https://github.com/WordPress/plugin-check-action/issues/579). This is the case not only here but on all of our repos that use that (including the AI plugin). There is a workaround mentioned in that linked issue though I've been waiting/hoping they would just fix the problem upstream. If you'd prefer, I can implement the workaround for now.

dkotter:
And right after leaving that ^ comment, Plugin Check Action v1.1.7 was released to fix the problem. I've bumped to that version in this PR and fixed a couple newly flagged plugin check errors

jeffpaul (approved):
:all-the-things: are passing, looks good, thanks!

PR #1119, opened 2026-06-16
Explore having the LLM return content in JSON schema instead of WordPress block markup

dkotter:
### Description of the Change

For the Content Generation Feature, we specify in our prompt that the content returned should follow the WordPress block markup (with some examples). This seems to work but is fairly fragile as it relies on the LLM knowing what this block markup should look like.

There's a few alternatives to that approach and this PR is exploring one of those: having the LLM return content formatted as JSON and then having us parse that into valid block markup. In theory this should be more reliable and does open up a path where we could support a wider-range of block types (for example, blocks specific to a site or that come from a plugin that some sites may be using).

Some of the work here comes from the `@10up/block-renderer-core` package (I was wanting to use the full suite but ran into issues getting that working in the block editor cc / @fabiankaegy). Our prompt is still static so won't include custom blocks but we now ask for things formatted as JSON and then parse that on the client-side to create our blocks.

### How to test the Change

1. Checkout this branch and run `npm i && npm run build`
2. Turn on the Content Generation Feature and configure it correctly
3. Create a new post and then click on the sparkle icon in the lower right corner of the editor
4. In the box that opens up, add your article summary you want to generate (i.e. brief history of WordPress)
5. Send that request off and ensure you see valid results rendered
6. Ensure you can insert that content into the editor and it renders properly there

### Changelog Entry

> Changed - For our Content Generation Feature, have the LLM return JSON data instead of block markup and then parse that out ourselves

### Credits

Props @fabiankaegy, @dkotter 

### Checklist:

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2F10up%2Fclassifai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-1119-1f3d11078e4d0009746b1dede8f9465378caeabf.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>


includes/Classifai/Helpers.php
```diff
@@ -988,3 +988,52 @@ function get_temperature( float $temperature, int $results = 1 ): float {
...
+ *
+ * Decode the tree with json_decode( $json ) (objects, not associative arrays)
+ * so that empty objects such as `"props":{}` survive re-encoding; decoding to
+ * associative arrays would turn them into `[]`, which the client-side block
+ * tree schema rejects.
+ *
+ * @param mixed $value Decoded block tree, or a nested value within it.
+ * @return mixed Sanitized value.
+ */
+function sanitize_generated_block_tree( $value ) {
```
peterwilsoncc, line 1010: Is it possible to ensure that blocks with specific parent requirements are enforced, eg column is a child of columns? It's in the instructions but it would be good to remove the element of trust.

includes/Classifai/Providers/Azure/OpenAI.php
```diff
@@ -954,8 +955,9 @@ public function generate_content( int $post_id = 0, array $args = array() ) {
 		$body = apply_filters(
 			'classifai_azure_openai_content_request_body',
 			array(
-				'messages'    => $messages,
-				'temperature' => 0.9,
+				'messages'        => $messages,
+				'temperature'     => 0.9,
+				'response_format' => array( 'type' => 'json_object' ),
```
peterwilsoncc, line 960: Document introduction of `response_format` in X.X.X and that prior versions require block formatted code.

:1234: Applies to other providers too.

includes/Classifai/Providers/Azure/OpenAI.php
```diff
@@ -977,17 +979,28 @@ public function generate_content( int $post_id = 0, array $args = array() ) {
...
 			}
 		}
 
-		return $return;
+		// The response should be a JSON BlockTree; validate before returning.
+		// Decode to objects (not arrays) so empty objects like "props":{} are
+		// preserved when re-encoded rather than becoming "props":[].
+		$decoded = json_decode( $content );
+		if ( null === $decoded || JSON_ERROR_NONE !== json_last_error() ) {
+			return new WP_Error( 'invalid_content_response', esc_html__( 'The generated content was not in the expected format. Please try again.', 'classifai' ) );
```
peterwilsoncc, line 997: For developers using the filter and returning block formatted code, a check here for `<!-- wp:` and gracefully falling back to the older format would be lovely.

:1234: Applies to other providers too.

peterwilsoncc (commented):
I've added some notes inline re validation and handling the old format in the event the filters are used.

The backcompat would need to be handled in the JS too, so if it's all a bit messy then maybe the filters can be renamed to include `_json` or similar so out of date plugins don't trigger errors and the use the engine instead.

PR #1122, opened 2026-06-18
Background image processing mode

dkotter:
### Description of the Change

For our Image Processing Features, we currently have two processing modes: automatic or manual. Automatic fires when an image is uploaded and blocks the request until processing is done. We then added the manual mode which allows you to manually trigger the Feature on an image-by-image basis.

This PR adds in a new mode that will trigger automatically but runs as a background process so requests aren't blocked. This utilizes Action Scheduler to schedule and run an event after an image has been uploaded. While this event is running, we disable the scan buttons and show a processing message. Once that event completes, we automatically update the UI with the result (assuming a user has kept the UI open).

This allows users to run these Features automatically on each item but avoid having requests be blocked while processing is happening. They can upload an image, insert it into the content and do additional work while behind the scenes, alt text generation is happening or OCR is running.

### How to test the Change

1. Checkout this PR and run `npm i && npm run build` 
2. Turn on an Image Processing Feature like Descriptive Text Generation
3. Enable the new `Automatically on upload (background)` processing mode
4. Go to the Media Library and ensure you're on the grid view
5. Upload an image and then open that image in the media modal
6. Ensure the generate button for that Feature is disabled and shows a processing message
7. For quicker testing, in a new window, go to `Tools > Scheduled Actions` and find the `classifai_schedule_image_process_job` scheduled action
8. Run that action
9. Back in the other window, ensure the button is no longer disabled and the alt text is populated (or whatever Feature you're testing)
10. Test again within a post using the image block. Add an image block, open the Media Library and drag in an image to upload
11. Run through the same tests again

### Screenshots

https://github.com/user-attachments/assets/6c888e47-88ec-4119-b60d-da8b130e481b

### Changelog Entry

> Added - New processing mode for Image Processing Features that will run the Feature automatically but on a background process

### Credits

Props @dkotter

### Checklist:

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2F10up%2Fclassifai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-1122-d54d78ff11b2b500a8d663c3e5485c741c3f1527.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Classifai/Features/AsyncImageProcessing.php
+++ b/includes/Classifai/Features/AsyncImageProcessing.php
@@ -1,0 +1,254 @@
+<?php
+
+namespace Classifai\Features;
+
+use Classifai\Services\ImageProcessing;
+
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Shared async (Action Scheduler) processing logic for Image Processing features.
+ *
+ * Used by the Feature classes that act on uploaded images (descriptive text,
+ * image tags, OCR and smart cropping). When a Feature's `processing_mode` is set
+ * to `automatic_async`, the upload hook enqueues an Action Scheduler job instead
+ * of running the provider synchronously during the upload request. The job hook
+ * and status meta key are shared, so the constants live on the
+ * {@see ImageProcessing} service (traits cannot declare constants on PHP 7.4).
+ */
+trait AsyncImageProcessing {
+
+	/**
+	 * Build the Action Scheduler job arguments for an attachment.
+	 *
+	 * The values are spread positionally into {@see handle_async_image_job()} by
+	 * Action Scheduler, so insertion order matters. `feature` + `type` +
+	 * `attachment_id` make the args unique per feature, which lets the shared
+	 * job hook de-dupe per feature via `as_has_scheduled_action()`.
+	 *
+	 * @param int    $attachment_id Attachment ID being processed.
+	 * @param string $type          Run type (e.g. `descriptive_text`, `tags`, `ocr`, `crop`).
+	 * @return array
+	 */
+	public function get_async_job_args( int $attachment_id, string $type ): array {
+		return array(
+			'attachment_id'   => $attachment_id,
+			'feature'         => static::ID,
+			'type'            => $type,
+			'calling_user_id' => get_current_user_id(),
+		);
+	}
+
+	/**
+	 * Enqueue an async image processing job.
+	 *
+	 * @param int    $attachment_id Attachment ID being processed.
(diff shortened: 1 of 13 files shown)


jeffpaul:
Action Scheduler just hit v4, is there a specific version of that plugin we want/need to require here (or anything that's new in v4 that perhaps may not have been accounted for In the original design for this PR)?

dkotter:
> Action Scheduler just hit v4, is there a specific version of that plugin we want/need to require here (or anything that's new in v4 that perhaps may not have been accounted for In the original design for this PR)?

I saw that release went out and did look through the changelog and nothing caught my eye as being immediately useful.

We also just a week or so ago bumped Action Scheduler to what was then the latest version: https://github.com/10up/classifai/pull/1087.

I think it would be good to bump to 4.0 but I'd rather wait to give that some time to be more fully tested before making the leap (I know the next version of Woo is shipping with it and wouldn't be surprised if that ends up finding a few things that need fixed)

jeffpaul:
Instead of adding another processing mode, why not just update the `Automatically on upload` to be the background approach?  Its not like that second automatic option that is added in this PR forces Action Scheduler as a plugin dependency install or something.

Screenshot 2026-06-24 at 10 32 59 AM

Otherwise I'm running into some issues testing on Playground and having images accessible, so I'll switch to testing on a test site elsewhere to see that this works as expected.

dkotter:
> Instead of adding another processing mode, why not just update the Automatically on upload to be the background approach? Its not like that second automatic option that is added in this PR forces Action Scheduler as a plugin dependency install or something.

So that is a valid approach we could take. I decided not to do that for now because this is a change to the flow. Because Action Scheduler works via cron, there is a delay between uploading and the image being processed. This delay may be seconds but could be longer depending on the site and the traffic it gets.

I see a valid use case for wanting everything to be blocked until that processing is done (which is the current automatic approach). This allows the user to see what AI generated right away whereas this new approach might take a bit to load in (and if you navigate out of the media UI, you won't see those changes until you refresh).

If we want this to be the new default right away, I'm fine with that. But also like the idea of letting people try this out and provide feedback before we make that decision (noting that does require someone actually using this new mode which may not happen).

jeffpaul (approved):

PR #1125, opened 2026-06-23
Add new "On Demand Mode" for Text to Speech processing

dkotter:
### Description of the Change

When the Text to Speech Feature is turned on, there are two "modes" currently: generate speech automatically when content is saved or require generation to be manually turned on at the post level first (so kind of automatic vs manual though both fire on save, there isn't an actual manual button you can press).

This PR introduces a new mode: on-demand. The idea here is to avoid generating speech for posts that no one is going to listen to (and thus costing you more for no reason). As a good example, if you have a large backlog of content that doesn't yet have speech generated, instead of generating speech for all of that upfront, you can generate it on-demand as a user requests it.

With this new mode turned on, the audio controls will show on the front-end and the first time a user clicks to listen, it will make a request to generate the audio first. After it's generated once, that audio is stored and re-used for any subsequent requests so unless the content changes, audio is only generated on-demand once.

Couple things to keep in mind:

1. This does mean the first time a user wants to listen to a post they'll have to wait for audio to generate. For longer posts, this could be 30 seconds or more
2. Because this is a front-end generation, we can't limit based on users being logged in. This does open the door for someone to trigger audio generation on a bunch of posts and thus lead to a large amount of usage. We don't allow multiple generation requests per post and we don't generate audio if it already exists so the most impact you'd see is one generation request per post. This may be a dealbreaker for merging this in, though I don't see this as being any different from automatically generating audio for each post
3. There's a chance on a high traffic site for multiple users to request audio at the same time. We do have protection in place to only generate once but for any users that try generating while a previous generation is happening, they'll get an error message until generation is done

### Screenshot and screenshare

New audio generation settings

https://github.com/user-attachments/assets/6d09789e-8ba7-4a42-b35e-3cd682fb0ec5

### How to test the Change

1. Checkout this PR and run `npm i && npm run build`
2. Turn on the Text to Speech Feature and choose the new On demand Audio generation mode
3. Create a new post and ensure the audio generation toggle is on in the sidebar
4. Publish this post and go to the front-end
5. Ensure the audio controls show
6. Click the play button and ensure you get a loading indicator
7. Ensure once that is done the audio plays
8. Refresh the page, click the play button again and ensure this runs immediately
9. If desired, create a new post and turn off the audio generation toggle
10. Ensure the audio controls don't show for that post

### Changelog Entry

> Added - On demand generation mode for Text to Speech. When enabled, TTS is never run automatically on publish but is only run when the audio is requested on the front-end by a user.

### Credits

Props @dkotter

### Checklist:

<a href="https://playground.wordpress.net?blueprint-url=data:application/json,%7B%22preferredVersions%22%3A%7B%22wp%22%3A%22latest%22%7D%2C%22steps%22%3A%5B%7B%22step%22%3A%22installPlugin%22%2C%22pluginData%22%3A%7B%22resource%22%3A%22url%22%2C%22url%22%3A%22https%3A%2F%2Fgithub.com%2F10up%2Fclassifai%2Freleases%2Fdownload%2Fci-artifacts%2Fpr-1125-2a3d9f4161c68f4ebf0d78494ba43b00710e66d6.zip%22%7D%7D%5D%7D" target="_blank" rel="noopener noreferrer">
</a>

### DIFF
--- a/includes/Classifai/Features/TextToSpeech.php
+++ b/includes/Classifai/Features/TextToSpeech.php
@@ -105,6 +105,7 @@ class TextToSpeech extends Feature {
 		add_action( 'add_meta_boxes', array( $this, 'add_meta_box' ) );
 		add_action( 'admin_notices', array( $this, 'show_error_if' ) );
 		add_action( 'save_post', array( $this, 'save_post_metadata' ), 5 );
+		add_action( 'save_post', array( $this, 'maybe_invalidate_on_demand_audio' ), 20 );
 		add_action( 'wp_ajax_classifai_get_tts_status', array( $this, 'ajax_get_audio_generation_status' ) );
 		add_action( 'admin_enqueue_scripts', array( $this, 'enqueue_admin_assets' ) );
 	}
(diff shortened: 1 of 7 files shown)


jeffpaul (approved):
A test to fix up, otherwise good to go

